PR #12, opened 2017-09-08
Create initial project/server structure.

chambridge:


CONTRIBUTING.rst
```diff
@@ -0,0 +1,86 @@
...
+quipucords uses `pytest`_ for testing.
+
+
+Running all tests:
+******************
+
+.. code-block:: bash
+
+    # Run all tests on the current Python interpreter
+    make tests
```
noahl: This says `make tests`, but the Makefile docstring says `make test`, and then the actual Makefile has `tests`. I suggest we use `test`, because it is one letter shorter and because it is more similar to `make lint`.

CONTRIBUTING.rst
```diff
@@ -0,0 +1,86 @@
...
+
+.. code-block:: bash
+
+    make lint
+
+Testing
+-------
+
+Before opening a pull requests, please make sure the `tests`_ pass
+in all of the supported Python environments (3.5, 3.6).
```
noahl: :+1: :+1: :+1: to dropping Python 2 for the controller.

elyezer: Python 3 :heart:

chambridge: Based on Jag's input earlier. Django wont support Python 2 much longer. We can fight the battle later on minimum Python support.

PR #22, opened 2017-09-13
Issues/20

chambridge:


quipucords/api/models.py
```diff
@@ -31,13 +31,24 @@ class HostCredential(Credential):
...
         if self.password:
             self.password = encrypt_data_as_unicode(self.password)
         if self.sudo_password:
             self.sudo_password = encrypt_data_as_unicode(self.sudo_password)
+
+    def save(self, *args, **kwargs):
+        """Save the model object
+        """
+        self.encrypt_fields()
         super(HostCredential, self).save(*args, **kwargs)
```
noahl: Since we are Python 3 only, you can just use `super().save(...)`.

(https://docs.python.org/3/library/functions.html#super)

quipucords/api/serializers.py
```diff
@@ -26,3 +28,23 @@ class HostCredentialSerializer(ModelSerializer):
     class Meta:
         model = HostCredential
         fields = '__all__'
+
+    def validate(self, attrs):
```
noahl: Can we make the validate method simpler? I think something like

`ssh_keyfile = 'ssh_keyfile' in attrs and attrs['ssh_keyfile']`, and similar for `password`. Then `ssh_keyfile` would be either `False`, if it wasn't in attrs, `None`, if it was in attrs but None, or a nonempty string with a path. The conditionals could use implicit conversion to bool to check the False and None cases at the same time.

chambridge:
@noahl merge conflicts resolved.

chambridge:
@noahl Updates made. good suggestions.

PR #47, opened 2017-09-21
Add base cli setup and testing. Close #27.

### LINKED ISSUE #27: Create scaffolding for command line interaction with server
## Specify type:
- Enhancement
___

## Description:
Provide python module and base CLI command class. 

Base class should provide:

- help setup
- validation
- command execution
- core processing for executing rest request and delegation of response handling

Module should also support general help and version information from command line

chambridge:


quipucords/cli/cli.py
```diff
@@ -0,0 +1,65 @@
...
+from argparse import ArgumentParser
+import cli.auth as auth
+from cli.auth.add import AuthAddCommand
+from . import __version__
+
+
+# pylint: disable=too-few-public-methods
+class CLI(object):
+    """Class responsible for displaying ussage or matching inputs
+    to the valid set of commands supported by rho.
```
noahl: Nit: "quipucords" instead of "rho"

noahl: Or probably "qpc"

quipucords/cli/clicommand.py
```diff
@@ -0,0 +1,48 @@
...
+# pylint: disable=too-few-public-methods
+class CliCommand(object):
+
+    """ Base class for all sub-commands. """
+    def __init__(self, subcommand, action):
+        self.subcommand = subcommand
+        self.action = action
+        self.args = None
+
+    def _validate_args(self):
```
noahl: Nit: can we please remove the leading `_` from this and `do_command`? Those always bothered me in Rho.

quipucords/cli/tests.py
```diff
@@ -0,0 +1,59 @@
...
+
+import contextlib
+import unittest
+import sys
+import six
+from cli.cli import CLI
+
+
+# pylint: disable=too-few-public-methods
+class HushUpStderr(object):
```
noahl: Let's move `HushUpStderr` and `redirect_stdout` to a common test utilities library.

setup.py
```diff
@@ -13,6 +23,12 @@
     ],
     include_package_data=True,
     license='GPLv3',
+    packages=find_packages(exclude=["test*.py"]),
     package_data={'': ['LICENSE']},
     url='https://github.com/quipucords/quipucords',
+    # non-python scripts go here
```
noahl: This says non-python scripts go here, but `bin/qpc` is Python. No request for change, I just don't understand what this does.

Makefile
```diff
@@ -31,16 +31,16 @@ install: build
...
 test-coverage:
-	coverage run --source=quipucords/ quipucords/manage.py test -v 2 quipucords;coverage report -m
+	coverage run --source=quipucords/ quipucords/manage.py test -v 2 quipucords/;coverage report -m
 
 lint-flake8:
 	flake8 . --ignore D203
 
 lint-pylint:
-	$(PYTHON) -m pylint --load-plugins=pylint_django quipucords/*/*.py
+	find . -name "*.py" | xargs $(PYTHON) -m pylint --load-plugins=pylint_django --disable=duplicate-code
```
noahl: What code shows up when you enable the duplicate code checking? I would prefer to keep it enabled and spend a little more effort on refactoring if the warnings are at all reasonable.

quipucords/cli/cli.py
```diff
@@ -0,0 +1,65 @@
...
+# version 3 (GPLv3). There is NO WARRANTY for this software, express or
+# implied, including the implied warranties of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv3
+# along with this software; if not, see
+# https://www.gnu.org/licenses/gpl-3.0.txt.
+#
+""" Quipucords Command Line Interface """
+
+from __future__ import print_function
+from argparse import ArgumentParser
```
noahl: :+1: :+1: for changing to a non-deprecated argument parser!

quipucords/cli/clicommand.py
chambridge: is that not convention for denoting a method as private/internal? Meaning those two methods should not be called directly.

Makefile
chambridge: Only duplicate code is in the test files and I couldn't get the disable to work across files (several stackoverflows on this):

```
R:  1, 0: Similar lines in 2 files
==cli.auth.tests:21
==cli.tests:21
    def setUp(self):
        # Temporarily disable stderr for these tests, CLI errors clutter up
        # nosetests command.
        self.orig_stderr = sys.stderr
        sys.stderr = HushUpStderr()

    def tearDown(self):
        # Restore stderr
        sys.stderr = self.orig_stderr
 (duplicate-code)
R:  1, 0: Similar lines in 2 files
==cli.auth.tests:12
==cli.tests:12
import unittest
import sys
import six
from cli.cli import CLI
from cli.tests_utilities import HushUpStderr, redirect_stdout

 (duplicate-code)

```

quipucords/cli/cli.py
chambridge: updated

quipucords/cli/tests.py
chambridge: done

setup.py
chambridge: fixed. copy & paste from rho. I don't understand it either.

Makefile
noahl: Okay, fine with me.

quipucords/cli/clicommand.py
noahl: Yes, that's the convention. Maybe it bothers me because it's meant to be overridden, I'm not sure. Anyway, LGTM.

PR #48, opened 2017-09-21
Create Model and ViewSet for NetworkProfile objects

### LINKED ISSUE #25: Define Django Rest Framework Network Profile API model and view set
## Specify type:
- Enhancement
___

## Description:
Create necessary model and view set for Network Profile model.

REST APIs should cover standard CRUD flow:

    Create
    Read
    Update
    Delete

Provide associated unit test to exercise the model and view set flow along with constraints.

Constraints:

- Network Profile must have a name
- Network Profile must have at least one host
- Network Profile host values must be of valid inputs (ip, ip ranges, hostnames)
- Network Profile can have an ssh port but it must be within a valid port range
- Network Profile must have valid host credentials

noahl:
This allows basic CRUD operations on NetworkProfiles, and does some
validation. The validation is not complete, but importing all of the
normalization and validation we do in rho will be a big change, and I
think it makes sense to get this reviewed first.

Closes #25 .

### DIFF
--- a/quipucords/api/models.py
+++ b/quipucords/api/models.py
@@ -53,3 +53,16 @@ class HostCredential(Credential):
 
     class Meta:
         verbose_name_plural = 'Host Credentials'
+
+
+class NetworkProfile(models.Model):
+    """A network profile connects a list of credentials and a list of hosts."""
+    name = models.CharField(max_length=64, unique=True)
+    # Comma-separated list of host specifiers
+    hosts = models.CharField(max_length=1024)
+    ssh_port = models.IntegerField()
+    # Comma-separated list of credential names. We identify
+    # credentials by name, not ID, to support the behavior where a
+    # user creates a credential of a given name, deletes it, and makes
+    # a new one of the same name.
+    credentials = models.CharField(max_length=1024)
--- a/quipucords/api/models.py
+++ b/quipucords/api/models.py
@@ -58,11 +58,32 @@ class HostCredential(Credential):
 class NetworkProfile(models.Model):
     """A network profile connects a list of credentials and a list of hosts."""
     name = models.CharField(max_length=64, unique=True)
-    # Comma-separated list of host specifiers
-    hosts = models.CharField(max_length=1024)
     ssh_port = models.IntegerField()
-    # Comma-separated list of credential names. We identify
-    # credentials by name, not ID, to support the behavior where a
-    # user creates a credential of a given name, deletes it, and makes
-    # a new one of the same name.
-    credentials = models.CharField(max_length=1024)
+    credentials = models.ManyToManyField(HostCredential)
+    # NetworkProfile also has the field hosts, which is created by the
+    # ForeignKey in HostRange below.
+
+
+class HostRange(models.Model):
+    """A HostRange is a subset of a network to scan.
+
+    It can be either an IP range or a DNS name range. A HostRange is
+    not the same as a set of hosts because there may be parts of the
+    IP or DNS range that don't correspond to any host, but we still
+    need to remember them because in the future there could be hosts
+    there.
+    """
+
+    # HostRanges provide a convenient way for a NetworkProfile to
+    # store a list of name ranges.  As of now we don't make any effort
+    # to deduplicate when different NetworkProfiles have overlapping
+    # (or identical) ranges. If we ever want to show the user exactly
+    # what parts of their network are being scanned, we will have to
+    # dedup.
+
+    # host_range is an IP address range or a DNS name range in Ansible
+    # format.
+    host_range = models.CharField(max_length=1024)
+    network_profile = models.ForeignKey(NetworkProfile,
+                                        models.CASCADE,
+                                        related_name='hosts')
--- a/quipucords/api/models.py
+++ b/quipucords/api/models.py
@@ -59,7 +59,7 @@ class NetworkProfile(models.Model):
     """A network profile connects a list of credentials and a list of hosts."""
     name = models.CharField(max_length=64, unique=True)
     ssh_port = models.IntegerField()
-    credentials = models.ManyToManyField(HostCredential)
+    credential_ids = models.ManyToManyField(HostCredential)
     # NetworkProfile also has the field hosts, which is created by the
     # ForeignKey in HostRange below.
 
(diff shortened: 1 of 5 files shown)


quipucords/api/serializers.py
```diff
@@ -39,3 +39,46 @@ def validate(self, attrs):
...
+
+
+class NetworkProfileSerializer(ModelSerializer):
+    """Serializer for the NetworkProfile model"""
+    class Meta:
+        model = NetworkProfile
+        fields = '__all__'
+
+    @staticmethod
+    def validate_name(name):
```
chambridge: Seems like we should have this for Host Credentials as well.

quipucords/api/models.py
```diff
@@ -53,3 +53,16 @@ def update(self, request, *args, **kwargs):
 
     class Meta:
         verbose_name_plural = 'Host Credentials'
+
+
+class NetworkProfile(models.Model):
+    """A network profile connects a list of credentials and a list of hosts."""
+    name = models.CharField(max_length=64, unique=True)
+    # Comma-separated list of host specifiers
+    hosts = models.CharField(max_length=1024)
```
chambridge: Doesn't seem to match the swagger docs where we are looking for an array of strings. I expected we would be trying to utilize some of the Many to Many Modeling capabilities in Django (https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.ManyToManyField.through_fields) so we can take an array and not a comma separated string.

quipucords/api/models.py
```diff
@@ -53,3 +53,16 @@ def update(self, request, *args, **kwargs):
...
+    ssh_port = models.IntegerField()
+    # Comma-separated list of credential names. We identify
+    # credentials by name, not ID, to support the behavior where a
+    # user creates a credential of a given name, deletes it, and makes
+    # a new one of the same name.
+    credentials = models.CharField(max_length=1024)
```
chambridge: Swagger doc had this as *credential_ids*. I think we should use an ID and and again some of the modeling relationships so we can consume an array here and we are referencing a stored Host Credential. Django should be doing some of the "existence" checking for us.

quipucords/api/serializers.py
noahl: This is based on what HostCredentials does. The advantage to the field-by-field approach is that if you're doing a partial update, Django won't check the fields that aren't part of the update, and therefore won't automatically fail the update. I will update HostCredentials too.

quipucords/api/models.py
noahl: I like using many-to-many fields, but it changes the semantics from what Rho does. In Rho, a user can delete a credential that they're using for a profile and then recreate a new credential with the same name and it will be fine. Using many-to-many fields would change that. I think it makes sense to use many-to-many, but do you think the change will be a problem for people?

noahl: Sorry, this comment was about credentials below.

quipucords/api/serializers.py
chambridge: Yes, I was suggesting we add the validate_name such that we check that its printable to Host Credentials serializer.

quipucords/api/models.py
chambridge: I think its fine to diverge from what `rho` does. While `rho` allowed for the above flow I'm not sure it was designed that way, but just a usable side effect. However, along the same lines it introduced an issue where users could unintentionally delete a credential and then run a scan with a profile that utilized it. I think the referential integrity is an improvement.

elyezer: Will this eventually become a one to many or many to many relationship? I mean having the size of 1024 will allow specifying at most 60 hosts considering IPv4 addresses considering IP addresses like `192.168.120.142` and separated by a comma without spaces around the comma. 

PS I have not checked the Swagger docs, I will do that!

quipucords/api/models.py
elyezer: I think it makes sense to use a relationship here. The side effect of recreating a credential with the same name can be worked around by taking advantage of an edit on the credential.

quipucords/api/serializers.py
```diff
@@ -39,3 +39,46 @@ def validate(self, attrs):
...
+        """Validate the name of the NetworkProfile."""
+        if not isinstance(name, str) or not name.isprintable():
+            raise ValidationError('NetworkProfile must have printable name.')
```
elyezer: It would be better to explicitly say what is printable, like characters in a-z, A-Z, and 0-9 for example. If I remember correctly from my past experience (~4-5 years ago) with Django this validation error's message can appear as a message on UI, that is why I think it would be better to improve it.

quipucords/api/serializers.py
```diff
@@ -39,3 +39,46 @@ def validate(self, attrs):
...
+
+        return name
+
+    @staticmethod
+    def validate_hosts(hosts):
+        """Make sure the hosts list is reasonable."""
+        host_parts = hosts.split(',')
```
elyezer: If it is decided to not use relationship with hosts then this will be considering spaces as valid information. I think this validation could be improved. It seems that it will accept values like `ghj,kl;,#$%` it is a comma separated value but the parts are not valid hosts.

Please excuse me if the hosts parts validation is done somewhere else, I need to catch up how Django and Django REST framework is working now a days.

chambridge: I think Noah was planning to do more of the validation in a follow on  PR per his open PR remarks, but we should go ahead and open an issue describing that work in that case.

quipucords/api/serializers.py
```diff
@@ -11,9 +11,11 @@
 """Module for serializing all model object for database storage"""
 
 import os
+from django.db import transaction
 from django.utils.translation import ugettext as _
-from rest_framework.serializers import ModelSerializer, ValidationError
-from api.models import Credential, HostCredential
+from rest_framework.serializers import ModelSerializer, ValidationError, \
+    SlugRelatedField, PrimaryKeyRelatedField
```
kdelee: Another option for breaking up long import lines is

```
from rest_framework.serializers import (
    ModelSerializer,
    ValidationError,
    SlugRelatedField,
    PrimaryKeyRelatedField,
)
```
This is also a bit more extensible, creating smaller diffs in the future if more imports are added.

chambridge:
Added the following issue to capture needed validation: https://github.com/quipucords/quipucords/issues/77

PR #49, opened 2017-09-25
Add qpc auth add cli and associated test cases. Closes #28.

### LINKED ISSUE #28: Create CLI command for adding a host credential
## Specify type:
- Enhancement

## Description:
Create command and test cases for adding a host credential

chambridge:


quipucords/cli/auth/tests.py
```diff
@@ -24,16 +35,96 @@ def setUp(self):
...
             sys.argv = ['/bin/qpc', 'auth', 'add', '--name', 'auth1']
             CLI().main()
-            self.assertEqual(auth_out.getvalue(),
-                             'Auth "auth1" was added\n')
+
+    def test_add_bad_key(self):
+        """Testing the add auth command when providing an invalid path for
+        the sshkeyfile.
+        """
+        auth_out = six.StringIO()
```
noahl: Question: since quipucords the server is requiring Python 3.5 or greater, does the CLI require Python 3 as well?

If yes, no need to use `six` here.

PR #76, opened 2017-09-29
Update readme with required package info

kdelee:
`/usr/bin/2to3` is required by one of the dependencies installed by pip,
and this is provided by the `python-tools` package in both Fedora and RHEL
7

Just a small update, but this not installed on a standard RHEL 7 or Fedora 25 install, and may slow users down, as the error  message is not totally clear, nor what package might provide this utility.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -79,6 +79,13 @@ Installing Dependencies
 From within the local clone root directory run the following command to install
 dependencies needed for development and testing purposes::
 
+First, you need to collect some packages available through either `yum` (RHEL)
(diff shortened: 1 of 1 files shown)


chambridge:
suggest `dnf` or `yum` install flow.

kdelee:
@chambridge I have updated it to suggest either yum or dnf.

kdelee:
@chambridge I don't have write access so could you rebase & merge?

chambridge:
@kdelee You should have access now.

PR #79, opened 2017-10-02
Ingest host data via a file. Closes #78.

### LINKED ISSUE #78: Add/edit hosts via CLI by providing input file
## Specify type:
- Enhancement

## Description:
Allow `--hosts` argument to ingest a file of newline separated host values.

chambridge:


quipucords/cli/network/add.py
```diff
@@ -54,6 +56,13 @@ def __init__(self, subparsers):
     def _validate_args(self):
         CliCommand._validate_args(self)
 
+        if self.args.hosts and len(self.args.hosts) == 1:
+            # check if a file and read in values
+            filename = self.args.hosts[0]
+            input_path = os.path.expanduser(os.path.expandvars(filename))
```
noahl: It looks like you put the `os.path.expanduser(os.path.expandvars(filename))` inside `read_in_file`, but I see that you still need this because of the `os.path.isfile()`. Can we change the interface so we don't have to duplicate this code?

Maybe `read_in_file` could raise a special exception if the path is not a file, and this code could catch that exception and keep going. Or maybe return `None` if the input's not a file, if that's easier.

quipucords/cli/network/edit.py
```diff
@@ -60,6 +63,13 @@ def _validate_args(self):
             self.parser.print_help()
             sys.exit(1)
 
+        if self.args.hosts and len(self.args.hosts) == 1:
+            # check if a file and read in values
+            filename = self.args.hosts[0]
+            input_path = os.path.expanduser(os.path.expandvars(filename))
```
noahl: Same as previous comment.

quipucords/cli/utils.py
```diff
@@ -98,3 +98,20 @@ def pretty_print(json_data):
...
+    """Read values from file into a list object. Expecting newline delimited.
+
+    :param filename: the filename to read
+    :returns: the list of values found in the file
+    """
+    result = None
+    input_path = os.path.expanduser(os.path.expandvars(filename))
+    try:
+        with open(input_path, 'r') as in_file:
+            result = in_file.read().splitlines()
```
noahl: Small note: `in_file.readlines()` does this.

chambridge: Calling that function gives back each line with the newline character still present.
`['1.2.3.4\n', '1.2.3.[1:10]\n']`

The above gives me back `['1.2.3.4', '1.2.3.[1:10]']` which I'm looking for and is less processing than then stripping of the newline character.

chambridge:
@noahl  Updates pushed per your comments.

quipucords/cli/utils.py
noahl: @chambridge I didn't know that! Thanks for explaining.

PR #125, opened 2017-10-24
Generate fingerprints with expanded fact set (cpu, virtualization, connection, sys creation time)

kholdaway:


quipucords/fingerprinter/__init__.py
```diff
@@ -39,9 +41,88 @@ def process_fact(self, fact_collection_id, fact):
...
         """
-        fingerprint = {'fact_collection_id': fact_collection_id,
-                       'os_name': fact['etc_release_name'],
-                       'os_release': fact['etc_release_release'],
-                       'os_version': fact['etc_release_version']}
+        # Set fact collection id
+        fingerprint = {'fact_collection_id': fact_collection_id}
+
+        # Set OS information
+        if fact['etc_release_name']:
```
chambridge: `if 'etc_release_name' in fact:`

PR #163, opened 2017-10-31
consolidate messages for content review. Closes #131.

### LINKED ISSUE #131: Consolidate CLI messages into single file for doc review
## Specify type:
- Enhancement

## Description:
Consolidate the CLI messages into a single file for doc review and translation.

chambridge:


qpc/utils.py
```diff
@@ -16,6 +16,8 @@
 import os
 import json
 from xdg.BaseDirectory import xdg_data_home, xdg_config_home
+from qpc.translation import _ as t
```
noahl: Why import `_` as `t` when it is `_` in all of the other modules?

chambridge: there was an existing use of `_` within the file, so the linter called out a shadow variable issue. Seemed simple enough to just do the above since it was such a limited usage.

PR #168, opened 2017-11-02
Break hosts into groups that match the size of the concurrency setting

kholdaway:


quipucords/scanner/utils.py
```diff
@@ -95,24 +95,42 @@ def construct_connect_inventory(hosts, credential, connection_port):
     return inventory
 
 
-def construct_scan_inventory(hosts, connection_port):
+def construct_scan_inventory(hosts,
+                             connection_port,
+                             group_prefix,
```
chambridge: do we need to pass group_prefix?

PR #172, opened 2017-11-02
Update the discover scan progress

### LINKED ISSUE #134: I want to see the progress of systems being scanned via a counter and total providing me with further user feedback on completeness of the scan.
## Specify type:
- Enhancement

## Description:
Scan Job status information for host scans should provide a value for the total number of systems (based on the discovery) to be scanned and a value for the number of systems completed.

**_REST:_**

**Method:** GET
**URI:** `api/v1/scans/<scan_job_id>/`
**Response:** 200
```
{
"id": <integer>,
"profile_id": <integer>,
"scan_type": "host",
"status: "running",
"max_concurrency": <positive integer>
"systems_count": <positive integer>,
"systems_scanned": <positive integer>
}
```

____
Work items:

- [x] #141 
- [ ] #142 
- [x] #143

### LINKED ISSUE #142: Showing progress during a discovery scan
## Specify type:
- Enhancement

## Description:
A discovery scan should display null values for `systems_count` & `systems_scanned` until the scan is completed for the first credential. 

Once that is done we should be able to update the `systems_count` with the total number of systems attempting to be scanned (success + failed) and we can update the `systems_scanned` with the total number of success. 

Each additional credential should update the `systems_scanned` with the accumulated # of successful connections.

When we are done with all credentials `systems_count==systems_scanned`.

If there is only 1 credential then you will essentially see the values go from null to `systems_count==systems_scanned`.

Associated with user story #134.

kholdaway:
Closes #142 
Closes #134

### DIFF
--- a/quipucords/scanner/discovery.py
+++ b/quipucords/scanner/discovery.py
@@ -117,6 +117,14 @@ class DiscoveryScanner(Thread):
             cred = hc_serializer.data
             connected, remaining = connect(remaining, cred, connection_port,
                                            forks=forks)
+
+            # Update the scan counts
+            if self.scanjob.systems_count is None:
+                self.scanjob.systems_count = len(connected) + len(remaining)
+                self.scanjob.systems_scanned = 0
+            self.scanjob.systems_scanned += len(connected)
+            self.scanjob.save()
+
             if remaining == []:
                 break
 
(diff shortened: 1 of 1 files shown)


quipucords/scanner/discovery.py
```diff
@@ -117,6 +117,14 @@ def discovery(self):
             cred = hc_serializer.data
             connected, remaining = connect(remaining, cred, connection_port,
                                            forks=forks)
+
```
chambridge: This also runs for a host scan. So need to toggle based on that.

kholdaway: Put check for discovery scan

PR #209, opened 2017-11-17
Reload running or pending scan jobs are server restart

### LINKED ISSUE #137: I want to be able to restart a paused host scan from REST and CLI.
## Specify type:
- Enhancement

## Description:
I want to be able restart a paused scan.

**_REST:_**

**Method:** PUT
**URI:** `api/v1/scans/<scan_job_id/restart/`
**Response:** 200
```
{
"id": <integer>,
"profile_id": <integer>,
"scan_type": "host",
"status: "running",
"max_concurrency": <positive integer>,
"systems_count": <positive integer>,
"systems_scanned": <positive integer>
}
```

**_CLI:_**
`qpc scan restart --id <scan_job_id>`
Add new command for hitting the restart endpoint

___
Work items:
- [x] #149 
- [x] #150 
- [x] #186 
- [ ] #187 
- [x] #203

### LINKED ISSUE #187: Backend updates for scan manager startup
## Specify type:
- Enhancement

## Description:
Scan manager should look through any 'pending' or 'running' scans on startup and queue them up. A 'running' scan should be the first in the queue.

kholdaway:
- Running jobs are load first, then pending
- Required manual env to specify hostname/port.

Closes #187 
Closes #137

### DIFF
--- a/quipucords/api/scanjob/model.py
+++ b/quipucords/api/scanjob/model.py
@@ -56,6 +56,7 @@ class ScanJob(models.Model):
         """Convert to string."""
         return '{' + 'id:{}, '\
             'scan_type:{}, '\
+            'status:{}, '\
             'profile:{}, '\
             'max_concurrency: {}, '\
             'systems_count: {}, '\
@@ -63,6 +64,7 @@ class ScanJob(models.Model):
             'failed_scans: {}, '\
             'fact_collection_id: {}'.format(self.id,
                                             self.scan_type,
+                                            self.status,
                                             self.profile,
                                             self.max_concurrency,
                                             self.systems_count,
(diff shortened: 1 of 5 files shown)


quipucords/api/scanjob/model.py
```diff
@@ -56,13 +56,15 @@ def __str__(self):
         """Convert to string."""
         return '{' + 'id:{}, '\
```
noahl: You don't have to, but one way to protect yourself from typos when you have this many strings to substitute is to pass them as keyword arguments to format, so it would look like `'id:{id}, scan_type:{scan_type}, ...'.format(id=self.id, scan_type=self.scan_type, ...)`

kholdaway: Good point.  I think most of the str method don't do this.  This may be a good fit and finish for all str methods.

PR #211, opened 2017-11-17
Network profile creation now validates for a host credential that is not in the db

### LINKED ISSUE #204: Bad create/update request for network profile gets 500 (internal server error)
## Specify type:
- Bug

- Medium
___

## Description:
If a user attempts to update a profile to use a credential that does not yet exist, they get a 500 error which tells them something is wrong with the server (happens with DEBUG = True and DEBUG = False), and in the logs it shows a traceback.

Instead, the user should ideally get a  `406 (Not Acceptable)`. According to [RFC 7231](https://tools.ietf.org/html/rfc7231#section-6.5.6) they should get back a list of acceptable credentials (i.e. ones that exist).
___

### Expected behavior:
If i attempt to update a profile with an credential id that does not exist, I should be told I did something wrong.

### Actual behavior:
If I do aforementioned bad thing, the server tells me it is broken.

### Steps to reproduce:
1) Create any number of credentials you like. 
2) Create a profile with some of these valid credentials. 
3) Attempt to update profile with invalid credentials (either add them to list or replace list with invalid)

### Environment information:

Server System | Fedora            | 25        | 3.6            | Bare Metal
Client System   | same

kholdaway:
Closes #204

### DIFF
--- a/quipucords/api/networkprofile/tests_networkprofile.py
+++ b/quipucords/api/networkprofile/tests_networkprofile.py
@@ -193,6 +193,14 @@ class NetworkProfileTest(TestCase):
              'ssh_port': '22',
              'credentials': []})
 
+    def test_create_invalid_credentials(self):
+        """A random int is not a valid credential."""
+        self.create_expect_400(
+            {'name': 'netprof1',
+             'hosts': '1.2.3.4',
+             'ssh_port': '22',
+             'credentials': [42]})
+
     def test_list(self):
         """List all NetworkProfile objects."""
         data = {'name': 'netprof',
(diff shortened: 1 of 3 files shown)


quipucords/api/networkprofile/tests_networkprofile.py
```diff
@@ -193,6 +193,14 @@ def test_create_empty_credentials(self):
              'ssh_port': '22',
              'credentials': []})
 
+    def test_create_invalid_credentials(self):
+        """A random int is not a valid credential."""
+        self.create_expect_400(
```
noahl: Okay, but I thought the return code was going to be 404 and not 400. Did it change?

kholdaway: 404 is not found.  Since we are operating on network profiles I would expect a 404 to refer to a missing network profile (not a parameter used to update it).  So if the network profile we intend to update is missing then 404.  But if we found the profile and the update bad is bad (incorrect host cred) its a bad body (400).

We are creating a new network profile its not possible to have a 404 since it doesn't exist yet.

Issue #240, opened 2017-12-04
UI React front-end 


cdcabrera:
# Opening a new issue
___

## Specify type:
- Enhancement

## Description:
React front-end
- [x] UI and build scripts,  PR #243 
- [x] Initial UI navigation layout, PR #251 
- [x] Base template and views, PR #274 #275 
- [x] Initial API hooks into mock, PR #282 
- [x] Base documentation for UI setup, PR #287

PR #282, opened 2017-12-07
Mock API updates

cdcabrera:
Additions for running a mock API from Swagger.
- Additional NPM script
- Bash updates

Part of #240

### DIFF
--- a/client/scripts/api.sh
+++ b/client/scripts/api.sh
@@ -16,8 +16,12 @@
       esac
   done
 
-  if [ -z "$(which docker)" ]; then
-    echo "Setting up API Docker container"
+  if [ -z "$(docker info | grep Containers)" ]; then
+    exit 1
+  fi
+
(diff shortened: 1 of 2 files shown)


client/scripts/api.sh
```diff
@@ -16,17 +17,25 @@
       esac
   done
 
-  if [ -z "$(which docker)" ]; then
-    echo "Setting up API Docker container"
-    docker pull $CONTAINER
+  if [ ! -z "$(which docker)" ]; then
```
jeff-phillips-18: Is the ! correct here?

cdcabrera: I'll double check

cdcabrera: I'll double check

jeff-phillips-18: That's pretty funny 😆

PR #296, opened 2017-12-13
Inspect for network added

kholdaway:
Still not fact endpoint call


chambridge:
Wrong base. Should be merged into dev.

PR #344, opened 2018-01-02
Make InspectResultCallback transactions atomic

### LINKED ISSUE #343: Make InspectResultCallback functions atomic
Right now there is a possibility that quipucords would mark a server as complete, but crash before it could store the results from that server (and updating the server completed count might or might not happen). Should wrap the callback functions in database transactions to prevent this from happening.

noahl:
This prevents issues where quipucords marks a server as complete, but
crashes before storing the results from that server.

Closes #343 .

### DIFF
--- a/quipucords/scanner/network/inspect_callback.py
+++ b/quipucords/scanner/network/inspect_callback.py
@@ -11,6 +11,7 @@
 """Callback object for capturing ansible task execution."""
 
 import logging
+from django.db import transaction
 from ansible.plugins.callback import CallbackBase
 from api.models import (ScanTask, InspectionResult,
                         SystemInspectionResult, RawFact)
@@ -43,6 +44,9 @@ class InspectResultCallback(CallbackBase):
         self.results = []
         self._ansible_facts = {}
 
+    # Needs to be atomic so that the host won't be marked as complete
+    # unless we actually save its results.
+    @transaction.atomic
     def v2_runner_on_ok(self, result):
         """Print a json representation of the result."""
         result_obj = _construct_result(result)
@@ -95,6 +99,10 @@ class InspectResultCallback(CallbackBase):
             else:
                 self._ansible_facts[host] = facts
 
+    # Same as above. The multiple save() statements are hidden inside
+    # _update_reachable_hosts, but they do need to be in the same
+    # transaction as the scan_task.save() below.
+    @transaction.atomic
     def v2_runner_on_unreachable(self, result):
         """Print a json representation of the result."""
         result_obj = _construct_result(result)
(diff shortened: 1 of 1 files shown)


noahl:
After talking to @kholdaway , I've updated to factor out some common code. I also tried to make the transaction blocks smaller, but was ultimately unsuccessful because the SystemInspectionResult and InspectionResult are in a many-to-many relation, and in order to add them to that relation, they both need primary keys. They don't get primary keys until they are saved in the database, which means I need to do the database save fairly early (or else do a big restructuring of the callback, which I don't want to do), which determines the size of the transaction.

kholdaway:
@noahl It looks like the builds are still failing

noahl:
@kholdaway I think I've fixed the builds, please take another look.

quipucords/scanner/network/inspect_callback.py
```diff
@@ -56,73 +58,87 @@ def v2_runner_on_ok(self, result):
...
-                    if inspect_result is None:
-                        inspect_result = InspectionResult(
-                            scan_task=self.scan_task, source=self.source)
-                        inspect_result.save()
-                    inspect_result.systems.add(sys_result)
-                    inspect_result.save()
-                    self.inspect_results.results.add(inspect_result)
-                    self.inspect_results.save()
-
+                    self._finalize_host(host)
```
kdelee: Is there a conceivable program flow where the host finishes before any other ansible facts get set for this host? (Thinking about a key error). Just thinking about a key error down on [line #92](https://github.com/quipucords/quipucords/pull/344/files#diff-11e150690b3e394303383728124f461fR92) that could be avoided with a self._ansible_facts.get(host)

quipucords/scanner/network/inspect_callback.py
```diff
@@ -43,6 +44,7 @@ def __init__(self, scan_task, inspect_results, display=None):
         self.results = []
         self._ansible_facts = {}
 
+    @transaction.atomic
```
kholdaway: This shouldn't be needed now.  I think having _finalize_host atomic is enough.

quipucords/scanner/network/inspect_callback.py
kholdaway: I don't think so.  The `host_done` is there to mark the end of of host scan.

quipucords/scanner/network/inspect_callback.py
noahl: Removed, thank you!

quipucords/scanner/network/inspect_callback.py
noahl: That seems like an edge case that could occur if the user runs a scan but specifies no facts or something like that. I will fix it, thanks!

noahl: Done.

PR #364, opened 2018-01-04
Issues/352

### LINKED ISSUE #352: Merge dmi fingerprints from rho to sonar
Part of #307 .

noahl:
Update DMI roles with tasks from rho.

Closes #352 .

### DIFF
--- a/roles/dmi/tasks/main.yml
+++ b/roles/dmi/tasks/main.yml
@@ -6,9 +6,15 @@
   ignore_errors: yes
   become: yes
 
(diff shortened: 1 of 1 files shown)


noahl:
The test failures look like network connectivity issues. The PR scan failed for Python 3.5 only, and the error is that it couldn't clone quipucords.git from GitHub. The push scan failed, again for Python 3.5, saying that it couldn't install Ansible. The Python 3.6 and 3.4 scans both succeeded.

PR #372, opened 2018-01-08
Update facts endpoint to support multiple network sources (tolerate vcenter but ignore)

kholdaway:
Closes #329 
  
Code coverage will decrease on this because there is some code that needs to be refactored as part of #348.  I will tackle test coverage in #348.

Issue #348, opened 2018-01-03
Scan results from our server should pass facts programmatically instead of REST


kholdaway:
# Opening a new issue
___

## Specify type:
- Enhancement

### Bug severity (if applicable):
- High

## Description:
We should allow external data from here, but for now we should pass data from our scans programmatically (via method call) rather than by HTTP REST to by pass the service account authentication steps for 1.0.  See https://github.com/quipucords/quipucords/issues/329#issuecomment-354687786

PR #374, opened 2018-01-08
Merge uname fingerprints from rho to sonar

### LINKED ISSUE #362: Merge uname fingerprints from rho to sonar
Part of #307 .

ilyavw:
This copies over the rho uname role to quipucords.

Closes #362

### DIFF
--- a/roles/uname/tasks/main.yml
+++ b/roles/uname/tasks/main.yml
@@ -1,0 +1,71 @@
+---
+
+- name: gather uname.os fact
+  raw: uname -s
+  register: uname_os
+  ignore_errors: yes
+
+- name: add uname.os to dictionary
+  set_fact:
+    uname: "{{ uname|default({}) | combine({ item: uname_os['stdout_lines'][0] | default('error') }) }}"
+  with_items:
+  - 'uname.os'
+
+- name: gather uname.hostname fact
+  raw: uname -n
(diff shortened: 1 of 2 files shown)


roles/uname/tasks/main.yml
```diff
@@ -0,0 +1,71 @@
+---
+
+- name: gather uname.os fact
+  raw: uname -s
+  register: uname_os
+  ignore_errors: yes
+
+- name: add uname.os to dictionary
```
chambridge: For Sonar we are flattening the fact structure. No need to add them to a dictionary

infinitewarp:
@chambridge : As we migrate these roles and facts from rho to sonar, do we also want to add them facts to the `SystemFacts` model (and add a DB migration and update `SystemFactsSerializer`)?

chambridge:
@infinitewarp: Good question. However, @kholdaway has been updating the fact flow with https://github.com/quipucords/quipucords/pull/372 and https://github.com/quipucords/quipucords/issues/348 to remove modeling of raw facts and instead have the finger printing engine handle the incoming format so it can create `SystemFingerprint` objects. The `SystemFingerprint` model will be updated in an upcoming story.  For early sprints it was very much 1-to-1 with the raw facts coming in and we were really just trying to support the end-to-end pipeline of data.

PR #376, opened 2018-01-08
Scan results from our server should pass facts programmatically instead of REST

### LINKED ISSUE #348: Scan results from our server should pass facts programmatically instead of REST

kholdaway:
Closes #348


quipucords/api/fact/util.py
```diff
@@ -0,0 +1,183 @@
...
+        invalid_field_obj[FACTS_KEY] = _(
+            messages.FC_REQUIRED_ATTRIBUTE)
+
+    result = {ERRORS_KEY: has_error}
+    if has_error:
+        error_json = {}
+        error_json[SOURCE_KEY] = source_json
+        error_json[ERRORS_KEY] = invalid_field_obj
+        result[RESULT_KEY] = error_json
+    return result
```
chambridge: Couldn't we just return a tuple of the results and errors instead of a dictionary with keys?

quipucords/api/fact/util.py
```diff
@@ -0,0 +1,183 @@
...
+    invalid_sources = []
+    for source_json in sources_json:
+        result = _validate_source_json(source_json)
+        if result[ERRORS_KEY]:
+            invalid_sources.append(result[RESULT_KEY])
+        else:
+            valid_sources.append(source_json)
+
+    has_errors = len(invalid_sources) > 0
+    return {ERRORS_KEY: has_errors,
```
chambridge: Couldn't we just return a tuple of the valid and invalid sources?

quipucords/api/fact/util.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+def validate_fact_collection_json(fact_collection_json):
+    """Validate fact_collection field.
+
+    :param fact_collection_json: dict representing a fact collection
+    :returns: dict with bool indicating error and result.
+    """
+    if not fact_collection_json.get(SOURCES_KEY):
+        return {ERRORS_KEY: True,
```
chambridge: Couldn't we return a tuple instead of a keyed dictionary?

kholdaway: 👍

quipucords/api/fact/util.py
kholdaway: 👍

quipucords/api/fact/util.py
kholdaway: 👍

PR #379, opened 2018-01-08
Port redhat_release role from rho to quipucords

### LINKED ISSUE #360: Merge redhat_release fingerprints from rho to sonar
Part of #307 .

infinitewarp:
This should run 3x faster than the corresponding role in rho because this
implementation is deduplicated and collects its information from rpm only
once instead of using three separate rpm commands.

Closes #360.

### DIFF
--- a/quipucords/scanner/network/inspect.py
+++ b/quipucords/scanner/network/inspect.py
@@ -139,9 +139,22 @@ class InspectTaskRunner(ScanTaskRunner):
 
         :returns: An array of dictionaries of facts
         """
-        roles = ['check_dependencies', 'connection', 'cpu', 'date',
-                 'dmi', 'etc_release', 'file_contents', 'jboss_eap',
-                 'ifconfig', 'uname', 'virt', 'virt_what', 'host_done']
+        roles = [
+            'check_dependencies',
+            'connection',
+            'cpu',
+            'date',
+            'dmi',
+            'etc_release',
+            'file_contents',
+            'jboss_eap',
+            'ifconfig',
+            'redhat_release',
+            'uname',
+            'virt',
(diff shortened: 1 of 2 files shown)


quipucords/scanner/network/inspect.py
```diff
@@ -139,9 +139,22 @@ def inspect_scan(self):
 
         :returns: An array of dictionaries of facts
         """
-        roles = ['check_dependencies', 'connection', 'cpu', 'date',
-                 'dmi', 'etc_release', 'file_contents', 'jboss_eap',
-                 'ifconfig', 'uname', 'virt', 'virt_what', 'host_done']
+        roles = [
```
infinitewarp: FWIW, I exploded this into multiple lines because otherwise we're going to thrash the commit history as we add to the list.

roles/redhat_release/tasks/main.yml
```diff
@@ -0,0 +1,23 @@
...
+  set_fact:
+    redhat_release_name: 'N/A (rpm not found)'
+    redhat_release_version: 'N/A (rpm not found)'
+    redhat_release_release: 'N/A (rpm not found)'
+
+- name: gather internal_redhat_release fact
+  raw: rpm -q --queryformat "%{NAME}\n%{VERSION}\n%{RELEASE}\n" --whatprovides redhat-release
+  register: internal_redhat_release
+  ignore_errors: yes
+  when: 'internal_have_rpm'
```
infinitewarp: Is it OK to rely on `internal_have_rpm` that is defined in another role (specifically, `check_dependencies`)? This _works_ but I believe only because we are already including the other role before this one.

PR #383, opened 2018-01-09
Add missing tasks to check_dependencies role

### LINKED ISSUE #350: Merge check_dependencies fingerprints from rho to sonar
Part of #307 .

ilyavw:
Closes #350

### DIFF
--- a/roles/check_dependencies/tasks/main.yml
+++ b/roles/check_dependencies/tasks/main.yml
@@ -90,6 +90,15 @@
   set_fact:
     internal_have_systemctl: "{{ internal_have_systemctl.rc == 0 }}"
 
+- name: gather internal_have_chkconfig
+  raw: command -v chkconfig
+  register: internal_have_chkconfig
+  ignore_errors: yes
+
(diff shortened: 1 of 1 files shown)


roles/check_dependencies/tasks/main.yml
```diff
@@ -81,6 +81,15 @@
   set_fact:
     internal_have_locate: "{{ internal_have_locate_cmd.rc == 0 }}"
 
+- name: gather internal_have_systemctl
+  raw: command -v systemctl
```
chambridge: Cool, but makes me wonder if we are missing a similar check for `chkconfig` that should have been used in some of the JBoss product roles? 

^ @noahl

noahl: I agree that we need `chkconfig`. I don't see it in my copy of check_dependencies, so I think it's likely missing.

ilyavw: Do we want to add that as part of this PR?

Issue #340, opened 2018-01-02
RHEL Deployment Report


chambridge:
## Specify type:
- Enhancement

## Description:
The RHEL deployment report will have the following endpoint:

`/reports?source_type=true&os_name=true&os_release=true&cpu_socket_count=true&system_creation_date=true%infrastructure_type=true&vm_host=true&vm_host_socket_count=true&vm_cluster=true&vm_datacenter=true`
which will provide a list of all the systems with (system_identifier, source_type, os_name, os_release, cpu_socket_count, system_creation_date, infrastructure_type, vm_host, vm_host_socket_count, vm cluster, vm_datacenter) tuple

PR #397, opened 2018-01-11
Updates to support group_count and filtering system fingerprints. Closes #339 #340

### LINKED ISSUE #339: Update OS Count endpoint
## Specify type:
- Enhancement

## Description:
The current `/reports` endpoint provides a report of the various operating systems found within an inspection and their count this API will be updated as follows:

`/reports?os_release=true`
which will provide a list of all the systems with (system identifier, os release) tuple

`/reports?group_count=os_release`
which will provide a count of all systems grouped by os release

chambridge:


quipucords/api/fingerprint/serializer.py
```diff
@@ -14,14 +14,13 @@
...
+                                        ModelSerializer)
 from api.models import SystemFingerprint
-from api.common.serializer import NotEmptySerializer
 
 
-class FingerprintSerializer(NotEmptySerializer):
+class FingerprintSerializer(ModelSerializer):
     """Serializer for the Fingerprint model."""
 
     os_name = CharField(required=True, max_length=64)
```
kholdaway: os_version should have required=False based on the model change made above.

chambridge: done

PR #401, opened 2018-01-12
Import eap in processing init so it gets loaded

### LINKED ISSUE #400: None of the EAP processors load to perform post processing
# Opening a new issue
___

## Specify type:
- Bug

- High
___

## Description:
scanner.network.processing.process.Processor requires classes to be loaded.  Python will only load a class if it has been imported somewhere.  As a result none of the scanner.network.processing.process.eap post processors run.
___

### Expected behavior:
eap post processors should be running and performing post processing.

ilyavw:
This makes sure the processors will be loaded and used in the post processing step.

Closes #400


noahl:
The code looks good, but please fix the build errors.

PR #403, opened 2018-01-12
Capture partial results for connection scan

### LINKED ISSUE #301: Capture partial results for a connection scan
# Opening a new issue
___

## Specify type:
- Enhancement

- High

___

## Description:
Capture partial results for a connection scan.  Currently, inspection scans do this.  
___

noahl:
Closes #301 .


quipucords/scanner/network/connect.py
```diff
@@ -30,6 +31,77 @@
...
+                    scan_task=scan_task, source=source)
+                conn_result.save()
+                conn_results.results.add(conn_result)
+                conn_results.save()
+        self.conn_result = conn_result
+
+        hosts = set((host_range.host_range
+                     for host_range in source.hosts.all()))
+        for sys_result in conn_result.systems.all():
+            hosts.discard(sys_result.name)
```
chambridge: I don't think we want to use the previously run connection results to determine where we pick up. I think we want to start over because we don't know how long the scan has been paused for or how long the server has been down.

quipucords/scanner/network/connect.py
```diff
@@ -30,6 +31,77 @@
...
+        self._remaining_hosts = hosts
+
+        if scan_task.systems_count is None:
+            scan_task.systems_count = len(hosts)
```
chambridge: I don't think this is the number of hosts. Just the number of entries provided. If its a range it will not capture the number in the range.
Below is the code from `rho` that pulled in the Ansible info to determine sys count: https://github.com/quipucords/rho/blob/master/rho/host_discovery.py#L127

quipucords/scanner/network/connect.py
noahl: Done. Thanks for explaining!

quipucords/scanner/network/connect.py
noahl: Fixed, thank you!

noahl:
@chambridge I just pushed a new commit that I think addresses your previous comments. Please take another look.

chambridge:
Approved. But coverage dropped 1% :-(

noahl:
@chambridge thank you! I went ahead and pushed another test. Would you take a quick look at that before I merge? Everything else is the same.

chambridge:
LGTM!

PR #409, opened 2018-01-13
Issues/358

abaiken:
I think that I have ported over the jboss_fuse_on_karaf role from rho to sonar. The one thing that can definitely be done better is that I copied the InitLineFinder() in karaf.py from eap.py and it would be better to create a common one in the processing directory that both can pull from. However, when I tried to do this I got a little confused because I wasn't sure how to pass the class an argument without doing an __init__ and things got a little weird so I undid it.


quipucords/scanner/network/processing/karaf.py
```diff
@@ -0,0 +1,114 @@
+# Copyright (c) 2018 Red Hat, Inc.
+#
+# This software is licensed to you under the GNU General Public License,
+# version 3 (GPLv3). There is NO WARRANTY for this software, express or
+# implied, including the implied warranties of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv3
+# along with this software; if not, see
+# https://www.gnu.org/licenses/gpl-3.0.txt.
+
+"""Initial processing of raw shell output from Ansible commands."""
```
chambridge: Maybe a more specific module comment? Seems like this is copy and paste. I'm guessing the eap.y has a similar  comment.

Issue #434, opened 2018-01-15
UI React front-end: API interaction, unit testing and view development


cdcabrera:
## Specify type:
- Enhancement
- UI

## Description:
- [x] About dialog/modal, #450 
- [x] App updates ... bug fixes #513 
- [x] Basic unit test(s) related to Redux, #447 
- [x] Basic unit test(s) related to Sources view
  - evaluated components as Jest snapshots associated with sprint goals
- [x] Basic unit test(s) related to Scans view
  - evaluated components as Jest snapshots associated with sprint goals
- [x] Build updates; General, Mock API, SCSS, #462 #471 #508  ... bug fixes #466 
- [x] Component additions; Toast Notifications, #469 
- [x] Credentials components, and view implementation, #500 #513 #516 
- [x] Login page, initial Django templates and basic routes, #518 
- [ ] ReadMe install updates #535 
- [x] Redux API states for endpoints, #437 #447, #463 
- [x] Routing updates to reflect Django paths, #442
- [x] Scans components, and view implementation, #446 #470 #513 #516 
- [x] Sources components, and updates #521

PR #437, opened 2018-01-15
UI Redux boilerplate for base API calls

cdcabrera:
Quick boilerplate Redux setup for consistency across credentials, sources, and scans API calls.

Updates #434


client/src/redux/actions/scansActions.js
```diff
@@ -1,22 +1,35 @@
...
   type: types.LOAD_SCANS_SUCCESS,
   data
 });
 
 const getScans = () => {
   return function(dispatch) {
+    dispatch(scansLoading(true));
     return scansApi
-      .getScans()
+      .getSources()
```
jeff-phillips-18: Should still be `getScans()`

cdcabrera: good catch!

cdcabrera: looks like creds is the same... fixing that one too.

PR #443, opened 2018-01-16
Add post-processing to check for fuse. Closes #355

### LINKED ISSUE #355: Merge fuse fingerprints from rho to sonar
Part of #307 .

infinitewarp:
Per the discussion in #355, this change just ports over the post-processing in rho that detects the presence of fuse.


infinitewarp:
@adberglund yes
https://github.com/quipucords/quipucords/blob/master/roles/jboss_eap/tasks/main.yml#L73

(and I've run it a few ways locally to verify it is set correctly)

PR #444, opened 2018-01-16
Add satellite type support to sources along with source options. Closes #411.

### LINKED ISSUE #411: Update sources CLI/API to support Satellite sources type
## Specify type:
- Enhancement

## Description:
Sources currently supports `network` and `vcenter` types. Need to add support for the source type `satellite`. The current options of a single host and  a single credential appear to be sufficient for access/inspection.

- Update Swagger documentation with new supported type.
- Update man page documenation with new supported type.
- Update model with `satellite` choice.
- Add new test cases creating satellite source. Like vcenter, only a single host, and a single credential. The credential must also be satellite type.

chambridge:


qpc/source/edit.py
```diff
@@ -56,6 +56,11 @@ def __init__(self, subparsers):
         self.parser.add_argument('--port', dest='port',
                                  metavar='PORT', type=validate_port,
                                  help=_(messages.SOURCE_PORT_HELP))
+        self.parser.add_argument('--satellite_version',
```
kdelee: I think English language learners and myself would rejoice if we could refer to Satellite as "sat" but I understand that may not be clear enough.

quipucords/api/source/tests_source.py
```diff
@@ -264,13 +285,38 @@ def test_create_too_many_creds(self):
...
+
     def test_create_req_host(self):
         """A vcenter source must have an host."""
         self.create_expect_400(
             {'name': 'source1',
              'source_type': Source.VCENTER_SOURCE_TYPE,
              'credentials': [self.vc_cred_for_upload]})
 
+    def test_sat_req_host(self):
+        """A satellite source must have an host."""
```
kdelee: ` """A satellite source must have a host."""`

quipucords/api/source/tests_source.py
```diff
@@ -279,6 +325,14 @@ def test_create_vc_with_hosts(self):
              'hosts': ['1.2.3.4', '1.2.3.5'],
              'credentials': [self.vc_cred_for_upload]})
 
+    def test_create_sat_with_hosts(self):
+        """A satellite source cannot have a host."""
```
kdelee: Is this supposed to read ` """A satellite source must not have multiple hosts."""` ?
Looks like same doc string is in above vcenter test

PR #446, opened 2018-01-17
UI: Add Scans view

jeff-phillips-18:
Add initial scans view with empty state, filtering and sorting. Will adjust as we get more info on mock and data from the API.

Updated title for UI as well.


client/src/components/scans/scans.js
```diff
@@ -1,16 +1,187 @@
...
+    return scans.filter(item => {
+      return this.matchesFilters(item, activeFilters);
+    });
+  }
+
+  sortScans(items) {
+    const { sortType, sortAscending } = this.props;
+
+    let sortId = sortType ? sortType.id : 'name';
+    console.log(sortId);
```
cdcabrera: Left over ```console.log...```. Also looks like ```name``` is the default filter & sort options... both disappear they're clicked.

screen shot 2018-01-17 at 1 34 03 pm

client/src/components/sources/sourceConstants.js
```diff
@@ -16,3 +16,23 @@ export const SourceFilterFields = [
     ]
   }
 ];
+
+const SourceSortFields = [
```
cdcabrera: Note: We should probably come up with some casing guidelines for bits other than classes. I've been uppercasing (yelling) my constants, may be able to add that to Prettier/Eslint

jeff-phillips-18: I don't mind uppercase for simple constants. Not sure I'd want to see it here.

client/src/components/scans/scans.js
jeff-phillips-18: Fixed it up best I could for now. Treating id like name...

PR #447, opened 2018-01-17
UI Expand API interaction

cdcabrera:
### Updates
- [x] Add basic reducer and action unit tests
- [x] Centralize constants for easier import
- [x] Expose additional scans, sources, facts, reports, and credentials endpoint methods. 
- [x] Package JSON updates for unit testing due to issue with Jest packaged with Create React App.
- [x] Package JSON update for PF React 1.0
- [ ] ~~Process.env wrapper~~ Follow up PR
- [x] Rename and expand constants for scans, sources and credentials to reflect API method endpoints.
- [x] Rename files under ```services``` to use ```[API_ENDPOINT]Service.js``` syntax

### Notes
- To run the unit tests either run the standard ```$ npm test``` or ```$ npm run test:unit```.
- Snapshots not checked in this round

updates #434


client/src/services/credentialsApi.js
```diff
@@ -1,6 +1,60 @@
+import jquery from 'jquery';
+
 class CredentialsApi {
-  static getCredentials() {
-    return fetch(process.env.REACT_APP_CREDENTIALS_API).then(response => {
+  static addCredential(data = {}) {
+    return fetch(process.env.REACT_APP_CREDENTIALS_API, {
```
jeff-phillips-18: Wondering if we should have a wrapper for process.env.<API>. 
`ApiServer.CREDENTIALS_API` would read nicer.

jeff-phillips-18:
Thinking we could wrap all the redux action constants as well in a reduxConstants file. Like:
```
import * as CredentialActions from './credentialsConstants';
<...>

const ReduxActionTypes = {
  Credentials: CredentialActions,
  <...>
};

export { ReduxActionTypes, CredentialActions, <...> };
```

cdcabrera:
Yeah, I agree on the wrapper, little more agnostic. I'll go ahead and add it in, easier to do it now since we're adding a bunch. 

Was also curious if under the ```services``` directory we wanted to rename the files from ```[API_ENDPOINT]Api.js``` to ```[API_ENDPOINT]Service.js```? Or we could use that as a distinction between other non-api related services...

jeff-phillips-18:
I prefer Service over API

cdcabrera:
@jeff-phillips-18 , created an additional commit for centralizing ```redux/actions```.  Left it in a separate branch for now but can pull it into this PR easy enough:
- https://github.com/cdcabrera/quipucords/commit/c144579fc895f09b238c1ae569aa3046d9bf5b88

client/src/redux/constants/credentialsConstants.js
```diff
@@ -1,9 +1,37 @@
...
+const GET_CREDENTIAL_LOADING = 'GET_CREDENTIAL_LOADING';
+const GET_CREDENTIAL_SUCCESS = 'GET_CREDENTIAL_SUCCESS';
+
+const GET_CREDENTIALS_ERROR = 'GET_CREDENTIALS_ERROR';
+const GET_CREDENTIALS_LOADING = 'GET_CREDENTIALS_LOADING';
+const GET_CREDENTIALS_SUCCESS = 'GET_CREDENTIALS_SUCCESS';
+
+const UPDATE_CREDENTIAL_ERROR = 'UPDATE_CREDENTIAL_ERROR';
+const UPDATE_CREDENTIAL_LOADING = 'UPDATE_CREDENTIAL_LOADING';
+const UPDATE_CREDENTIAL_SUCCESS = 'UPDATE_CREDENTIAL_SUCCESS';
```
jeff-phillips-18: The naming here is a bit confusing. the `ADD` constants seem like I'm adding something rather than updating the ADD call results.

I'm wondering if the actions should simplify and just have CREDENTIAL_ERROR, CREDENTIALS_LOADING, CREDENTIALS_SUCCESS

Then the action can have more info in it as to which call it was for? Some of these constant defs get to be a rather long list (like scans 😮 )

You can leave as is for now and we can address later.

client/src/redux/actions/credentialsActions.js
```diff
@@ -1,40 +1,40 @@
-import * as types from '../constants/credentialsConstants';
-import credentialsApi from '../../services/credentialsApi';
+import { credentialsTypes as types } from '../constants';
```
jeff-phillips-18: I'd rather not change this from credentialsTypes to types here. More readable below if we leave it as credentialsTypes, imo.

client/src/redux/constants/credentialsConstants.js
cdcabrera: Indeed they do get long in name and the sheer number of them. 

I'm possibly up for trying to handle more in the actions themselves which, agreed, for sure would need an additional check if they're not broken out here. Though either way somethings probably going to get lengthy, if it's an additional switch/label we pass to, or more likely a value we pass back to the subscriber, that may work... like ```actionType: 'error'``` and we simply reduce the constants to... though this could also be in "addition too" 

```js
// within a component
import { credentialTypes } from '../../redux/constants';
...
Store.dispatch({ type: credentialTypes.ADD});
...
```

As a start in a follow up PR can look at something as simple as removing "CREDENTIAL" from the name itself and making sure to use the namespacing when the constants are imported... like.

```js
// within an Action...
import { credentialTypes } from '../constants';
...
Store.dispatch({ type: credentialTypes.ADD_ERROR });
...
```

The nice part is, it's already lengthy, deleting/removing things feels pretty satisfying when it comes around =)

client/src/redux/actions/credentialsActions.js
cdcabrera: We can leave, sure, I can do a replace on those. Probably should make sure to change them across the board. Still have all of the commits so cannn do!

jeff-phillips-18:
I'm ok with a follow-up PR to address the issues and the process.env wrapper

PR #462, opened 2018-01-19
Let's get Sassy! (in the client ui).

jeff-phillips-18:


jeff-phillips-18:
Updated to use `run-s` in the build command to ensure there is no race condition.

jeff-phillips-18:
Updated remaining css files to Sass. Added comment for file name in each sass file, this helps when debugging the css.

PR #469, opened 2018-01-22
Add toast notifications capability

jeff-phillips-18:


jeff-phillips-18:
Updated per comments above

PR #470, opened 2018-01-22
Updates to Sources/Scans UI views

jeff-phillips-18:
updates #434


client/src/components/app.test.js
```diff
@@ -1,11 +1,21 @@
...
 
-    ReactDOM.render(<App />, div);
+    ReactDOM.render(
+      <Provider store={store}>
+        <Router>
+          <App />
+        </Router>
+      </Provider>,
+      div
+    );
```
cdcabrera: This took a bit of search... so part of the issue with these components is we have an async aspect, hence one of the problems with Fetch (not the non-existent ```finally``` method). Adding this ```React.unmountComponentAtNode(div)``` after line 19 ensures the initial mount and immediate unmount of the component for a quick test, soooo like so:

```js
ReactDOM.render(
  <Provider store={store}>
    <Router>
      <App />
    </Router>
  </Provider>,
  div
);

ReactDOM.unmountComponentAtNode(div);
```
If we combine this with converting the ```finally``` to a ```then``` or simply removing them. The tests should all pass with just a general console warning about the redirect... which we can address later.

Found this here... https://github.com/facebook/create-react-app/issues/3482#issuecomment-347347454

client/src/components/scans/scans.js
```diff
@@ -102,18 +141,81 @@ class Scans extends React.Component {
...
+
+  importSources() {
+    Store.dispatch({
+      type: toastNotificationTypes.TOAST_ADD,
+      alertType: 'error',
+      header: 'NYI',
+      message: 'Adding sources is not yet implemented'
+    });
+  }
+
```
cdcabrera: Good first round!

Probably a concept we have to evolve since it's hard to visualize, combining these ```dispatches``` into grouped actions ... so in the end, maybe something like
```js
runScans()  {
  Store.dispatch({
    type: scansTypes.DO_ALL_THINGS_FOR_RUN_SCANS,
    ...
```

PR #471, opened 2018-01-23
UI Unit test additions, fixes, and build updates

cdcabrera:
* Unit tests added for reducers
* Updated app unit test to do a quick render.
* Quick fix for limited support of the "finally" Promise method as related to Fetch
* Service worker moved
* Build and app dependency updates
* Convert content and vertical nav component import

updates #434


client/src/redux/actions/credentialsActions.js
```diff
@@ -24,11 +24,11 @@ const getCredentials = () => {
       .getCredentials()
       .then(success => {
         dispatch(getCredentialsSuccess(success));
+        dispatch(getCredentialsLoading(false));
       })
       .catch(error => {
         dispatch(getCredentialsError(true, error.message));
```
jeff-phillips-18: Should `dispatch(getCredentialsLoading(false));` here as well.

client/src/redux/actions/factsActions.js
```diff
@@ -24,11 +24,11 @@ const addFacts = () => {
       .addFacts()
       .then(success => {
         dispatch(addFactsSuccess(success));
+        dispatch(addFactsLoading(false));
       })
       .catch(error => {
         dispatch(addFactsError(true, error.message));
-      })
```
jeff-phillips-18: Ditto.

client/src/redux/actions/reportsActions.js
```diff
@@ -24,11 +24,11 @@ const getReports = () => {
       .getReports()
       .then(success => {
         dispatch(getReportsSuccess(success));
+        dispatch(getReportsLoading(false));
       })
       .catch(error => {
         dispatch(getReportsError(true, error.message));
```
jeff-phillips-18: and here...

client/src/redux/actions/sourcesActions.js
```diff
@@ -24,11 +24,11 @@ const getSources = () => {
       .getSources()
       .then(success => {
         dispatch(getSourcesSuccess(success));
+        dispatch(getSourcesLoading(false));
       })
       .catch(error => {
         dispatch(getSourcesError(true, error.message));
```
jeff-phillips-18: enough said.

client/src/redux/actions/credentialsActions.js
cdcabrera: Indeed +1

PR #488, opened 2018-01-24
Add capability to skip requests for a system that already has an inspection result. Closes #487.

### LINKED ISSUE #487: Satellite inspect scan should check and skip any systems that were previously covered due to a pause
## Specify type:
- Enhancement

## Description:
Current inspection task will restart and add systems that could have already been covered leading to duplicates and a longer scan than needed.  Should check first before getting host details.

chambridge:


quipucords/scanner/satellite/six.py
```diff
@@ -389,6 +389,17 @@ def host_details(self, org_id, host_id, host_name):
         :returns: dictionary of host details
         """
         details = {}
+        sys_result = None
+        sys_results = self.inspect_result.systems.all()
```
kholdaway: Why do systems.all() instead of doing a filter() here?

quipucords/scanner/satellite/six.py
```diff
@@ -481,6 +492,17 @@ def host_details(self, host_id, host_name):
...
```
kholdaway: Same.  Seems like you can just filter here.

quipucords/scanner/vcenter/inspect.py
```diff
@@ -71,6 +73,18 @@ def run(self):
                 self.connect_scan_task.id)
             return ScanTask.FAILED
 
+        # Prep inspect result
```
kholdaway: Probably should make this method atomic since it is doing several db transactions

quipucords/scanner/vcenter/inspect.py
```diff
@@ -171,8 +176,19 @@ def recurse_datacenter(self, vcenter):
                     for host in hosts:  # Iterate through Hosts in the Cluster
                         vms = host.vm
                         for virtual_machine in vms:
-                            self.get_vm_info(data_center_name, cluster_name,
-                                             host, virtual_machine)
+                            vm_name = virtual_machine.summary.config.name
+                            sys_results = self.inspect_result.systems.all()
```
kholdaway: Just add filter here.

quipucords/scanner/satellite/six.py
chambridge: done.

quipucords/scanner/satellite/six.py
chambridge: done.

quipucords/scanner/vcenter/inspect.py
chambridge: done

quipucords/scanner/vcenter/inspect.py
chambridge: done

PR #506, opened 2018-01-25
Handle interrupted scans that have completed everything but fingerprint processing

### LINKED ISSUE #377: Sonar server should look for scan that have persisted results that have not been processed by engine
# Opening a new issue
___

## Specify type:
- Enhancement

## Description:
It is possible that a scan will complete, store results, and then be interrupted before the fingerprint engine has processed the results.  Fact collections have state which can either be persisted or completed.  The sonar startup code should look for scans that are persisted and make sure the engine completes the processing.
___

kholdaway:
- Moved fact_collection_json into the fact_collection model instead of an external file.  This impacts the job, util, and fact view
- ScanJobs now reflect the status of the fingerprint processing.  Interrupted job logic will now cover fingerprint processing.  In other words, a scanjob is not complete until the fingerprint processing is complete.

Closes #377

### DIFF
--- a/quipucords/fingerprinter/__init__.py
+++ b/quipucords/fingerprinter/__init__.py
@@ -84,7 +84,7 @@ def process_fact_collection(sender, instance, **kwargs):
 def _process_sources(fact_collection):
     """Process facts and convert to fingerprints.
 
-    :param json_fact_collection: Collected raw facts for all sources
+    :param fact_collection: FactCollection containing raw facts
     :returns: list of fingerprints for all systems (all scans)
     """
     network_fingerprints = []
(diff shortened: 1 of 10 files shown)


quipucords/fingerprinter/__init__.py
```diff
@@ -84,19 +81,18 @@ def process_fact_collection(sender, instance, **kwargs):
     instance.save()
 
 
-def _process_sources(raw_facts):
+def _process_sources(fact_collection):
     """Process facts and convert to fingerprints.
 
-    :param raw_facts: Collected raw facts for all sources
+    :param json_fact_collection: Collected raw facts for all sources
```
abaiken: param name in description doesn't match param

kholdaway: Good catch.  Fixed.

Issue #422, opened 2018-01-15
Use Scan options in Satellite inspection for optional product collection


chambridge:
## Specify type:
- Enhancement

## Description:
Currently every scan collects data on all facts for all products. We need to extend the scan capabilities to be able to capture a limitation on the facts to be collected.

### Inspection Scans
- Satellite inspection scan will execute APIs to collect optional product facts

chambridge:
This item was opened before the implementation of the Satellite inspection task. Based on how it works, there isn't a way to avoid collecting data on specific products. It essentially collects all entitlements. As such I'm closing this item.

Issue #421, opened 2018-01-15
Use Scan options in Network inspection for optional product collection


chambridge:
## Specify type:
- Enhancement

## Description:
Currently every scan collects data on all facts for all products. We need to extend the scan capabilities to be able to capture a limitation on the facts to be collected.

### Inspection Scans
- Network inspection scan will use extra_vars to pass enabling optional product scans

abaiken:
Closed by https://github.com/quipucords/quipucords/pull/517

Issue #420, opened 2018-01-15
Enable Scan options in CLI for optional product collection


chambridge:
## Specify type:
- Enhancement

## Description:
Currently every scan collects data on all facts for all products. We need to extend the scan capabilities to be able to capture a limitation on the facts to be collected.

### CLI Updates
- Add flags to scan start to enable optional products (`--optional-products  jboss-eap jboss-fuse`)
- CLI will transform flags into options object for post
- Update man pages with new options and detail optional product values

abaiken:
Closed by https://github.com/quipucords/quipucords/pull/517

Issue #419, opened 2018-01-15
Enable Scan options in API for optional product collection


chambridge:
## Specify type:
- Enhancement

## Description:
Currently every scan collects data on all facts for all products. We need to extend the scan capabilities to be able to capture a limitation on the facts to be collected.

### Model updates
Create `ScanOptions` model to detail the product capture limitations:
- JBoss EAP (True/False; default True)
- JBoss Fuse (True/False; default True, if true must also collect EAP)
- JBoss BRMS (True/False; default True, if true must also collect EAP)

### API updates
Post with scan options
Get should display scan options
- Update Swagger with scan options object

abaiken:
Closed by https://github.com/quipucords/quipucords/pull/517

PR #517, opened 2018-01-27
Issues/419

abaiken:
I believe that this solves issues #419, #420, #421 but not #422 (yet). If someone would look it over to give feedback that would be awesome!


quipucords/api/scanjob/model.py
```diff
@@ -274,3 +287,40 @@ def validate_status_change(self, target_status, valid_current_status):
                          target_status, self.status)
             return True
         return False
+
+    def create_product_status_dict(self):
+        """Construct a dictionary based on the disabled products.
```
chambridge: Feels like this method needs a better name/description. Took me a while to understand what product status was.

quipucords/api/scanjob/tests_scanjob.py
```diff
@@ -215,7 +215,10 @@ def test_pause_restart_task(self, start_scan):
     def test_successful_create(self, start_scan):
         """A valid create request should succeed."""
         data = {'sources': [self.source.id],
-                'scan_type': ScanTask.SCAN_TYPE_CONNECT}
+                'scan_type': ScanTask.SCAN_TYPE_CONNECT,
+                'options': {'disable_optional_products': {'jboss_eap': True,
```
chambridge: I feel like if null had been the default you wouldn't have needed to update so many tests. Just create new ones to test the flow with this value.

quipucords/api/scanjob/model.py
```diff
@@ -274,3 +287,40 @@ def validate_status_change(self, target_status, valid_current_status):
...
+        """
+        # Grab the optional products status dict and create
+        # a default dict (all products default to True)
+        product_status = self.get_optional_products_args()
+        product_default = {self.OPT_JBOSS_EAP: True,
+                           self.OPT_JBOSS_FUSE: True,
+                           self.OPT_JBOSS_BRMS: True}
+
+        # If specified, turn off fact collection for fuse
+        if not product_status[self.ARG_JBOSS_FUSE]:
```
chambridge: You should generally use the`get` method to avoid a `KeyError` (i.e `product_status.get(self.ARG_JBOSS_FUSE)`, it will return `None` if its not set). Same comment for the lines below.

quipucords/api/scanjob/model.py
```diff
@@ -274,3 +287,40 @@ def validate_status_change(self, target_status, valid_current_status):
...
+                (not product_default[self.OPT_JBOSS_BRMS]):
+            product_default[self.OPT_JBOSS_EAP] = False
+
+        return product_default
+
+    def get_optional_products_args(self):
+        """Access disabled_optional_products as a dict instead of a string.
+
+        :returns: python dict containing the status of optional products
+        """
```
chambridge: With my suggestion of just storing null you could just add a step here that checks for `None` and returns and empty dictionary so your `create_product_status_dict()` method doesn't need to change.

quipucords/scanner/network/connect.py
```diff
@@ -142,6 +142,7 @@ def run_with_result_store(self, result_store):
         serializer = SourceSerializer(self.scan_task.source)
         source = serializer.data
 
+        optional_products_status = self.scan_job.create_product_status_dict()
```
chambridge: This is unnecessary as the connection flow doesn't run the playbook, it just checks for authentication.

quipucords/scanner/network/connect.py
```diff
@@ -157,7 +158,7 @@ def run_with_result_store(self, result_store):
             callback = ConnectResultCallback(result_store, credential)
             try:
                 connect(remaining_hosts, callback, cred_data,
-                        connection_port, forks=forks)
+                        connection_port, optional_products_status, forks=forks)
```
chambridge: not needed.

quipucords/scanner/network/connect.py
```diff
@@ -183,13 +184,17 @@ def run_with_result_store(self, result_store):
         return ScanTask.COMPLETED
 
 
-def connect(hosts, callback, credential, connection_port, forks=50):
+# pylint: disable=too-many-arguments
+def connect(hosts, callback, credential, connection_port,
+            optional_products_status, forks=50):
```
chambridge: Not needed.

quipucords/scanner/network/inspect.py
```diff
@@ -228,7 +230,7 @@ def construct_scan_inventory(hosts, connection_port, concurrency_count):
     :param hosts: The collection of hosts/credential tuples
     :param connection_port: The connection port
     :param concurrency_count: The number of concurrent scans
-    :returns: A dictionary of the ansible invetory
+    :returns: A dictionary of the ansible inventory
```
chambridge: 👍

quipucords/scanner/network/utils.py
```diff
@@ -89,10 +89,12 @@ def write_inventory(inventory):
     return write_to_yaml(inventory)
 
 
-def create_ansible_objects(inventory_file, forks=50):
+def create_ansible_objects(inventory_file, optional_products_status, forks=50):
```
chambridge: lets just call this `extra_vars` instead of something specific to this user story for use in the future.

quipucords/scanner/network/connect.py
```diff
@@ -204,7 +209,8 @@ def connect(hosts, callback, credential, connection_port, forks=50):
                                       'args': parse_kv('echo "Hello"')}}]}
 
     _handle_ssh_passphrase(credential)
-    result = run_playbook(inventory_file, callback, playbook, forks=forks)
+    result = run_playbook(inventory_file, callback, playbook,
+                          optional_products_status, forks=forks)
```
chambridge: I'd prefer we reference it as `extra_vars` to keep it generic.

```
result = run_playbook(inventory_file, callback, playbook,
                          extra_vars=optional_products_status, forks=forks)
```

quipucords/scanner/network/utils.py
```diff
@@ -109,26 +111,30 @@ def create_ansible_objects(inventory_file, forks=50):
...
                           host_list=inventory_file)
+    variable_manager.extra_vars = optional_products_status
     variable_manager.set_inventory(inventory)
 
     return options, variable_manager, loader, inventory
 
 
-def run_playbook(inventory_file, callback, play, forks=50):
+def run_playbook(inventory_file, callback, play,
+                 optional_products_status, forks=50):
```
chambridge: optional_products_status --> extra_vars

docs/swagger.yml
```diff
@@ -931,6 +931,15 @@ definitions:
         type: "integer"
         format: "int64"
         description: "The maximum concurrent host scans. default is 50."
+      disable_optional_products:
+        type: "array"
```
chambridge: I don't think this is correct based on the values below.  Looks like this should be its own defined object as its not an array list due to design changes.

quipucords/api/scanjob/serializer.py
```diff
@@ -26,12 +28,14 @@ class ScanOptionsSerializer(NotEmptySerializer):
     """Serializer for the ScanOptions model."""
 
     max_concurrency = IntegerField(required=False, min_value=1, default=50)
+    disable_optional_products = CustomJSONField(required=False)
```
chambridge: I believe we need a good amount of validation logic added for this field. Since this is just taking JSON right now any valid JSON could be stored to the API. We need to add code here to restrict the appropriate JSON to an object with specific keys and values of True/False.

abaiken: Is there an example of this/something close to this? I am not sure exactly how to go about it (new to JSON).

qpc/messages.py
```diff
@@ -116,6 +116,10 @@
 SCAN_RESTARTED = 'Scan "%s" restarted.'
 SCAN_CANCELED = 'Scan "%s" canceled.'
 
+DISABLE_OPT_PRODUCTS_HELP = 'The product inspection exclusions. '\
+    'Contains the list of products to exclude from inspection. '\
+    'Valid choices are jboss_eap, jboss_fuse, and jboss_brms.'
+
```
abaiken: @jlprevatt Would you look at the message that I added and give feedback/let me know if changes are needed? Thank you!

quipucords/api/scanjob/model.py
```diff
@@ -288,3 +299,62 @@ def validate_status_change(self, target_status, valid_current_status):
...
+            raise ValidationError(_(messages.SJ_EXTRA_VARS_DICT))
+        for key in extra_vars:
+            if not isinstance(extra_vars[key], bool):
+                raise ValidationError(_(messages.SJ_EXTRA_VARS_BOOL))
+            elif key not in [ScanJob.JBOSS_EAP,
+                             ScanJob.JBOSS_BRMS,
+                             ScanJob.JBOSS_FUSE]:
+                raise ValidationError(_(messages.SJ_EXTRA_VARS_KEY))
+        return extra_vars
+
```
abaiken: @chambridge here is where I do validation. I put it in the model because I felt that it could be used by the function below (line 329) and it seemed to make sense there.

docs/source/man.rst
```diff
@@ -302,7 +302,7 @@ Scanning
 Use the ``qpc scan start`` command to run connect and inspection scans on the sources. This command scans all of the host names or IP addresses that are defined in the supplied source, and then writes the report information to a comma separated values (CSV) file. Note: Any ssh-agent connection setup for a target host '
               'will be used as a fallback if it exists.
 
-**qpc scan start --sources=** *source_list* **[--max-concurrency=** *concurrency* **]**
+**qpc scan start --sources=** *source_list* **[--max-concurrency=** *concurrency* **]** **--disable-optional-products=** *products_list*
```
abaiken: @chambridge When I added disable optional products to line 305, do I need to add the options instead of just product list ie. `*[jboss_eap, jboss_fuse, jboss_brms]*` or do I leave it as `*products_list*` ? Thank you!

qpc/messages.py
jlprevatt: For the help strings that contain a list of possible values for an option, I've tried to settle on the following standard structure for the "values" sentence:
Valid values: jboss_eap, jboss_fuse, jboss_brms.

abaiken: Awesome, I will fix it. Thanks Jamie!

docs/swagger.yml
```diff
@@ -924,22 +924,27 @@ definitions:
...
-        type: "array"
-        description: "The product capture exclusions."
-        items:
-          type: "string"
-          enum:
-            - jboss_eap
-            - jboss_fuse
-            - jboss_brms
+      disabled_optional_products:
+        $ref: "#definitions/DisableOptionalProducts"
```
chambridge: I think you can still add a description with the reference definition.

qpc/scan/start.py
```diff
@@ -116,10 +113,12 @@ def _build_data(self):
             'sources': self.source_ids,
             'scan_type': scan.SCAN_TYPE_INSPECT,
             'options': {
-                'max_concurrency': self.args.max_concurrency,
-                'disable_optional_products':
-                    self._create_optional_product_status_dict()}
+                'max_concurrency': self.args.max_concurrency}
         }
+        disable_optional_products = self._create_optional_product_status_dict()
```
chambridge: small naming suggestion: 
`self._create_optional_product_status_dict()`  -->  `self._get_optional_products()`

quipucords/api/scanjob/model.py
```diff
@@ -302,31 +299,34 @@ def validate_status_change(self, target_status, valid_current_status):
...
         """Construct a dictionary based on the disabled products.
 
         :returns: a dictionary representing the updated collection
-        status of the optional products
+        status of the optional products to be assigned as the extra
+        vars for the ansibile task runner
         """
         # Grab the optional products status dict and create
         # a default dict (all products default to True)
         product_status = self.get_optional_products_args()
```
chambridge: maybe just rename to `self.get_optional_products_args()` --> `self.get_optional_products()`

quipucords/api/scanjob/model.py
```diff
@@ -302,31 +299,34 @@ def validate_status_change(self, target_status, valid_current_status):
             return True
         return False
 
-    def create_product_status_dict(self):
+    def create_extra_vars(self):
```
chambridge: name is ok, maybe just `get_extra_vars()`

quipucords/api/scanjob/model.py
chambridge: I think this code should be in the serializer as by this point in the code you have potentially already stored JSON that is not valid for our use and the Network inspect task may blow up with a `ValidationError` instead of the user seeing the issue on the POST.

chambridge: to be clear though it should be a function of the `ScanOptionsSerializer` class defined as `def validate_disable_optional_products(disable_optional_products):`

quipucords/api/scanjob/tests_scanjob.py
```diff
@@ -457,11 +493,19 @@ def test_details(self, start_scan):
     def test_update_not_allowed(self, start_scan):
         """Completely update a Source."""
         data_discovery = {'sources': [self.source.id],
-                          'scan_type': ScanTask.SCAN_TYPE_CONNECT}
+                          'scan_type': ScanTask.SCAN_TYPE_CONNECT,
+                          'options': {'disable_optional_products':
```
chambridge: should have an update with not a JSON object (i.e. python dictionary) to test the validation error.

quipucords/api/scanjob/tests_scanjob.py
```diff
@@ -259,7 +264,10 @@ def test_create_invalid_srcs_id(self):
     @patch('api.scanjob.view.start_scan', side_effect=dummy_start)
     def test_create_default_host_type(self, start_scan):
         """A valid create request should succeed with defaulted type."""
-        data = {'sources': [self.source.id]}
+        data = {'sources': [self.source.id],
+                'options': {'disable_optional_products': {'jboss_eap': True,
```
chambridge: should have a create with a test of 'disable_optional_products' when its not a JSON object/python dictionary. Could be a string, array list, etc.

docs/source/man.rst
abaiken: @chambridge did you get a chance to look at this? Thank you!

PR #519, opened 2018-01-27
Add status message to ScanJob and ScanTask

### LINKED ISSUE #467: Capture error message in scan task and scan job for failed scan
## Specify type:
- Enhancement

## Description:
Currently scan jobs and scan tasks only contain a status that could result in a failure. Some minimal information should be passed to the user on a failure to help them resolve the issue, even if the minimal message is to review the server logs due to an unknown exception.

kholdaway:
Closes #467


kholdaway:
Are you referring to just the scantask/job or all the error messages from satellite/vcenter/network connect and inspect?  I can externalize these at part of the PR, but most of the messages in this PR were moved from logger.error to a message variable.  I figured we could tackle that as part of the logging cleanup.

chambridge:
I’m just referring to the job/task messages that we expect to present to users via the api or cli.

kholdaway:
I externalized the status messages for ScanJob and ScanTask.

kholdaway:
https://github.com/quipucords/quipucords/issues/528

PR #521, opened 2018-01-27
Add expansion views for list views

jeff-phillips-18:
Updates #434

### DIFF
--- a/client/src/components/scans/scanListItem.js
+++ b/client/src/components/scans/scanListItem.js
@@ -11,6 +11,9 @@ import {
 
 import { bindMethods } from '../../common/helpers';
 
+import { ScanSourceList } from './scanSourceList';
+import { ScanHostsList } from './scanHostList';
+
 class ScanListItem extends React.Component {
   constructor() {
     super();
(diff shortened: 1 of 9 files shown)


client/src/components/scans/scanListItem.js
```diff
@@ -37,8 +40,22 @@ class ScanListItem extends React.Component {
...
+      case 'systemsFailed':
+        return (
+          <ScanHostsList hosts={item.failed_hosts} status="error-circle-o" />
+        );
+      case 'sources':
+        return <ScanSourceList scan={item} />;
+      case 'scans':
+        return <JSONPretty json={item} />;
+      default:
+        return <JSONPretty json={item} />;
```
cdcabrera: Optional: Let the switch cascade down...
```js
case 'scans':
default:
...
```

client/src/components/scans/scans.js
```diff
@@ -49,6 +49,16 @@ class Scans extends React.Component {
...
+
+        // TODO: Get real hosts or wait until expansion?
+        scan.hosts = [];
+        for (let i = 0; i < scan.systems_scanned; i++) {
+          scan.hosts.push('host' + (i + 1));
+        }
+        scan.failed_hosts = [];
+        for (let i = 0; i < scan.systems_failed; i++) {
+          scan.failed_hosts.push('failedHost' + (i + 1));
+        }
```
cdcabrera: Note: Eventually should be part of the initial API response. However, it would be neat to have a background task running after the initial view API call, that's something we should probably figure out anyways.

PR #536, opened 2018-01-29
Timeout individual Ansible tasks

### LINKED ISSUE #484: Make Sonar more robust to ssh hangs
If ssh to a host hangs for any reason, Sonar should time out, record that the host timed out, and keep collecting data from other hosts and generating reports.

noahl:
Provide a custom ssh executable to Ansible that runs regular ssh
inside of 'timeout', and use this to add a timeout to every task. Add
infrastructure for testing this ssh executable by making it time out a
program that will hang forever. Also add a test for SSH crashes.

Closes #484 .

### DIFF
--- a/quipucords/scanner/network/inspect.py
+++ b/quipucords/scanner/network/inspect.py
@@ -10,8 +10,10 @@
 #
 """ScanTask used for network connection discovery."""
 import logging
+import os.path
 from ansible.errors import AnsibleError
 from ansible.executor.task_queue_manager import TaskQueueManager
+from api.credential.serializer import CredentialSerializer
 from api.models import (ScanTask, ConnectionResult,
                         SystemConnectionResult)
 from scanner.task import ScanTaskRunner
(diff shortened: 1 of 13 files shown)


quipucords/scanner/network/inspect.py
```diff
@@ -79,7 +84,24 @@ def run(self):
...
+                raise ScannerException(msg)
+
+            logger.info('Inspect scan task started for %s.',
+                        self.scan_task)
+            log_msg = '%d total connected, %d completed, %d'\
+                ' remaining, and make %d failed hosts'
+            logger.info(log_msg,
+                        num_total, num_completed, num_remaining, num_failed)
+
+            self.inspect_scan(connected)
```
chambridge: What was the point in moving this block of code out of `inspect_scan` and changing the `inspect_scan` signature?  Testability?

chambridge:
Looks like failing tests for the crash simulation.

quipucords/scanner/network/inspect.py
noahl: Yes, testability. Otherwise the test would have been much more convoluted.

noahl:
@chambridge I'll keep working on the failing tests. The tests are failing on Travis and not my machine, and I don't know why, but it means I have a longer than usual turnaround time on testing my code.

PR #540, opened 2018-01-29
Convert Source HostRange to a JSON array

### LINKED ISSUE #529: Convert source host range to CustomJSONField and TextField model
## Specify type:
- Enhancement

## Description:
The code to create a Source object is quite complex. Based on improvements in other portions of the code I suggest we update the Source model to have a `hosts` field that is a `TextField`. In the serializer we can use a `CustomJsonField` and just add some validation that the incoming is an array of strings. From there the existing validation logic should be valid.

kholdaway:
Closes #529


quipucords/api/source/model.py
```diff
@@ -55,35 +56,16 @@ class Source(models.Model):
...
-    # Source also has the field hosts, which is created by the
-    # ForeignKey in HostRange below.
+    hosts = models.TextField(unique=False, null=False)
 
     def __str__(self):
         """Convert to string."""
-        return '{ id:%s, name:%s, type:%s }' %\
-            (self.id, self.name, self.source_type)
+        return '{ id:%s, name:%s, type:%s , hosts:%s}' %\
+            (self.id, self.name, self.source_type, self.hosts)
```
chambridge: I feel like this could potentially be a huge string to print out. What if there are 10K hosts? Might be best to leave it out.

quipucords/api/source/serializer.py
```diff
@@ -123,14 +97,16 @@ def create(self, validated_data):
         source_type = validated_data.get('source_type')
         credentials = validated_data.pop('credentials')
         hosts_data = validated_data.pop('hosts', None)
+        if hosts_data:
+            hosts_list = json.loads(hosts_data)
```
chambridge: I think you are assuming that this value will be a list. I think we need to do validation, where we are checking if `isinstance(data, list)` and throw a validation error if the format is wrong. Also need to check that each of the items in the list is a string.

quipucords/api/source/serializer.py
```diff
@@ -232,24 +206,27 @@ def update(self, instance, validated_data):
         source_type = instance.source_type
         credentials = validated_data.pop('credentials', None)
         hosts_data = validated_data.pop('hosts', None)
+        if hosts_data:
+            hosts_list = json.loads(hosts_data)
```
chambridge: Same comment as above.

quipucords/api/source/serializer.py
```diff
@@ -263,17 +240,18 @@ def update(self, instance, validated_data):
                 SourceSerializer.check_credential_type(source_type,
                                                        credentials[0])
         elif source_type == Source.SATELLITE_SOURCE_TYPE:
-            if not hosts_data and self.partial is False:
+            if not hosts_list and self.partial is False:
+                print('here2')
```
chambridge: I guess `here2`!

quipucords/api/source/serializer.py
```diff
@@ -363,7 +334,8 @@ def validate_name(name):
     @staticmethod
     def validate_hosts(hosts):
         """Make sure the hosts list is present."""
-        if not hosts:
+        hosts_list = json.loads(hosts)
+        if not hosts_list:
             raise ValidationError(_(messages.SOURCE_HOSTS_CANNOT_BE_EMPTY))
```
chambridge: json format validation could occur here, as I believe this will be called before the `create` method.

quipucords/api/source/serializer.py
```diff
@@ -398,31 +370,30 @@ def validate_hosts(hosts):
 
         normalized_hosts = []
         host_errors = []
-        for host in hosts:
+        for host in hosts_list:
```
chambridge: think you could have just done `for host_range in hosts_list:`, since technically the string can still be a range of hosts. But no big deal.

quipucords/api/source/serializer.py
```diff
@@ -460,9 +429,9 @@ def validate_hosts(hosts):
                 # This is an unexpected case. Allow/log for analysis
                 normalized_hosts.append(host)
                 logging.warning('%s did not match a pattern or produce error',
-                                host_range)
+                                host)
         if len(host_errors) is 0:
-            return normalized_hosts
+            return json.dumps(normalized_hosts)
```
chambridge: Is there value in dumping to JSON instead of passing the python object?  Since you just end up turning around and doing a `json.loads` in the `create` method.

quipucords/api/source/tests_source.py
```diff
@@ -449,10 +449,13 @@ def test_retrieve(self):
...
         self.assertEqual(response.status_code, status.HTTP_200_OK)
-        self.assertIn('credentials', response.json())
-        creds = response.json()['credentials']
+        response_json = response.json()
+        self.assertIn('credentials', response_json)
+        creds = response_json['credentials']
 
         self.assertEqual(creds, [self.net_cred_for_response])
+        self.assertIn('hosts', response_json)
+        self.assertEqual(response_json['hosts'][0], '1.2.3.4')
```
chambridge: Would be could to have a test here where we try and create a source with other valid json not just an array of strings. Should have errors for that.

quipucords/api/source/serializer.py
kholdaway: Removed.

quipucords/api/source/model.py
kholdaway: Removed.

quipucords/api/source/serializer.py
kholdaway: check not needed here since it is done in validate_hosts

quipucords/api/source/serializer.py
kholdaway: check not needed here since it is done in validate_hosts

quipucords/api/source/serializer.py
kholdaway: Done.

quipucords/api/source/serializer.py
kholdaway: Agree.

quipucords/api/source/serializer.py
kholdaway: Good point.  Changed.

quipucords/api/source/tests_source.py
kholdaway: Added a couple of tests for invalid data types of hosts array.

quipucords/api/source/serializer.py
```diff
@@ -335,6 +330,15 @@ def validate_name(name):
...
         hosts_list = json.loads(hosts)
+        if not isinstance(hosts_list, list):
+            raise ValidationError(_(messages.SOURCE_HOST_MUST_BE_JSON_ARRAY))
+
+        len_list = len(hosts_list)
+        hosts_list = [
+            string for string in hosts_list if isinstance(string, str)]
+        if len_list != len(hosts_list):
+            raise ValidationError(_(messages.SOURCE_HOST_MUST_BE_JSON_ARRAY))
+
```
chambridge: probably should be after the `if not host_list:` check.

Also why do the comprehension with length check vs just iteration through the objects and throwing a validation error if any single item is not a string?
This would save memory and kick out on the first item hit.

PR #546, opened 2018-01-30
Use slave provided by the OpenStack jenkins plugin

elyezer:
We have slaves now provided by the OpenStack jenkins plugin.


elyezer:
Rebased, will merge after travis being happy

elyezer:
rebased once again, now I won't miss merging

PR #556, opened 2018-01-31
added start/end time to Scan Job/ Scan Task model & serializer

abaiken:
Check that I am grabbing start/end time in the correct place (I am not sure if we still care about end_time if the scan job/task fails but I do grab it just in case.


quipucords/scanner/network/connect.py
```diff
@@ -9,7 +9,9 @@
 # https://www.gnu.org/licenses/gpl-3.0.txt.
 #
 """ScanTask used for network connection discovery."""
+import datetime
 import logging
+from pytz import timezone
```
chambridge: No need for this we should just use UTC time.

quipucords/scanner/network/connect.py
```diff
@@ -9,7 +9,9 @@
 # https://www.gnu.org/licenses/gpl-3.0.txt.
 #
 """ScanTask used for network connection discovery."""
+import datetime
```
chambridge: lets just import the method we need `from datetime.datetime import utcnow`

quipucords/scanner/network/connect.py
```diff
@@ -129,9 +131,11 @@ def __init__(self, scan_job, scan_task, conn_results):
         super().__init__(scan_job, scan_task)
         self.conn_results = conn_results
         self.scan_task = scan_task
+        self.timezone = timezone('EST')
 
     def run(self):
         """Scan network range ang attempt connections."""
+        self.scan_task.start_time = datetime.datetime.now(self.timezone).time()
```
chambridge: Lets keep the full time ... what if the scan takes more than a day.

`self.scan_task.start_time = utcnow()`

quipucords/api/scanjob/model.py
```diff
@@ -70,6 +70,8 @@ class ScanJob(models.Model):
     options = models.ForeignKey(
         ScanOptions, null=True, on_delete=models.CASCADE)
     fact_collection_id = models.IntegerField(null=True)
+    start_time = models.TimeField(null=True)
```
chambridge: Lets use `DateTimeField` as we may need to scan times longer than a day.

quipucords/api/scantasks/model.py
```diff
@@ -63,6 +63,8 @@ class ScanTask(models.Model):
     systems_scanned = models.PositiveIntegerField(null=True)
     systems_failed = models.PositiveIntegerField(null=True)
     sequence_number = models.PositiveIntegerField(null=True)
+    start_time = models.TimeField(null=True)
```
chambridge: Lets use `DateTimeField` as we may need to scan times longer than a day.

quipucords/scanner/network/connect.py
```diff
@@ -181,6 +187,7 @@ def run_with_result_store(self, result_store):
         # Clear cache as results changed
         self.result = None
 
+        self.scan_task.end_time = datetime.datetime.now(self.timezone).time()
```
chambridge: I'd move this logic to here:
https://github.com/quipucords/quipucords/blob/master/quipucords/api/scantasks/model.py#L90
https://github.com/quipucords/quipucords/blob/master/quipucords/api/scantasks/model.py#L114
https://github.com/quipucords/quipucords/blob/master/quipucords/api/scantasks/model.py#L124
https://github.com/quipucords/quipucords/blob/master/quipucords/api/scantasks/model.py#L108

Same for scan job:
https://github.com/quipucords/quipucords/blob/master/quipucords/api/scanjob/model.py#L160
https://github.com/quipucords/quipucords/blob/master/quipucords/api/scanjob/model.py#L224
https://github.com/quipucords/quipucords/blob/master/quipucords/api/scanjob/model.py#L252
https://github.com/quipucords/quipucords/blob/master/quipucords/api/scanjob/model.py#L267

quipucords/api/scanjob/tests_scanjob.py
```diff
@@ -315,6 +315,29 @@ def test_create_invalid_disable_optional_products_val(self):
             'options': {'disable_optional_products':
                         ['Extra vars values must be type boolean.']}})
 
+    def test_create_invalid_start_time(self):
```
noahl: If users are allowed to pass custom start and end times, is it practical to write a test where they pass correctly-formatted ones?

abaiken: I can write a test to make sure that they can pass correctly formatted ones, but should users ever be able to pass start/end times?

abaiken: (I would think that users cannot set start/end times) which makes my other test cases somewhat useless (haha) because the models will be setting the start/end times. I just wanted to make sure that it failed with invalid values.

chambridge: Should set the serializer as read-only=true.

PR #557, opened 2018-01-31
Update cli and test cases to support https. Closes #551.

### LINKED ISSUE #551: Add https configuration to CLI
## Specify type:
- Enhancement

## Description:
Currently the qpc command line only supports http requests to the server. Now that the server can be run with https, configuration needs to be added so the default is to use https. Additionally since the container will ship with defaulted self-signed certificate and key the default will be to not verify the certificate. However, if the user updates to their cert we can try and validate with it.

CLI update:
`qpc server config` will have new optional flags
- `--use-http`  for developers; not documented
- `--ssl-verify=/path/to/cert`  to verify the cert for the server

chambridge:


elyezer:
This will require updating the automation CLI tests too. Thank you letting us know.

PR #563, opened 2018-01-31
Update to use RCUE styling

jeff-phillips-18:
- Add RCUE dependency
- Bumps patternfly-react to 1.6.0
- Updates nav bar top border
- Updates masthead imagery
- Updates About Modal


client/scripts/api.sh
```diff
@@ -17,7 +17,7 @@ mockApi()
 
   if [ -z "$(docker ps | grep $CONTAINER)" ] && [ "$UPDATE" = false ]; then
     echo "Starting API..."
-    docker run -d --rm -p $PORT:8000 -v "$FILE:/data/swagger.yaml" --name quipucords-mock $CONTAINER >/dev/null
+    docker run -d --rm -p $PORT:8000 -v "$FILE:/data/swagger.yaml" --name quipucords-mock $CONTAINER
```
cdcabrera: Probably good to remove this change from the PR

client/src/styles/entitlements/_overrides.scss
```diff
@@ -11,7 +11,12 @@ button.unavailable:hover, li.unavailable > a:hover {
   opacity: 0;
 }
 
-.navbar-brand .navbar-brand-icon, .modal-dialog .modal-footer > img {
-  height: 35px;
-}
+.about-modal-pf {
+  background-color: #1a1a1a;
+  background-image: url('/assets/img/About-Modal-background.png');
```
cdcabrera: We including this with RCUE down the road?

There's kind of a hidden aspect of the Create-React-App with Webpack... you can make the path relative, like so
```
background-image: url('../images/About-Modal-background.png');
```

And it should run and compile with the updated absolute path.

jeff-phillips-18: I'm not sure what the RCUE plan is for this. It currently copies images from patternfly so it does not have the correct image for this.

jeff-phillips-18: Excellent, all images in one place now.

Issue #534, opened 2018-01-29
UI front-end sprint: Testing, Form Integration, & View refinement


cdcabrera:
## Specify type:
- Enhancement
- UI

## Description:
- [x] Add Credentials modal/dialog, #596 #601 #603 #619 
- [x] API Redux updates, #619 #661 #677 
  - authorization headers
  - update/remove boilerplate
  - view and component optimizations
- [x] Build updates, #567 #685 #688 
  - Swagger
  - ~~Style linter~~ Postponed while considering SCSS integration into the Create React App setup through imports
- [x] Credentials view updates, #601 #603 #633 #661 #677 
- [x] Design updates, #563 #566 #653 
- Localization
  - Initial research
- [x] Paging implementation,  #677 
- [x] Patternfly React updates, #641 
- [x] Scans view updates, #633 #677
- [x] Sources view updates, #633 #677
- [x] Testing updates
  - Travis unit test review
  - Redux unit tests
- [x] Tooltips, #566
- [x] Toolbar optimize, #658  #677 
- [x] User authentication, #661 #687 
- [x] Wizard and form integrations, #581 #685 #689

cdcabrera:
#689 closes this issue

PR #566, opened 2018-01-31
UI: Update list views per mocks, add tooltips

jeff-phillips-18:
Updates #534

### DIFF
--- a/client/src/components/credentials/credentialListItem.js
+++ b/client/src/components/credentials/credentialListItem.js
@@ -1,36 +1,43 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { ListView, Button, Icon, Checkbox } from 'patternfly-react';
+import { SimpleTooltip } from '../simpleTooltIp/simpleTooltip';
 
 class CredentialListItem extends React.Component {
   render() {
     const { item, onItemSelectChange, onEdit, onDelete } = this.props;
 
     let itemIcon;
+    let credentialTypeText;
+
     switch (item.cred_type) {
       case 'vcenter':
         itemIcon = <ListView.Icon type="pf" name="virtual-machine" />;
+        credentialTypeText = 'VCenter';
         break;
       case 'network':
         itemIcon = <ListView.Icon type="pf" name="network" />;
+        credentialTypeText = 'Network';
         break;
       case 'satellite':
         itemIcon = <ListView.Icon type="fa" name="space-shuttle" />;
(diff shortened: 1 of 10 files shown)


client/src/components/sources/sources.js
```diff
@@ -12,33 +13,39 @@ import {
...
 import {
   toastNotificationTypes,
   confirmationModalTypes
 } from '../../redux/constants';
 import { bindMethods } from '../../common/helpers';
 
+import SourcesToolbar from './sourcesToolbar';
+import SourcesEmptyState from './sourcesEmptyState';
+import { SourceListItem } from './sourceListItem';
+import { CreateScanDialog } from './createScanDialog';
```
cdcabrera: Looks like a file is missing from checking? Get an error trying to load this.

client/src/components/credentials/credentialListItem.js
```diff
@@ -1,36 +1,43 @@
...
         itemIcon = <ListView.Icon type="pf" name="network" />;
+        credentialTypeText = 'Network';
         break;
       case 'satellite':
         itemIcon = <ListView.Icon type="fa" name="space-shuttle" />;
+        credentialTypeText = 'Satellite';
         break;
       default:
         itemIcon = null;
+        credentialTypeText = '';
```
cdcabrera: Note: Maybe at some point when we get enough defaults going, maybe we look at populating these variables earlier. In this case around lines 10 & 11, instead of duplicating some effort on a ```default:``` case.

client/src/components/sources/sources.js
```diff
@@ -280,7 +311,14 @@ class Sources extends React.Component {
...
           {this.renderList(filteredItems)}
-        </Grid>
+        </Grid>,
+        <CreateScanDialog
+          key="createScanDialog"
+          show={scanDialogShown}
+          sources={multiSourceScan ? selectedItems : [currentScanSource]}
+          onCancel={this.hideScanDialog}
+          onScan={this.createScan}
+        />
```
cdcabrera: Per above, the support component is missing... Commenting the reference for the component and removing this block makes the app run again...

client/src/components/credentials/credentialListItem.js
jeff-phillips-18: Yeah, I hate to set it to '' just to reset it in the valid cases. Likely the default never happens.

jeff-phillips-18:
Thanks @cdcabrera, added the missing dialog.

Issue #478, opened 2018-01-23
qpc gives missing source error when not logged in


noahl:
If I run `python bin/qpc scan start --sources sonar-jbossfuseonkaraf-rhel-7-vc` before I log in, I get
```
Source "sonar-jbossfuseonkaraf-rhel-7-vc" does not exist.
```

When I log in, the source appears. It would be helpful if the error message reflected the fact that I am not logged in, rather than talking about a missing source.

noahl:
I also tried to add a source when not logged in and got this exciting error message:
```
An error occurred while processing the "--cred" input values. Failed to add source "kevan-scan". For more information, see the server log file.
```

Logging in fixed the problem.

noahl:
This was closed by @tgerla 's PR. I'l assign it to him and close the issue.

kholdaway:
It seems to be a duplicate issue.

PR #576, opened 2018-02-01
Fix issue #478

tgerla:
- check for a valid login token before proceeding with login-required operations
- delete any old login tokins before any login attempt


qpc/clicommand.py
```diff
@@ -69,6 +69,8 @@ def _do_command(self):
         self.response = request(method=self.req_method, path=self.req_path,
                                 params=self.req_params,
                                 payload=self.req_payload)
+
+        print(self.response.status_code)
```
chambridge: Looks like this print is breaking the tests. Should be removed anyways.

PR #581, opened 2018-02-02
UI Initial add sources wizard

cdcabrera:
Initial wizard steps added for updating sources, should only display basic sample copy. Since it affects the shared ```components/sources/sources.js``` file, checking it in early.

Need an opinion about where to place this component next round? Either leave it under -
- ```components/sources/..``` 

or place it under it's own  directiory like 
- ```components/addSourceWizard/..```

updates #534

### DIFF
--- a/client/src/components/sources/sources.js
+++ b/client/src/components/sources/sources.js
@@ -5,11 +5,12 @@ import Store from '../../redux/store';
 
 import {
   Alert,
-  Grid,
   EmptyState,
-  Row,
+  Grid,
+  Icon,
   ListView,
-  Modal
+  Modal,
+  Row
 } from 'patternfly-react';
 
 import { getSources } from '../../redux/actions/sourcesActions';
@@ -23,13 +24,13 @@ import SourcesToolbar from './sourcesToolbar';
 import SourcesEmptyState from './sourcesEmptyState';
 import { SourceListItem } from './sourceListItem';
 import { CreateScanDialog } from './createScanDialog';
+import { AddSourceWizard } from './addSourceWizard';
 
 class Sources extends React.Component {
   constructor() {
     super();
 
     bindMethods(this, [
-      'addSource',
       'importSources',
       'scanSource',
       'scanSources',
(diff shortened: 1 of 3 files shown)


client/src/components/sources/sources.js
```diff
@@ -148,13 +152,21 @@ class Sources extends React.Component {
...
       type: toastNotificationTypes.TOAST_ADD,
       alertType: 'error',
       header: 'NYI',
       message: 'Adding sources is not yet implemented'
     });
+    */
+  }
+
+  hideAddSourceWizard() {
+    this.setState({ addSourceWizardShown: false });
```
jeff-phillips-18: We should probably ask for confirmation so an inadvertent click outside the wizard doesn't dismiss and force the user to start over again.

client/src/components/sources/sources.js
```diff
@@ -148,13 +152,37 @@ class Sources extends React.Component {
...
+      Store.dispatch({
+        type: confirmationModalTypes.CONFIRMATION_MODAL_HIDE
+      });
+
+      this.setState({ addSourceWizardShown: false });
+    };
+
+    Store.dispatch({
+      type: confirmationModalTypes.CONFIRMATION_MODAL_SHOW,
+      title: 'Quit Adding Source',
```
jeff-phillips-18: How about 'Cancel Add Source'?

client/src/components/sources/sources.js
```diff
@@ -148,13 +152,37 @@ class Sources extends React.Component {
...
       type: toastNotificationTypes.TOAST_ADD,
       alertType: 'error',
       header: 'NYI',
       message: 'Adding sources is not yet implemented'
     });
+    */
+  }
+
+  quitAddSourceWizard() {
+    let heading = <h3>Are you sure you want to quit adding a source?</h3>;
```
jeff-phillips-18: cancel rather than quit

client/src/components/sources/sources.js
cdcabrera: heh so i started to use the ```cancel``` wording except... there's a "cancel" button on the confirmation modal... that feels like a dark pattern...

client/src/components/sources/sources.js
cdcabrera: I'd love to use the "cancel" wording, but unless we change the name of the "cancel" button, I think we should be careful on that one.

cdcabrera: Per @serenamarie125 @bclarhk we're aiming to use ```no``` and ```yes``` for the buttons

PR #596, opened 2018-02-02
UI: Add Create Credential dialog

jeff-phillips-18:
Added some helper functions as well


client/src/components/createCredentialDialog/createCredentialDialog.js
```diff
@@ -0,0 +1,453 @@
...
+      <Grid.Col componentClass={Form.ControlLabel} sm={5}>
+        {label}
+      </Grid.Col>
+    );
+  }
+
+  renderAuthForm() {
+    const { authorizationType, password, sshKeyFile } = this.state;
+
+    if (authorizationType === 'usernamePassword') {
```
cdcabrera: Optional: possible switch, or removing the "else" on line ```228``` since they're both returning values

client/src/components/createCredentialDialog/createCredentialDialog.js
```diff
@@ -0,0 +1,453 @@
...
+      credentials.find(credential => {
+        return credential.name === name;
+      })
+    );
+  }
+
+  validateCredentialName(credentialName) {
+    return (
+      credentialName &&
+      credentialName.length > 0 &&
```
cdcabrera: Optional: truthy check

client/src/components/createCredentialDialog/createCredentialDialog.js
```diff
@@ -0,0 +1,453 @@
...
+              <SplitButton
+                className="form-control"
+                bsStyle="default"
+                title={becomeMethod}
+                id="become-method-button"
+              >
+                {becomeMethods.map((nextMethod, index) => (
+                  <MenuItem
+                    key={index}
+                    eventKey={'become' + index}
```
cdcabrera: String/template literal

client/src/components/createCredentialDialog/createCredentialDialog.js
```diff
@@ -0,0 +1,453 @@
...
+        break;
+      case 'network':
+        credentialTypeText = 'Network';
+        break;
+      case 'satellite':
+        credentialTypeText = 'Satellite';
+        break;
+      default:
+        credentialTypeText = '';
+    }
```
cdcabrera: This setup looks related to the ```helpers.js``` work where we should probably start considering using constant objects with key/value pairs.

```js
const credentialTypeText = someGlobalConstant[credentialType] || '';
```

client/src/redux/actions/credentialsActions.js
```diff
@@ -33,9 +33,51 @@ const getCredentials = () => {
...
+      .catch(error => {
+        dispatch(addCredentialError(true, error.message));
+        dispatch(addCredentialLoading(false));
+        if (addCallback) {
+          addCallback(true, error.message);
+        }
+      });
+  };
+};
+
```
cdcabrera: Note: we need to re-evaluate the callback setup... I'm thinking next round either we return a promise instead, that way we can remove the checks/logic around it... Or we figure out how to work with state value observation

jeff-phillips-18:
Updated. Leaving constants/strings stuff for a follow on.

client/src/redux/actions/credentialsActions.js
cdcabrera: not sure why I didn't recommend this at first, but @jeff-phillips-18 this is probably the use case for using [Redux ```Store.subscribe```](https://redux.js.org/docs/api/Store.html#subscribe)

PR #617, opened 2018-02-06
Add ssl option beta info for vCenter and Satellite. Closes #615.

### LINKED ISSUE #615: Add beta documentation tweaks for VCenter & Satellite for SSL options
## Specify type:
- Enhancement

## Description:
New options have been added to support SSL connectivity options for VCenter and Satellite. We need to document these options for users as they will commonly use some option to either turn off cert validation or stop the use of SSL entirely.

____
Related to issue: https://github.com/quipucords/quipucords/issues/423

chambridge:
@jlprevatt FYI

Minor update to beta doc to provide more setup configuration for Satellite or vCenter around SSL connection configuration.

### DIFF
--- a/docs/source/beta.rst
+++ b/docs/source/beta.rst
@@ -283,6 +283,9 @@ For example, for a vCenter source where the name is **myvcenter**, the vCenter s
 
   # qpc source add --type vcenter --name myvcenter --hosts 192.0.2.10 --cred vcenter_admin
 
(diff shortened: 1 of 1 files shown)


chambridge:
@elyezer It maybe possible. I'll have to see if vCenter supports that as well (trying to solve lowest common denominator). However, I also think may be something we wait to see if it is requested as I think most customers may not care since they are running this server to talk to another of there on-premise servers so they would implicitly trust the endpoint they are talking to since its all within their infrastructure.

PR #630, opened 2018-02-06
Add pagination to the credentials list api. Closes #613.

### LINKED ISSUE #613: Support pagination in credentials List API and CLI
## Specify type:
- Enhancement

## Description:
Current credentials List API does not support pagination. We need to update this prior to the initial release. We will be utilizing the standard Django format: http://www.django-rest-framework.org/api-guide/pagination/#pagenumberpagination

The current output should be rolled into the `results` field.

CLI should support paging by being able to make a new request if a `next` value exists. We will prompt with an `Press enter to see the next set of results.`

_____

chambridge:
FYI - @quipucords/ux-ui


jeff-phillips-18:
We will need to be able to sort on the get call, sorting by name and cred_type should be all we need (IMO).

chambridge:
@jeff-phillips-18 
https://github.com/quipucords/quipucords/issues/631

quipucords/quipucords/settings.py
```diff
@@ -112,9 +112,13 @@
 
 WSGI_APPLICATION = 'quipucords.wsgi.application'
 
+DEFAULT_PAGINATION_CLASS = 'rest_framework.pagination.PageNumberPagination'
+
 REST_FRAMEWORK = {
     'DEFAULT_FILTER_BACKENDS':
         ('django_filters.rest_framework.DjangoFilterBackend',),
+    'DEFAULT_PAGINATION_CLASS': DEFAULT_PAGINATION_CLASS,
+    'PAGE_SIZE': 10
```
kholdaway: Is this fixed or can it be specified per query?

qpc/cred/list.py
```diff
@@ -53,8 +54,20 @@ def _build_req_params(self):
...
+
+        if json_data.get('next'):
+            next_link = json_data.get('next')
+            params = urlparse.parse_qs(urlparse.urlparse(next_link).query)
+            page = params.get('page', ['1'])[0]
+            if self.req_params:
+                self.req_params['page'] = page
+            else:
+                self.req_params = {'page': page}
+            input('Press enter to see the next set of results.')
```
kholdaway: Does the man page need to describe this?

quipucords/quipucords/settings.py
chambridge: This is just the default. Can be overridden.

qpc/cred/list.py
chambridge: I don’t believe so.

jeff-phillips-18:
The UI allows the user to set the page size as well as go to each page of results. Will that be possible?

Here is the pagination control:

image

PR #633, opened 2018-02-06
UI: update views, sort/filters per mocks

jeff-phillips-18:
Updates #534


jeff-phillips-18:
Well.... that just outta work. It definitely used to. Lemme take a look.

PR #640, opened 2018-02-07
Fix typo in messages and update test to match

kdelee:
I noticed this while I was working and thought it best just to go ahead and fix it.


kdelee:
Travis had a problem installing node. I don't think I have the right permissions to ask it to try again.

PR #645, opened 2018-02-07
Make tasks return the empty string for errors

### LINKED ISSUE #626: Make sure we return blank strings for errors
Some of the tasks that were carried over from Rho use Ansible error handling to return error messages when a task fails. Make sure that all of these are removed and replaced with blank strings, because that is how we are representing errors in Sonar.

noahl:
This is how we are modeling errors in Sonar, so we need to convert all
of the tasks to match.

Closes #626 .


roles/cpu/tasks/main.yml
```diff
@@ -69,7 +69,7 @@
 
 - name: extract result value for cpu.socket_count
   set_fact:
-    cpu_socket_count: "{{ internal_cpu_socket_count_cmd.get('stdout') | trim | default('error') if internal_have_dmidecode else 'N/A (dmidecode not found)' }}"
+    cpu_socket_count: "{{ internal_cpu_socket_count_cmd.get('stdout') | trim | default('') if internal_have_dmidecode else '' }}"
```
chambridge: do we need the if condition now?

roles/date/tasks/main.yml
```diff
@@ -41,5 +41,5 @@
 
 - name: extract result value for date.yum_history
   set_fact:
-    date_yum_history: "{{  internal_date_yum_history_cmd['stdout_lines'] | select | first | default(None) if internal_have_yum else 'N/A (yum not found)' }}"
+    date_yum_history: "{{  internal_date_yum_history_cmd['stdout_lines'] | select | first | default('') if internal_have_yum else '' }}"
```
chambridge: do we need the if condition anymore?

roles/dmi/tasks/main.yml
```diff
@@ -8,12 +8,12 @@
 
 - name: extract result value for dmi.bios-vendor
   set_fact:
-    dmi_bios_vendor: "{{  internal_dmi_bios_vendor_cmd['stdout'] | trim | default('error') if internal_have_dmidecode else 'N/A (dmidecode not found)' }}"
+    dmi_bios_vendor: "{{  internal_dmi_bios_vendor_cmd['stdout'] | trim | default('') if internal_have_dmidecode else '' }}"
```
chambridge: do we need the if condition any more?

roles/dmi/tasks/main.yml
```diff
@@ -24,12 +24,12 @@
 
 - name: extract result value for dmi.bios-version
   set_fact:
-    dmi_bios_version: "{{  internal_dmi_bios_version_cmd['stdout'] | trim | default('error') if internal_have_dmidecode else 'N/A (dmidecode not found)' }}"
+    dmi_bios_version: "{{  internal_dmi_bios_version_cmd['stdout'] | trim | default('') if internal_have_dmidecode else '' }}"
```
chambridge: do we need the if condition any more?

roles/dmi/tasks/main.yml
```diff
@@ -40,12 +40,12 @@
 
 - name: extract result value for dmi.system-manufacturer
   set_fact:
-    dmi_system_manufacturer: "{{  internal_dmi_system_manufacturer_cmd['stdout'] | trim | default('error') if internal_have_dmidecode else 'N/A (dmidecode not found)' }}"
+    dmi_system_manufacturer: "{{  internal_dmi_system_manufacturer_cmd['stdout'] | trim | default('') if internal_have_dmidecode else '' }}"
```
chambridge: do we need the if condition any more?

roles/dmi/tasks/main.yml
```diff
@@ -56,12 +56,12 @@
 
 - name: extract result value for dmi.processor-family
   set_fact:
-    dmi_processor_family: "{{  internal_dmi_processor_family_cmd['stdout'] | trim | default('error') if internal_have_dmidecode else 'N/A (dmidecode not found)' }}"
+    dmi_processor_family: "{{  internal_dmi_processor_family_cmd['stdout'] | trim | default('') if internal_have_dmidecode else '' }}"
```
chambridge: do we need the if condition any more?

roles/dmi/tasks/main.yml
```diff
@@ -72,10 +72,10 @@
 
 - name: extract result value for dmi.system-uuid
   set_fact:
-    dmi_system_uuid: "{{  internal_dmi_system_uuid_cmd['stdout'] | trim | default('error') if internal_have_dmidecode else 'N/A (dmidecode not found)' }}"
+    dmi_system_uuid: "{{  internal_dmi_system_uuid_cmd['stdout'] | trim | default('') if internal_have_dmidecode else '' }}"
```
chambridge: do we need the if condition any more?

noahl:
@chambridge Thanks for making me look into this more! We do need the if statement, because it protects us from a jinja exception in the case where the initial variable is not defined.

However, we do not need the `default('')`, because jinja will do that automatically. We can write these as just `internal_cpu_socket_count_cmd['stdout'] | trim if internal_have_dmidecode else ''`.

I just pushed a new commit with all of these templates simplified. Please take a look!

PR #649, opened 2018-02-07
Intermediate results for network inspect scan

### LINKED ISSUE #647: Improve network scan intermediate status
# Opening a new issue
___

## Specify type:
- Enhancement

- High

___

## Description:
Currently, a network scan will only log info when the host is complete.  This could be slow depending on the concurrency setting (may need 50 hosts to complete before you see logs).  This work will add a log entry when a role is started for a host.
___

kholdaway:
Closes #647


quipucords/scanner/network/inspect.py
```diff
@@ -188,7 +188,14 @@ def inspect_scan(self, connected, roles=DEFAULT_ROLES,
         inventory_file = write_inventory(inventory)
 
         error_msg = ''
+        group_index = 0
+        log_message = 'START PROCESSING GROUPS of size %d' % forks
+        self.scan_task.log_message(log_message)
         for group_name in group_names:
```
chambridge: suggestion for more pythonic:
```
for idx, val in enumerate(ints):
    print(idx, val)
```

roles/host_done/tasks/main.yml
```diff
@@ -1,5 +1,9 @@
 ---
 
+- name: internal_host_started_processing_role
+  set_fact:
+    internal_host_started_processing_role: "host_done"
+
 - name: extract virt-what error code
```
chambridge: Seems like a bad task name that came in.

quipucords/scanner/network/inspect.py
kholdaway: Great suggestion.

roles/host_done/tasks/main.yml
kholdaway: Fixed.

PR #652, opened 2018-02-07
Add pagination support for scans. Closes #648.

### LINKED ISSUE #648: Support pagination in scans List API and CLI
## Specify type:
- Enhancement

## Description:
Current scans List API does not support pagination. We need to update this prior to the initial release. We will be utilizing the standard Django format: http://www.django-rest-framework.org/api-guide/pagination/#pagenumberpagination

The current output should be rolled into the `results` field.

CLI should support paging by being able to make a new request if a `next` value exists. We will prompt with an `Press enter to see the next set of results.`

_____

chambridge:


jeff-phillips-18:
@chambridge Is there a plan to update the swagger.yml for this?

chambridge:
@jeff-phillips-18 https://github.com/quipucords/quipucords/issues/668 Thanks for catching this, I'll get it in this afternoon.

Issue #666, opened 2018-02-08
Server should validate the existence of the keyfile, not CLI


kdelee:
## Specify type:
- Bug

### Bug severity (if applicable):
- Blocker
___

## Description:
The command line client, `qpc` attempts to validate the path to the keyfile before it even talks to the server. This does not work when `qpc` is on a different filesystem than the server.

___

### Expected behavior:
As it stands, the sshkeys are expected to be on the same filesystem as the server is running on.
So, for example, if I am running the server in a docker container, and I have mounted a local directory `~/quipucords/sshkeys` to the container's `/sshkeys`, then based on the documentation I expect to to put the key in my local dir and then specify the container path to qpc.

### Steps to reproduce:
List clear steps to reproduce the bug.
0) Run the server from a container and mount the container's `/sshekys` to something that is NOT that path on the local system. I used `~/quipucords/sshkeys/:/sshkeys:z` (because SELinux is enabled)
1) Copy my sshkey into `~/quipucords/sshkeys`. So now it lives at `~/quipucords/sshkeys/my_fav_id_rsa`
2) Create a credential that uses it by proving the path in the container, which is `/sshkeys/my_fav_id_rsa`
e.g. 
```
qpc cred add --name myfav_key --username elijah --sshkeyfile /sshkeys/my_fav_id_rsa --type network
```
`qpc` attempts to do the validation of the path and says:
```
The file path provided, /sshkeys/my_fav_id_rsa, could not be found on the system. Provide a valid path for the "--sshkeyfile" argument.
```
I'm not 100% certain what is going on because it ALSO gives me same error if the file exists on local filesystem but is not something that exists in the container.
e.g.
```
qpc cred add --name myfav_key --username elijah --sshkeyfile=/home/elijah/.ssh/my_fav_id_rsa

The file path provided, /home/elijah/.ssh/my_fav_id_rsa, could not be found on the system. Provide a valid path for the "--sshkeyfile" argument.
```
ALSO fails and gives me the same error message, even though this DOES exist on local file system.

Only thing that works for me right now is if the local path and the container path are identical.
I must make there be SOME file in the identical path as in the container, and then it will work.
It appears to actually use the key in the mounted volume `~/quipucords/sshkey/my_fav_id_rsa --> /sshkeys/my_fav_id_rsa` but `qpc` requires some file to exist locally. The following works for me:
```
cp ~/.ssh/my_fav_id_rsa ~/quipucords/sshkeys
sudo mkdir /sshkeys
[...]

Issue #654, opened 2018-02-07
Possible ways to cut down on container size/layers?


kdelee:
While building the container I noticed that it downloads and installs some packages that should only be needed for development, like `coverage`.

Can we think about breaking out the dependencies into `dev` and standard packages in `setup.py` or by having different requirements files? I'm sure there are several ways to go about this.

chambridge:
This is doable. Reason it was there is because its part of the `INSTALLED_APPS` in the `settings.py`. But the code could be adjusted to add `'coverage'` to the list if `PRODUCTION` is `False` and the single `pip` entry could be removed. The rest of the development dependencies are not pulled as the container image has a step that removes pulling in the dependencies in `dev-requirements.txt`

PR #667, opened 2018-02-08
Beta doc updates, remove coverage, fix cli add edit sshkeyfile check. Closes #664 #654 #666.

### LINKED ISSUE #664: Beta doc updates for SELinux and simpler usability
## Specify type:
- Enhancement

## Description:
Capture some of the updates from Elijah on SELinux and usability.

chambridge:


docs/source/beta.rst
```diff
@@ -174,29 +181,34 @@ There are several configurable options that must be considered:
...
+  # mkdir log
 
-  docker run --name quipucords -d -p443:443 -v /opt/quipucords/sshkeys:/sshkeys -v /opt/quipucords/data:/var/data -v /opt/quipucords/log:/var/log -i quipucords:latest
+Following these steps we can now launch the Quipucords server with the following docker command, if your system does not have **SELinux** enabled::
+
+  # sudo docker run --name quipucords -d -p 443:443 -v ~/quipucords/sshkeys:/sshkeys -v ~/quipucords/data:/var/data -v ~/quipucords/log:/var/log -i quipucords:latest
+
+If your system does have **SELinux** enabled you will need to append ":z" to each volume mapping as follows::
+
+  # sudo docker run --name quipucords -d -p 443:443 -v ~/quipucords/sshkeys:/sshkeys:z -v ~/quipucords/data:/var/data:z -v ~/quipucords/log:/var/log:z -i quipucords:latest
```
kdelee: 👍

quipucords/quipucords/settings.py
```diff
@@ -73,9 +74,12 @@
     'django_filters',
     'drf_generators',
     'api',
-    'coverage'
 ]
 
+if not PRODUCTION:
+    INSTALLED_APPS.append('coverage')
+
```
kdelee: 👍

quipucords/quipucords/settings.py
```diff
@@ -56,7 +56,8 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = bool(os.environ.get('DJANGO_DEBUG', True))
```
kdelee: Doesn't have to be this PR but should the default be `False` ?

qpc/cred/utils.py
```diff
@@ -12,30 +12,11 @@
...
-    """
-    keyfile_path = os.path.abspath(os.path.normpath(keyfile))
-    if not os.path.isfile(keyfile_path):
-        print(_(messages.VALIDATE_SSHKEY % (keyfile)))
-        parser.print_help()
-        sys.exit(1)
-    else:
-        # set filename to the resolved keyfile_path
-        return keyfile_path
-
```
kdelee: I see you removing the path check for the sshkey in the CLI but don't see you adding anything to the server side. Is that something you want to do later? 🙁 Guess that would make creating a sshkey have to return a `201` and then later report whether or not it worked ¯\_(ツ)_/¯

docs/source/beta.rst
```diff
@@ -174,29 +181,34 @@ There are several configurable options that must be considered:
...
+If you selected to expose port 443 then you would use the following option when running the image ``-p443:443``. If you wish to re-map the port on your system Docker's mapping is -p<host_port>:<container_port>. If you choose for example to re-map the port to 8443 the option to supply would be ``-p8443:443``. Additionally, Docker supplies an option to select a free port for all exposed ports by using the ``-P`` option; the port mapping is then available from the ``sudo docker ps`` command.
 
 
 For the next three configuration options we will take a simple setup strategy for the Quipucords server and create a "home directory" for the server.
 
-1. Create the home directory ``/opt/quipucords``::
+1. Create the home directory ``~/quipucords``::
 
-  mkdir -p /opt/quipucords
+  # mkdir -p ~/quipucords
```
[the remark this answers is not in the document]
chambridge: I see I was just trying to provide a consistent command prompt symbol. The doc previously had `$`, `#`, or nothing.

qpc/cred/utils.py
chambridge: Already had this validation correctly on the server:
https://github.com/quipucords/quipucords/blob/master/quipucords/api/credential/serializer.py#L147

Should not have been in the CLI.

quipucords/quipucords/settings.py
chambridge: Its false in the container environment: https://github.com/quipucords/quipucords/blob/master/Dockerfile#L58

But allows dev to run with it set to True.

kdelee:
FWIW I think maybe this section could get refactored:

```
Verify the server is responding correctly by launching a browser to https://<ip_address>:<port>/admin. If 
your browser is on the same system as the server and you exposed port 443 the URL would be 
https://localhost/admin. When your browser loads you should see the administrative login dialog. From 
here you can log into the server and change the default password. The server comes defaulted with a user 
admin and password pass. You should find the “Change Password” selection in the upper right navigation 
bar.
```

Maybe a heading that says "Verify install" and line break and new sub heading "Change admin password"

quipucords/quipucords/settings.py
kdelee: ah cool

qpc/cred/utils.py
kdelee: 🎉 🎈

PR #674, opened 2018-02-09
Make user name prompt match password prompt

kdelee:
Currently there is no space after the `User name:` prompt and this is slightly annoying 🙄 . 
```
$ qpc server login
User name:admin
Password: (your cursor starts here)
```


jlprevatt:
Retroactively agree, good catch.

PR #687, opened 2018-02-09
UI: add user actions and login checking

jeff-phillips-18:
Updates #534


client/.env.development
```diff
@@ -9,3 +9,5 @@ REACT_APP_SCANS_SERVICE_RESTART=http://localhost:4000/api/v1/scans/{0}/restart/
 REACT_APP_SCANS_SERVICE_RESULTS=http://localhost:4000/api/v1/scans/{0}/results/
 
 REACT_APP_SOURCES_SERVICE=http://localhost:4000/api/v1/sources/
+
+REACT_APP_DEBUG_MIDDLEWARE=false
```
cdcabrera: Note: Not sure this is the best location for this setting... Something that would limit it on a per dev basis feels better. In a follow up PR I'll look at removing this and updating the ReadMe to reflect that:
>"[if you want to run the logger create a .env.local file with REACT_APP_DEBUG_MIDDLEWARE=true]"

client/src/redux/reducers/scansReducer.js
```diff
@@ -48,7 +48,6 @@ const scansReducer = function(state = initialState, action) {
         'view',
         {
           scans: action.payload.results,
-
```
cdcabrera: Heh =)

client/src/redux/store.js
```diff
@@ -12,7 +13,14 @@ const hydrateStore = () => {
...
 if (process.env.NODE_ENV !== 'production') {
-  middleware.push(createLogger());
+  if (process.env.REACT_APP_DEBUG_MIDDLEWARE === 'true') {
+    middleware.push(createLogger());
+  }
+  cookies.set(
+    'csrftoken',
+    'xLPbEvorrDHqC9poUjUpKtJcf2kmskkwVzHpkiwqO57zD9ViBtrtqEud6llTnVPH'
+  );
+  console.warn('Warning: Loading spoof auth token.');
```
cdcabrera: Note: In a follow up PR we could probably investigate some development "helper" functions that handle these one-offs dev checks.

Issue #690, opened 2018-02-12
UI front-end sprint: Testing & View refinement, Initial authentication and compile


cdcabrera:
## Specify type:
- Enhancement
- UI

## Description:
- [x] Build updates #727 #754  #774 #804 #832 
- [x] Form refinement #728 #749 #689 #804
  - Wizard ```689, 804``` 
  - Credential ```749```
- [x] Testing updates #774 #781 #804 
  - Travis, coordinate
  - Unit ```774, 781, 804``` 
- [x] User authentication #713 #761 
- [x] View refinement #712 #728 #749 #761 #798 #805 #809 #833 #846 
  - Login ```712, 761``` 
  - Sources ```805, 809```
  - Scans ```728, 805, 809, 833```
  - Credentials ```728```

kdelee:
@cdcabrera, @elyezer was wondering about your progress on this. Should he wait to merge his UI test until this issue is done? He was getting authentication errors and was wondering if you all were not yet collecting and passing the authentication token after login.

cdcabrera:
@kdelee @elyezer indeed, the authentication error your seeing is related to that, and yes, it would probably be good for you to wait for part of this issue to be resolved.  

We've been using these collected issues as tracking for sprint UI work. In fact, saying that, I may update the issue name to reflect that better.

#### What the UI still has to implement... 
Basically we have to place an authentication component with the token in-between the API calls and the UI. We created the base for this last sprint, the integration piece is this sprint. We'll keep you guys in the loop.

chambridge:
Should this be closed?

cdcabrera:
@chambridge close, we have one remaining bit for today #846

PR #712, opened 2018-02-12
UI: Update login dialog to RCUE styling

jeff-phillips-18:
Updates #690 

image

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -25,7 +25,10 @@
     "watch:css": "npm run build:css && node-sass --include-path ./src/styles $npm_package_sassDepIncludes src/styles/entitlements.scss -o src/styles/css --watch --recursive",
     "build:js": "react-scripts build",
     "build": "run-s build:css build:js",
-    "build:prod": "npm run build && mkdir -p ../quipucords/client && cp -R ./build/* ../quipucords/client",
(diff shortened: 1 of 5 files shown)


quipucords/quipucords/templates/registration/login.html
```diff
@@ -1,9 +1,28 @@
 {% extends "base.html" %}
 
+{% block includes %}
+  {% load static %}
```
cdcabrera: Optional: feels meta including a variable/param inside a block include, maybe move it outside

client/package.json
```diff
@@ -34,7 +34,8 @@
     "start:quipucords": "open http://127.0.0.1:8000/index.html; source ../venv/bin/activate; cd ../; make serve",
     "test": "run-s test:**",
     "test:lint-js": "eslint --ext=js --ext=jsx src",
-    "test:unit": "react-scripts test --env=jsdom"
+    "test:unit": "react-scripts test --env=jsdom",
+    "postinstall": "sh ./scripts/postinstall.sh"
```
cdcabrera: Round 2: nice! looking at it now, wonder if we should eventually convert it to also be part of the ```build:prod```

PR #719, opened 2018-02-13
Option to create connection scan for creation of source Closes #611

### LINKED ISSUE #611: Have an option to create a connection scan when creating a source
## Specify type:
- Enhancement

## Description:
Currently the creation of a source does not trigger a scan. We want to expose a query parameter on the POST command, `scan=True`, that will create a `connect` ScanJob for this source.
____

abaiken:


quipucords/api/source/view.py
```diff
@@ -116,6 +117,26 @@ def create(self, request, *args, **kwargs):
 
         # Create expanded host cred JSON
         expand_credential(json_source)
+
+        # check to see if a connection scan was requested
+        # through query parameter
+        scan = request.query_params.get('scan', False)
```
chambridge: what happens if someone provides `?scan=ImNotABoolean`?  I feel like you could add a check for is_boolean, similar to: https://github.com/quipucords/quipucords/blob/master/quipucords/api/common/util.py#L21

We should throw a Validation error if the value isn't True or False

quipucords/api/source/view.py
```diff
@@ -116,6 +117,26 @@ def create(self, request, *args, **kwargs):
...
+        # If the scan was requested, create a connection scan
+        if scan:
+            # Grab the last source that was created
+            scan_source = self.queryset.latest('id')
```
chambridge: I would think that the response object has the `id` and you shouldn't need to query for it.

quipucords/api/common/util.py
```diff
@@ -45,6 +45,28 @@ def convert_to_int(value):
...
+
+    :param value: The value to check
+    :returns True if it is a bool, False if not
+    """
+    if value == 'True' or value == 'False':
+        return True
+    return False
+
+
+def convert_to_boolean(value):
```
kholdaway: Python should do this with bool()

quipucords/api/common/util.py
```diff
@@ -45,6 +45,28 @@ def convert_to_int(value):
...
+    """Convert a string 'True' or 'False' to boolean.
+
+    :param value: The value to convert
+    :return The value as a bool
+    """
+    if value == 'True':
```
kholdaway: I would compare with a case insensitive compare or convert to lower first.

quipucords/api/common/util.py
kholdaway: Actually, sounds like bool does the wrong thing here.  This suggestion seems reasonable instead. https://stackoverflow.com/questions/715417/converting-from-a-string-to-boolean-in-python  May want to check if you are dealing with a string first.

quipucords/api/common/util.py
kholdaway: https://stackoverflow.com/questions/715417/converting-from-a-string-to-boolean-in-python  Test for is bool.

quipucords/api/messages.py
```diff
@@ -48,6 +48,7 @@
 SOURCE_CRED_WRONG_TYPE = 'Credentials must have the same type as source.'
 SOURCE_TYPE_REQ = 'A value for source_type must be provided to ' \
                   'create a source.'
+SOURCE_CONNECTION_SCAN = 'Scan parameter must be type boolean.'
```
chambridge: SOURCE_CONNECTION_SCAN = 'The query parameter scan must be a boolean.'

PR #728, opened 2018-02-14
UI: Fixes for nav selection, add scan actions

jeff-phillips-18:


client/src/components/sources/sources.js
```diff
@@ -123,18 +115,11 @@ class Sources extends React.Component {
...
-    });
-
-    this.setState({ selectedItems: selectedItems });
-
     Store.dispatch({
-      type: sourcesTypes.SOURCES_SELECTED,
-      selectedSources: selectedItems
+      type: item.selected
+        ? sourcesTypes.DESELECT_SOURCE
+        : credentialsTypes.SELECT_SOURCE,
```
cdcabrera: Looks like ```SELECT_SOURCE``` doesn't exist for ```credentialsTypes``` ... for testing I changed it to
```js
? sourcesTypes.DESELECT_SOURCE
: sourcesTypes.SELECT_SOURCE,
```

client/src/redux/reducers/credentialsReducer.js
```diff
@@ -28,8 +29,49 @@ const initialState = {
...
   switch (action.type) {
+    // Persist
+    case credentialsTypes.SELECT_CREDENTIAL:
+      // Do nothing if it is already selected
+      if (selectedIndex(state, action.credential) !== -1) {
+        return state;
+      }
+
+      action.credential.selected = true;
+
```
cdcabrera: I think this should probably change, it's considered an anti-pattern to mutate the original objects coming in. In this case the original action object, it does work though... maybe a ```const tempActionCred = Object.assign(...``` in there

client/src/redux/reducers/credentialsReducer.js
```diff
@@ -28,8 +29,49 @@ const initialState = {
...
+    case credentialsTypes.DESELECT_CREDENTIAL:
+      const index = selectedIndex(state, action.credential);
+
+      // Do nothing if it is not already selected
+      if (index === -1) {
+        return state;
+      }
+
+      action.credential.selected = false;
```
cdcabrera: Same mutation comment.

client/src/redux/reducers/sourcesReducer.js
```diff
@@ -28,20 +29,45 @@ const initialState = {
...
-          reset: false
-        }
-      );
+    case sourcesTypes.SELECT_SOURCE:
+      // Do nothing if it is already selected
+      if (selectedIndex(state, action.source) !== -1) {
+        return state;
+      }
+
+      action.source.selected = true;
```
cdcabrera: Same mutation comment above

client/src/redux/reducers/sourcesReducer.js
```diff
@@ -28,20 +29,45 @@ const initialState = {
...
+
+      const addedSelections = Object.assign({}, state.persist, {
+        selectedSources: [...state.persist.selectedSources, action.source]
+      });
+      return Object.assign({}, state, { persist: addedSelections });
```
cdcabrera: Note: This is incredibly similar to the removed code with an addition, just long form

client/src/redux/reducers/sourcesReducer.js
```diff
@@ -28,20 +29,45 @@ const initialState = {
...
+
+    case sourcesTypes.DESELECT_SOURCE:
+      const index = selectedIndex(state, action.source);
+
+      // Do nothing if it is not already selected
+      if (index === -1) {
+        return state;
+      }
+
+      action.source.selected = false;
```
cdcabrera: Same mutation comment above.

client/src/components/sources/sources.js
jeff-phillips-18: copy/paste error.

client/src/redux/reducers/credentialsReducer.js
jeff-phillips-18: Yeah, I just hated having to set selected value then dispatch an action to get it into the selected items list. I can check if hte item is in the selectedItems list rather than keep the flag, but that happens a lot.

client/src/redux/reducers/sourcesReducer.js
jeff-phillips-18: You are correct sir. I will adjust to use the setStateProp

client/src/redux/reducers/credentialsReducer.js
jeff-phillips-18: Removed the mutation and added the check in the views.

jeff-phillips-18:
Yeah, already working on a bunch of fixes to handle API results so I'll work that issue next as well.

PR #745, opened 2018-02-15
Issues/665

### LINKED ISSUE #665: Log data coming from Ansible
Make sure all data coming into Sonar from Ansible is logged to a single file at a known location in the filesystem.

This is for debugging purposes later - if anything goes wrong with Sonar, we can ask customers to send us that file and then look for whatever confused us.

noahl:
This PR is the first part of the logging work. It logs data from Ansible to a series of log files. It knows how to remove the files when they age out or when the total size gets too big. Future work is to log data from Satellite and VCenter too (shouldn't be much more code beyond this) and to be able to reingest data from logs (another substantial piece of work).

Closes #665 .

### DIFF
--- a/quipucords/quipucords/settings.py
+++ b/quipucords/quipucords/settings.py
@@ -281,7 +281,7 @@ LOGGING = {
     },
 }
 
-INPUT_LOG_FILE = os.path.join(BASE_DIR, 'input_log')
+INPUT_LOG_BASENAME = os.path.join(BASE_DIR, 'input_log')
 
 # Reverse default behavior to avoid host key checking
 os.environ.setdefault('ANSIBLE_HOST_KEY_CHECKING', 'False')
--- a/quipucords/quipucords/settings.py
+++ b/quipucords/quipucords/settings.py
@@ -281,5 +281,7 @@ LOGGING = {
     },
 }
 
+INPUT_LOG_FILE = os.path.join(BASE_DIR, 'input_log')
+
 # Reverse default behavior to avoid host key checking
 os.environ.setdefault('ANSIBLE_HOST_KEY_CHECKING', 'False')
--- a/quipucords/scanner/input_log.py
+++ b/quipucords/scanner/input_log.py
@@ -1,0 +1,108 @@
+# Copyright (c) 2018 Red Hat, Inc.
+#
+# This software is licensed to you under the GNU General Public License,
+# version 3 (GPLv3). There is NO WARRANTY for this software, express or
+# implied, including the implied warranties of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv3
+# along with this software; if not, see
+# https://www.gnu.org/licenses/gpl-3.0.txt.
+
+"""Log all input to the filesystem for later debugging."""
+
+import io
+import json
+import multiprocessing as mp
+import uuid
+from django.db import transaction
+from api import log_state
+from quipucords import settings
+
+UUID_CACHE = None
+
+
+@transaction.atomic()
+def get_sonar_uuid():
+    """Get the UUID of this Sonar installation."""
+    global UUID_CACHE  # pylint: disable=global-statement
+    if UUID_CACHE is not None:
+        return UUID_CACHE
+
+    uuid_obj = log_state.DatabaseUUID.objects.all().first()
+    if uuid_obj is None:
+        uuid_obj = log_state.DatabaseUUID(uuid=uuid.uuid4())
+        uuid_obj.save()
+    UUID_CACHE = uuid_obj.uuid
+
+    return uuid_obj.uuid
+
+
+# This is a potential performance bottleneck since we are touching the
+# database every time we write a record. This isn't necessary - we
+# could get the state of the counter by reading the last few entries
+# of the existing log files on startup. I don't plan to implement that
+# unless we find that this is actually a problem.
+@transaction.atomic()
+def next_sequence_number():
+    """Get the next sequence number and increment the counter."""
+    seq = log_state.LatestSequenceNumber.objects.all().first()
+    if seq is None:
+        seq = log_state.LatestSequenceNumber(number=0)
+        seq.save()
+
+    number = seq.number
+    seq.number += 1
+    seq.save()
+
+    return number
+
+
+INPUT_LOG = None
+WRITE_LOCK = mp.Lock()
+
+
+def log_fact(host, fact, value):
+    """Write a new fact to the input log.
+
+    :param host: string. the name of the host the fact came from.
+    :param fact: json object. The fact identifier.
+    :param value: the value of the fact.
+    """
+    global INPUT_LOG  # pylint: disable=global-statement
+    if INPUT_LOG is None:
+        INPUT_LOG = open(settings.INPUT_LOG_FILE, 'a')
+
+    database_uuid = get_sonar_uuid()
+    seq = next_sequence_number()
+
+    with WRITE_LOCK:
+        INPUT_LOG.write(json.dumps({'database_uuid': str(database_uuid),
+                                    'host': host,
+                                    'fact': fact,
+                                    'sequence_number': seq,
+                                    'result': value}))
+        INPUT_LOG.write('\n')
+        INPUT_LOG.flush()
+
+
+def disable_log_for_test():
+    """Disable the input log for testing purposes."""
+    global INPUT_LOG
+    INPUT_LOG = io.StringIO()
+
+
+RAW_PARAMS = '_raw_params'
+
+
+def log_ansible_result(result):
+    """Log a shell command."""
+    # pylint: disable=protected-access
+    args = result._task.args
+
+    if RAW_PARAMS in args:
+        log_fact(result._host.name,
+                 {'type': 'shell', 'command': args[RAW_PARAMS]},
+                 result._result)
+    else:
+        print('*** Ansible result:', result)
+    # If _raw_params isn't in args, then args is not a raw command and
+    # we don't need to record it. This occurs for set_facts tasks.
--- a/quipucords/scanner/input_log.py
+++ b/quipucords/scanner/input_log.py
@@ -11,11 +11,15 @@
 
 import io
 import json
+import logging
 import multiprocessing as mp
+import os
 import uuid
 from django.db import transaction
 from api import log_state
 from quipucords import settings
+
+logger = logging.logger(__name__)
 
 UUID_CACHE = None
 
@@ -54,6 +58,67 @@ def next_sequence_number():
     seq.save()
 
     return number
+
+
+class RotatingLogFile(object):
+    """A log file that automatically rotates itself when full."""
+
+    # This class maintains a set of log files with names 'basename-0',
+    # 'basename-1', .... It automatically switches to a new one when
+    # the last log file is either too big or too old, and removes
+    # earlier ones when they exceed the size or age limits. A few
+    # flaws:
+    #   - It only re-evaluates its log file list on writes. If you
+    #     never write anything, it will never rotate the logs.
+    #   - It is not safe to use more than one instance of this class
+    #     with the same basename. It makes no attempt at locking, and
+    #     it doesn't look for notifications of new log files after
+    #     it's gathered its initial set.
+
+    def __init__(self, basename, max_size, max_age,
+                 step_size=None, step_age=None):
+        # Log files will be named 'basename-0', 'basename-1', etc.
+        self.basename = basename
+        self.max_size = max_size
+        self.max_age = max_age
+
+        # The size and age limits where we start a new log file.
+        self.step_size = step_size or max_size // 10
+        self.step_age = step_age or max_age // 10
+
+        # self.log_files will be a list of dicts with entries 'name',
+        # 'created', and 'size'. The list will be sorted with the
+        # newest entries in the front.
+        self.log_files = []
+        self.file_counter = 0
+
+        for name in os.listdir(settings.INPUT_LOG_DIRECTORY):
+            if name.startswith(basename):
+                base, sep, counter = name.partition('-')
+                if base != basename:
+                    logger.error('Bad log file name: %s', name)
+                    continue
+
+                self.file_counter = max(self.file_counter, int(counter))
+
+                stat_result = os.stat(name)
+                self.log_files.append({'name': name,
+                                       'created': stat_result.st_mtime,
+                                       'size': stat_result.st_size})
+
+        self.log_files = sorted(self.log_files,
+                                key=lambda x: x['created'], reverse=True)
+        self.fp = open(log_files[0]['name'], 'ab')
+
+    def ensure_log_file_ready(self):
+        if not self.log_files:
+            name = '{0}-{1}'.format(self.basename, self.file_counter)
+            # We handle the encoding ourselves and write to fp in
+            # binary mode so that a) we can guarantee consistent
+            # encoding of the log files and b) we can count the number
+            # of bytes we're writing.
+            self.fp = open(name, 'ab')
+            stat_resul
 
 
 INPUT_LOG = None
--- a/quipucords/scanner/input_log.py
+++ b/quipucords/scanner/input_log.py
@@ -86,7 +86,7 @@ def log_fact(host, fact, value):
 
 def disable_log_for_test():
     """Disable the input log for testing purposes."""
-    global INPUT_LOG
+    global INPUT_LOG  # pylint: disable=global-statement
     INPUT_LOG = io.StringIO()
 
 
@@ -102,7 +102,5 @@ def log_ansible_result(result):
         log_fact(result._host.name,
                  {'type': 'shell', 'command': args[RAW_PARAMS]},
                  result._result)
-    else:
-        print('*** Ansible result:', result)
     # If _raw_params isn't in args, then args is not a raw command and
     # we don't need to record it. This occurs for set_facts tasks.
--- a/quipucords/scanner/input_log.py
+++ b/quipucords/scanner/input_log.py
@@ -9,19 +9,22 @@
 
 """Log all input to the filesystem for later debugging."""
 
-import io
 import json
 import logging
 import multiprocessing as mp
 import os
+import time
 import uuid
 from django.db import transaction
 from api import log_state
 from quipucords import settings
 
-logger = logging.logger(__name__)
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
 UUID_CACHE = None
+
+DEFAULT_MAX_SIZE = 1 << 30  # 1 GiB
+DEFAULT_MAX_AGE = 365 * 24 * 60 * 60  # 1 year, except on leap years
 
 
 @transaction.atomic()
@@ -60,6 +63,7 @@ def next_sequence_number():
     return number
 
 
+# pylint: disable=too-many-instance-attributes
 class RotatingLogFile(object):
     """A log file that automatically rotates itself when full."""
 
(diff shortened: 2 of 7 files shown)


chambridge:
Looks like you have build failures. I see linting problems. In the meantime can you please include a zip of a log that has several scans of overlapping systems, so we can see the output.

quipucords/quipucords/settings.py
```diff
@@ -283,6 +283,8 @@
     },
 }
 
+INPUT_LOG_BASENAME = os.path.join(BASE_DIR, 'input_log')
```
chambridge: I think this should go into the specified logs directory when running in production.
I also think we need to improve the name for the variable and log its self, as it is not very descriptive. Can we also use `.log` extension instead of underscore? This will help with auto suggest tools that read logs.

Is there any mechanism to toggle this function ON/OFF via environment variable?

quipucords/scanner/input_log.py
```diff
@@ -0,0 +1,301 @@
+# Copyright (c) 2018 Red Hat, Inc.
+#
+# This software is licensed to you under the GNU General Public License,
+# version 3 (GPLv3). There is NO WARRANTY for this software, express or
+# implied, including the implied warranties of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv3
+# along with this software; if not, see
+# https://www.gnu.org/licenses/gpl-3.0.txt.
+
+"""Log all input to the filesystem for later debugging."""
```
chambridge: Can we have a better name than `input_log` I don't think it is very descriptive. What input? Its really more a recorder or logger.

quipucords/scanner/input_log.py
```diff
@@ -0,0 +1,301 @@
...
+import uuid
+from django.db import transaction
+from api import log_state
+from quipucords import settings
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+UUID_CACHE = None
+
+DEFAULT_MAX_SIZE = 1 << 30  # 1 GiB
```
chambridge: How do you specify a different max size?
How large is the file after a single scan of our 21 systems?

quipucords/scanner/input_log.py
```diff
@@ -0,0 +1,301 @@
...
+DEFAULT_MAX_AGE = 365 * 24 * 60 * 60  # 1 year, except on leap years
```
chambridge: How can this be changed? Environment variable?

quipucords/scanner/input_log.py
```diff
@@ -0,0 +1,301 @@
...
+
+
+@transaction.atomic()
+def get_sonar_uuid():
```
chambridge: Let's not use "Sonar" its an internal initiative name.

quipucords/scanner/input_log.py
```diff
@@ -0,0 +1,301 @@
...
+
+    for name in os.listdir(dirname):
+        if name.startswith(prefix):
+            result.append((name, os.stat(os.path.join(dirname, name))))
+
+    return result
+
+
+# pylint: disable=too-many-instance-attributes
+class RotatingLogFile(object):
```
chambridge: Feel like we could have used `RotatingFileHandler`, instead of inventing our own.
I think we care more about the size of the data then whether it ages out after a year. How much data do we generate for a single system today?
I'm interested in understanding what we think an average amount would be, so we know when we'd hit a GB.

quipucords/scanner/network/connect_callback.py
```diff
@@ -40,6 +41,7 @@ def __init__(self, result_store, credential, display=None):
 
     def v2_runner_on_ok(self, result):
         """Print a json representation of the result."""
+        input_log.log_ansible_result(result)
```
chambridge: May want to try/catch around this so any IO or other related failure in this flow doesn't kill a scan.

quipucords/scanner/network/connect_callback.py
```diff
@@ -52,6 +54,7 @@ def v2_runner_on_ok(self, result):
 
     def v2_runner_on_unreachable(self, result):
         """Print a json representation of the result."""
+        input_log.log_ansible_result(result)
```
chambridge: May want to try/catch around this so any IO or other related failure in this flow doesn't kill a scan.

quipucords/scanner/network/connect_callback.py
```diff
@@ -65,6 +68,7 @@ def v2_runner_on_unreachable(self, result):
 
     def v2_runner_on_failed(self, result, ignore_errors=False):
         """Print a json representation of the result."""
+        input_log.log_ansible_result(result)
```
chambridge: May want to try/catch around this so any IO or other related failure in this flow doesn't kill a scan.

quipucords/scanner/network/inspect_callback.py
```diff
@@ -105,6 +106,7 @@ def v2_runner_on_failed(self, result, ignore_errors=False):
 
     def handle_result(self, result):
         """Handle an incoming result object."""
+        input_log.log_ansible_result(result)
```
chambridge: May want to try/catch around this so any IO or other related failure in this flow doesn't kill a scan.

quipucords/scanner/network/inspect_callback.py
```diff
@@ -191,6 +193,7 @@ def _finalize_host(self, host, host_status):
     @transaction.atomic
     def v2_runner_on_unreachable(self, result):
         """Print a json representation of the result."""
+        input_log.log_ansible_result(result)
```
chambridge: May want to try/catch around this so any IO or other related failure in this flow doesn't kill a scan.

noahl:
@chambridge here is an example log file.
[input_log-1.zip](https://github.com/quipucords/quipucords/files/1732517/input_log-1.zip)

I don't think you can determine what facts come from what scan jobs right now. I made a point not to build the scan job into the way we identify data, because I thought the scan job structure might change at some point. However, I could add a tag to every record saying what scan job (and scan task for that matter) it is from. What do you think?

quipucords/scanner/input_log.py
noahl: I looked at `RotatingFileHandler`, but it seems to be very tied to the `logging` module's API where there are a record class and a formatter class that work together with the handler. We would have to write a custom formatter to get the output we want, and package up our output in a way that is not really what the logging module expects. I can do it that way if you want, it just seemed simpler to write this myself. (https://github.com/python/cpython/blob/3.6/Lib/logging/__init__.py)

quipucords/scanner/input_log.py
noahl: How about `remote_data_log`? I'm not tied to either name., I don't think they're that good, but I've been trying to think of a better name and I'm drawing a blank right now.

quipucords/scanner/input_log.py
noahl: Moved this and max_age to `settings.py` and added the ability to override with an environment variable.

quipucords/api/log_state.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class LatestSequenceNumber(models.Model):
+    """The latest sequence number."""
+
+    number = models.BigIntegerField()
+
+
+class DatabaseUUID(models.Model):
+    """The UUID of this Sonar installation."""
```
chambridge: Remove Sonar from here.

quipucords/api/log_state.py
```diff
@@ -0,0 +1,23 @@
+# Copyright (c) 2018 Red Hat, Inc.
+#
+# This software is licensed to you under the GNU General Public License,
+# version 3 (GPLv3). There is NO WARRANTY for this software, express or
+# implied, including the implied warranties of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv3
+# along with this software; if not, see
+# https://www.gnu.org/licenses/gpl-3.0.txt.
+"""Store the input log state in the database."""
```
chambridge: update with final name here.

quipucords/scanner/input_log.py
noahl: I got about 2.3MB of log from scanning all of `10.10.[181:182].[0:255]`. That scan found 24 reachable IPs, so very roughly, 100KB/machine/scan.

quipucords/scanner/input_log.py
```diff
@@ -0,0 +1,301 @@
...
+
+    return uuid_obj.uuid
+
+
+# This is a potential performance bottleneck since we are touching the
+# database every time we write a record. This isn't necessary - we
+# could get the state of the counter by reading the last few entries
+# of the existing log files on startup. I don't plan to implement that
+# unless we find that this is actually a problem.
+@transaction.atomic()
```
chambridge: What is all the UUID stuff providing us at this point? I can only see it making it harder to compress the file since it will be filled with UUIDs. I know down the line you want it for replaying, but I think things will likely change to much between now and then and even this will not be backwards compatible with where we end up.

quipucords/scanner/input_log.py
chambridge: It just might solve many of the the flaws you pointed to below and is well tested.

quipucords/scanner/input_log.py
noahl: Changed to `get_database_uuid`. I forgot to do that earlier; thanks for catching it.

quipucords/quipucords/settings.py
noahl: I'll move the logs to the right directory as soon as we agree on a name for them. :)

quipucords/api/log_state.py
noahl: Done, thanks.

chambridge:
I think we want context to help us utilize this for debugging. I think it would be valuable to have data around the things like:
- Scan Job
- Scan Task
- Source
- Credential

I could see were we want to differentiate between a host that was scanned with different credentials.

quipucords/scanner/input_log.py
noahl: Fair point. The UUIDs in the log file will all be the same, so they should compress pretty well. We don't need the UUIDs for replaying; they're there to make sure that we don't accidentally mix records from two different installations on our end. As long as they have different UUIDs, we can at least notice that the UUIDs don't match.

quipucords/scanner/input_log.py
noahl: Added an environment variable.

noahl:
@chambridge for the `LogFileHandler`, I spent a while looking at how the built-in one works. It's the same idea as what I wrote - it has an open file handle and a lock. To write a record, it acquires the lock, writes to the file, flushes the file, and releases the lock. It also has the same weaknesses - it doesn't support multiple processes writing to the same log file or using multiple instances of the class for the same log file within the same process. (The point I made about not rotating logs unless there are writes is moot, because it doesn't try to delete logs that are too old.)

I think I could make it work. The log file writer itself would be better-tested, but it would mean introducing some complexity around making our log file format fit into its interfaces, which would open up some opportunity for bugs there. I think we should not try to use the built-in one, but I don't feel super strongly if you really want to use it.

noahl:
@chambridge about the other points you made, I'm part-way through plumbing the scan job, scan task, source, and credential information through. A few questions:
  - For scanjob and scantask, do you mean just the numerical IDs of the ScanJob and ScanTask?
  - For the source, we already record the host we're running against. Are you saying that you want to user-visible name of the source too?

Thanks!

chambridge:
> For scanjob and scantask, do you mean just the numerical IDs of the ScanJob and ScanTask?
Yes, IDs are fine.

> For the source, we already record the host we're running against. Are you saying that you want to user-visible name of the source too?
A host could exist in multiple sources that have different credentials. Could be overlapping IP ranges, etc.

abaiken:
I always try to go with existing tools rather than write my own, but in this case it seems like the built-in one might have the same flaws as the one you have written. If the built-in LogFileHandler gives us any advantages over the current(other than being well-tested) than you should go with that one. Otherwise, if it would take more work to try and use it than use the one that you have written, I would just stick with the current (ie. I would do whatever will save us the most work/be the most effective now & in the future. But both options seem to be really similar, other than that one is well-tested and needs tweaking vs. one that already works so I don't have a strong preference...)

kholdaway:
I also think we should use the build in logging functions rather than creating our own.  Logging frameworks tend to be mature in most languages.  I would stick with python provided libraries.

Additionally, there are many ways to age out data.  I worry that a year of data is way too much to consume for a given restore task.  It may be worth exploring different approaches to partition data.

noahl:
Okay, I see everyone's points above. I'll start working on using `RotatingLogHandler` for the logs.

While I was working on plumbing the extra information Chris was asking about, I discovered that adding the credential to the log is going to be a lot of work - much more than the other three pieces of information combined. If it's okay with everyone, I'd like to do that in a separate PR just to keep the review manageable.

noahl:
Never mind about credentials, I found an easier way to handle them. They'll be in this PR, just need to finish the rest of the code.

quipucords/scanner/network/connect_callback.py
noahl: Done, via a helper function.

quipucords/scanner/network/connect_callback.py
noahl: Done.

quipucords/scanner/network/connect_callback.py
noahl: Done.

quipucords/scanner/network/inspect_callback.py
noahl: Done.

quipucords/scanner/input_log.py
noahl: Or maybe `scan_data_log`.

noahl:
I just pushed a major update which I think resolves all of the comments here. I also rebased on master. Please take a look :)

quipucords/quipucords/settings.py
```diff
@@ -283,6 +283,10 @@
     },
 }
 
+SCAN_DATA_LOG_BASENAME = os.path.join(BASE_DIR, 'scan_data_log')
```
chambridge: This is ok as a default, but it should be configurable like the other logs so it can end up in the same directory as the logs in the container setup. See line 205.

quipucords/scanner/network/connect_callback.py
```diff
@@ -32,14 +33,16 @@ class ConnectResultCallback(CallbackBase):
     scan, as we scan it.
     """
 
-    def __init__(self, result_store, credential, display=None):
+    def __init__(self, result_store, credential, scan_task, display=None):
```
chambridge: Are you finding any value in capturing the connect data? I hesitate to collect this since it has no input on the actual facts.

quipucords/scanner/scan_data_log.py
```diff
@@ -0,0 +1,188 @@
...
+    record['database_uuid'] = str(get_database_uuid())
+    record['sequence_number'] = next_sequence_number()
+
+    if not _DRY_RUN:
+        _HANDLER.emit(record)
+
+
+# pylint: disable=too-many-arguments
+def log_fact(host, fact, value, scan_job, scan_task, source):
+    """Write a new fact to the input log.
```
chambridge: input log --> scan data log ; multiple places.

quipucords/scanner/test_scan_data_log.py
```diff
@@ -0,0 +1,43 @@
+# Copyright (c) 2018 Red Hat, Inc.
+#
+# This software is licensed to you under the GNU General Public License,
+# version 3 (GPLv3). There is NO WARRANTY for this software, express or
+# implied, including the implied warranties of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv3
+# along with this software; if not, see
+# https://www.gnu.org/licenses/gpl-3.0.txt.
+
+"""Test the input logger."""
```
chambridge: rename from input logger.

quipucords/quipucords/settings.py
noahl: Done.

quipucords/scanner/network/connect_callback.py
noahl: I captured it for completeness, but it might be interesting if we have a scan that's not connecting to a particular host to see if the connect scan status was failed or unreachable. I don't feel strongly about this if you want to get rid of it, though.

quipucords/scanner/scan_data_log.py
noahl: Fixed in multiple places, thanks!

quipucords/scanner/test_scan_data_log.py
noahl: Fixed, thanks!

quipucords/quipucords/settings.py
noahl: Right now it's a separate environment variable, to match the other log. What do you think of the idea of either having a LOG_DIRECTORY variable where all the logs go (and still allowing overrides for each type of log) or making this default to the same directory as DJANGO_LOG_FILE but with a different name?

noahl:
Attaching a zipped log from the latest version. This includes several overlapping systems.

It came in at around 147kB/system uncompressed, ~6kB/system compressed.
[scan_data_log.zip](https://github.com/quipucords/quipucords/files/1749450/scan_data_log.zip)

quipucords/quipucords/settings.py
chambridge: I think its fine to do it with the LOG_DIRECTORY env. I didn't see a separate env var set for it.
Is there a variable to disable this function?

quipucords/scanner/network/connect_callback.py
chambridge: Reviewing the log. I don't see much value in this, I'd skip it for now.

chambridge:
After reviewing the attached log I have a couple of question.
What is the current value of the database uuid and sequence number? I feel like its projecting into the future, for a feature that will likely change and cause an unnecessary migration. And best to defer it until later.

Thoughts @abaiken @kholdaway?

noahl:
@chambridge The uuid and sequence number together are a unique identifier for every input fact. I put them in so we could use them later to identify which database records are affected by each input fact, which we could use to update the database correctly when we reingest input.

Fine with me to take them out for now.

quipucords/quipucords/settings.py
noahl: I didn't upload my latest changes, sorry. I'll push again in a moment with a LOG_DIRECTORY, which can be overridden for the individual log files.

For disabling, how about setting the scan data log file path to "" disables it?

quipucords/scanner/network/connect_callback.py
noahl: Will do.

kholdaway:
@noahl Given the uuid and sequence number are not used and aimed at a multi-tenant logging solution, I suggest we leave them out for now.  Typically, its best to design for something after its well known.  I suspect that we will have a multi-tenant logging requirement at some point.  When that happens, we will also know the exact datastore and capabilities we can leverage to segment data per tenant.

noahl:
@kholdaway agreed about leaving things out until they're needed, but keep in mind that the original design doc for this included the full multi-tenant logging solution. So I think this is not "designing for a future hypothetical" as much as "adding a feature needed for the next step in the plan". I am fine leaving it out until it's really needed, though.

quipucords/quipucords/settings.py
chambridge: I'd prefer an explicit DISABLE env ans this will be more intuitive for a user than setting an empty string. If they for some reason want to disable this log capture.

noahl:
@chambridge I just pushed a new commit that I think fixes all of your comments. Please take a look.

noahl:
Sorry, just a minute, need to add the DISABLE variable.

noahl:
@chambridge just pushed the DISABLE environment variable. This branch is ready for review.

quipucords/quipucords/settings.py
```diff
@@ -201,9 +201,18 @@
 LOGGING_HANDLERS = os.getenv('DJANGO_LOG_HANDLERS', 'console').split(',')
 VERBOSE_FORMATTING = '%(levelname)s %(asctime)s %(module)s ' \
     '%(process)d %(thread)d %(message)s'
-DEFAULT_LOG_FILE = os.path.join(BASE_DIR, 'app.log')
+
+LOG_DIRECTORY = os.getenv('LOG_DIRECTORY', BASE_DIR)
+DEFAULT_LOG_FILE = os.path.join(LOG_DIRECTORY, 'app.log')
 LOGGING_FILE = os.getenv('DJANGO_LOG_FILE', DEFAULT_LOG_FILE)
 
+DEFAULT_SCAN_DATA_LOG = os.path.join(LOG_DIRECTORY, 'scan_data_log')
```
chambridge: Will the log file end with the '.log' suffix? and 'scan_data_log' is a base name for the rotating filename?

chambridge: Also maybe end in `_FILE` to stay consistent with `DEFAULT_LOG_FILE`.

quipucords/quipucords/settings.py
```diff
@@ -201,9 +201,18 @@
...
+SCAN_DATA_LOG = os.getenv('SCAN_DATA_LOG', DEFAULT_SCAN_DATA_LOG)
```
chambridge: Again just for consistency maybe better as `SCAN_DATA_LOG_FILE`.

quipucords/scanner/network/connect_callback.py
chambridge: Were you going to pull this out?

quipucords/quipucords/settings.py
noahl: Just renamed to end in `.log` instead of `_log`. The `RotatingFileHandler` will make files `scan_data.log.1`, `scan_data.log.2`, etc. for old logs.

noahl: And changed to end in `_FILE`.

quipucords/quipucords/settings.py
noahl: Done.

chambridge:
🎉 🥇

PR #757, opened 2018-02-16
Add most recent connection task data to a source. Closes #610.

### LINKED ISSUE #610: Add latest connection scan reference to source
## Specify type:
- Enhancement

## Description:
UI requires information related to a source's most recent connection scan (if one exists).

Source view should be updated (for list & retrieve) to obtain the most recent connection scan.

Swagger documentation should be updated for `SourceOut` to include an object property `connection` with the following information: `id, start time, end time, status, success_systems, failed_systems`

If no connection scan is found this property will be omitted.
________
Related to issue #https://github.com/quipucords/quipucords/issues/607

chambridge:
FYI
@quipucords/ux-ui

### DIFF
--- a/quipucords/api/source/tests_source.py
+++ b/quipucords/api/source/tests_source.py
@@ -11,12 +11,15 @@
 """Test the API application."""
 
 import json
+from datetime import datetime
 from unittest.mock import patch
 from django.test import TestCase
 from django.core.urlresolvers import reverse
 from rest_framework import status
-from api.models import Credential, Source
+from api.models import (Credential, Source, ScanTask, ScanJob)
 import api.messages as messages
+from api.source.view import format_source
+from api.serializers import SourceSerializer
 
 
 def dummy_start():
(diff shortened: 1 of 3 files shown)


quipucords/api/source/tests_source.py
```diff
@@ -118,6 +121,50 @@ def create_expect_400_with_query(self,
...
+    def test_format_source(self):
+        """Test the format source method."""
+        start = datetime.now()
+        source = Source(
+            name='source1',
+            source_type='network',
+            port=22)
+        source.save()
+        end = datetime.now()
+        scan_task = ScanTask(source=source,
```
kholdaway: You should use https://github.com/quipucords/quipucords/blob/master/quipucords/scanner/test_util.py#L23 to create a test job/task.  There are a lot of dependencies now in the code for state.

quipucords/api/source/view.py
```diff
@@ -39,12 +39,44 @@ def expand_credential(json_source):
...
+    :param json_source: JSON source data from serializer
+    :returns: JSON data
+    """
+    json_source = expand_credential(json_source)
+    source_id = json_source.get('id')
+
+    scan_task_qs = ScanTask.objects.filter(
+        source__id=source_id,
+        scan_type=ScanTask.SCAN_TYPE_CONNECT)
+    if scan_task_qs:
```
kholdaway: I think this will not be null if if the results are empty.  You may want to move scan_task_qs.latest('start_time') above this check.

quipucords/api/source/tests_source.py
chambridge: updated

quipucords/api/source/view.py
chambridge: put in a guard with `count() > 0`

PR #761, opened 2018-02-17
UI: Use getUser and logout API services

jeff-phillips-18:
Updates issue #690


client/package.json
```diff
@@ -28,6 +28,7 @@
     "build:ui": "run-s build:css build:js",
     "build:asset-deps": "mkdir -p public/assets/rcue && cp -R node_modules/rcue/dist/* public/assets/rcue",
     "build:asset-images": "mkdir -p public/assets/images && cp -R src/styles/images/* public/assets/images",
+    "build:asset-css": "npm run build:css && mkdir -p public/assets/css && cp -R src/styles/css/* public/assets/css",
```
jeff-phillips-18: This does cause the css to get built twice, once or the the build:ui and once here for build:assets-css. Not sure the best way around this since build:js does the copy from build to public so the public/asses/css must be done before that. build:ui should build all the local ui stuff (including css) but build:assets should build the complete set of assets.

cdcabrera: Gotcha. Had been looking at integrating the CSS/SCSS/LESS into the Create React App setup (using imports on JS files). That would definitely affect this, possibly making it so the ```build:ui``` has to run since CSS and JS would be integrated. 

I think what you have is good for this round, and it gets us thinking about what needs to take place, next round. 

From what I can gather all we really need is a link to the CSS and possibly images from the Django side. We could keep it simple and manually link those resources once the build gets checked in. Or the slightly more complicated opposite end approach... and we could look at integrating the Django templates into our side under the public directory. So instead those get copied over. They wouldn't necessarily run on our side, but we could use the Create React App to fill in the paths for us... if it writes them out as strings.

quipucords/quipucords/templates/registration/logged_out.html
```diff
@@ -1,10 +1,70 @@
...
 
+{% load static %}
+
+{% block includes %}
+<link rel="stylesheet" href="{% static '/assets/rcue/css/rcue.css' %}">
+<link rel="stylesheet" href="{% static '/assets/rcue/css/rcue-additions.css' %}">
+<link rel="stylesheet" href="{% static '/assets/css/entitlements.css' %}">
+{% endblock %}
+
 {% block page-title %}
```
cdcabrera: Quick fix ```{% block pagetitle %}```

quipucords/quipucords/templates/registration/logged_out.html
```diff
@@ -1,10 +1,70 @@
...
 Red Hat Entitlements Reporting
 {% endblock %}
 
+{% block logo %}
```
cdcabrera: Quick fix ```{% block pagelogo %}```

quipucords/quipucords/templates/registration/logged_out.html
jeff-phillips-18: Thanks. Fixed.

quipucords/quipucords/templates/registration/logged_out.html
jeff-phillips-18: Thanks. Fixed.

client/package.json
```diff
@@ -46,6 +47,7 @@
     "axios": "^0.17.1",
     "bootstrap": "^3.3.7",
     "js-cookie": "^2.2.0",
+    "moment": "^2.20.1",
```
cdcabrera: I see moment implemented, but not used anywhere?

client/src/components/app.js
```diff
@@ -61,60 +58,45 @@ class App extends React.Component {
...
+        <EmptyState className="full-page-blank-slate">
           <Alert type="error">
             <span>Login error: {session.errorMessage}</span>
           </Alert>
         </EmptyState>
       );
     }
 
-    if (session.pending || !session.fulfilled) {
+    if (session.pending || !session.fulfilled || (!session.loggedIn && !session.wasLoggedIn)) {
```
cdcabrera: Optional: looks like Prettier was trying to run on this page, didn't complete its task. 

Was going to ask, we can look at expanding the column width to 100 or 120, right now prettier has a default for 80 columns, you be up for it if we increased it?

client/src/services/userService.js
```diff
@@ -1,11 +1,12 @@
+import axios from 'axios';
 import cookies from 'js-cookie';
+import { helpers } from '../common/helpers';
 
 class UserService {
   static authorizeUser() {
-    // ToDo: ReEvaluate placement of this spoof for auth. Also consider using a helper function.
-    if (process.env.REACT_APP_ENV === 'development') {
+    // ToDo: ReEvaluate placement of this spoof for auth
+    if (helpers.DEV_MODE) {
```
cdcabrera: I'm having trouble finding this in helpers?

client/package.json
jeff-phillips-18: soon... it will be used in sources to show most recent scan. I can remove for now and add back if you like.

client/src/components/app.js
jeff-phillips-18: totally up for it!

client/src/services/userService.js
jeff-phillips-18: bah... also coming soon. Multi-tasking.... will fix.

client/package.json
cdcabrera: That works, lets go for it

client/src/services/userService.js
```diff
@@ -22,37 +22,23 @@ class UserService {
...
-      response =>
-        new Promise(resolve => {
-          resolve({
-            userName: arr[Math.floor(Math.random() * arr.length)]
-          });
-        })
-    );
+    return axios({
+      method: 'get',
+      url: `${process.env.REACT_APP_USER_SERVICE}current/`,
```
cdcabrera: Optional: Didn't catch these before, and we can iterate on it... but the full url, probably needs to be placed in the ```dotenv``` files

client/src/services/userService.js
```diff
@@ -22,37 +22,23 @@ class UserService {
...
-          }
-
-          throw new Error(
-            `Error logging out, token doesn't exist for ${response.userName}.`
-          );
-        })
-    );
+    return axios({
+      method: 'put',
+      url: `${process.env.REACT_APP_USER_SERVICE}logout/`,
```
cdcabrera: Optional: Same comment about moving this to the ```dotenv``` files

quipucords/client/assets/css/entitlements.css
```diff
@@ -0,0 +1,251 @@
+/**
```
cdcabrera: Looks like additional files have been added under ```quipucords/client/*``` that should probably be removed if we're aiming for an on-the-fly type of product compile... #764 ... though this particular file does align partially with that concept because of the login templates. I'm partial to removing these files since they'll be updated as we update the UI build

jeff-phillips-18:
Fixed the paths to be in the `dotenv` files.
Removed the mistakenly added files.

PR #765, opened 2018-02-19
Modify ScanJob to pull data from Scan model

kholdaway:


quipucords/api/messages.py
```diff
@@ -81,7 +81,10 @@
 # scan jobs messages
 PLURAL_SCAN_JOBS_MSG = 'Scan Jobs'
 SJ_REQ_SOURCES = 'Scan job must have one or more sources.'
+SJ_SCAN_IDS_INV = 'Scan identifiers must be integer values.'
```
chambridge: Scan job identifiers ...

quipucords/api/messages.py
```diff
@@ -81,7 +81,10 @@
...
 SJ_SOURCE_IDS_INV = 'Source identifiers must be integer values.'
+SJ_SCAN_DO_NOT_EXIST = 'Scan with id=%d could not be'\
```
chambridge: Scan job with id ...

quipucords/api/scanjob/serializer.py
```diff
@@ -90,24 +113,26 @@ class Meta:
...
-            options = ScanJobOptions()
-        options.save()
-        scanjob.options = options
-        scanjob.save()
-
-        return scanjob
+        scan_job = super().create(validated_data)
+
+        # Copy scan config into job to capture config for historic reference
+        copy_scan_info_into_job(scan_job)
```
chambridge: Maybe consider some kind of `post_init()` method on the ScanJob model, instead of this in the serializer?

quipucords/api/source/view.py
```diff
@@ -152,15 +160,23 @@ def create(self, request, *args, **kwargs):
                     # Grab the source id
                     source_id = response.data['id']
                     # Define the scan options object
-                    scan_options = ScanJobOptions()
```
chambridge: I'd rather have just a ScanJob, so no changes.

quipucords/api/messages.py
kholdaway: These are right.  SJ_ prefix is for messages in the ScanJob classes and its referencing the Scan not the Scan job identifier.

quipucords/api/messages.py
kholdaway: These are right.  SJ_ prefix is for messages in the ScanJob classes and its referencing the Scan not the Scan job identifier.

quipucords/api/scanjob/serializer.py
kholdaway: This is now done when a job is queued.  Basically, the scan configuration is captured when the job is ready to run.

kholdaway: I pushed a change for this

quipucords/api/source/view.py
kholdaway: Done.

PR #774, opened 2018-02-20
fix up eslint and tests

jeff-phillips-18:
Updates #690


cdcabrera:
Just a couple of notes about 
- the dotenv references, think we can use something similar to the NPM script, but for tests

- and iterating on the way the action dispatch objects have to get laid out for testing.

cdcabrera:
Approved!

PR #775, opened 2018-02-20
Handle invalid or expired tokens for all CLI requests. Closes #771.

### LINKED ISSUE #771: Invaild Token or Expired Token not handled by CLI for any secondary request
## Specify type:
- Bug

- High

## Description:
If you utilize a command that performs a secondary request (e.g. uses a name to get an id), these commands are not currently wrapped in any flow that handles these Token issues so you can get incorrect messages.

___

### Expected behavior:
CLI secondary commands should respond with proper error for token issues.

### Actual behavior:
Incorrect error message

chambridge:


qpc/request.py
```diff
@@ -26,6 +27,43 @@
...
+    exception_class = ValueError
+
+
+def handle_token_errors(response):
+    """Handle token errors.
+
+    :param response: The response object.
+    :returns: The response object.
+    """
+    token_expired = {'detail': 'Token has expired'}
```
kholdaway: Will 'Token has expired' ever be translated?

chambridge: No, third party tool doesn't.
https://github.com/JamesRitchie/django-rest-framework-expiring-tokens/blob/master/rest_framework_expiring_authtoken/authentication.py#L34

kholdaway: 👍

PR #780, opened 2018-02-21
Add full update to scan

kholdaway:


quipucords/api/scan/serializer.py
```diff
@@ -89,6 +89,28 @@ def create(self, validated_data):
 
         return scan
 
+    @transaction.atomic
+    def update(self, instance, validated_data):
+        """Update a scan."""
+        # If we ever add optional fields to Scan, we need to
+        # add logic here to clear them on full update even if they are
+        # not supplied.
```
abaiken: Scan has optional fields (disable_optional_products & max_concurrency) Instead of adding logic to clear them on full updates, wouldn't omitting them just be a partial update (once it is written)?

PR #781, opened 2018-02-21
UI: Add scan action and reducer tests

jeff-phillips-18:
Updates #690

### DIFF
--- a/client/src/redux/actions/tests/scansActions.test.js
+++ b/client/src/redux/actions/tests/scansActions.test.js
@@ -1,15 +1,19 @@
-import configureMockStore from 'redux-mock-store';
 import promiseMiddleware from 'redux-promise-middleware';
-import moxios from 'moxios';
+import { createStore, applyMiddleware, combineReducers } from 'redux';
 import expect from 'expect';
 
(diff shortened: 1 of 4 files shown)


client/src/redux/actions/tests/scansActions.test.js
```diff
@@ -53,14 +53,17 @@ describe('ScansActions', function() {
       });
     });
```
cdcabrera: Note: Next round we could look at moving the ```moxios.wait...``` up to the ```beforeEach```, unless of course it needs to be customized

client/src/redux/actions/tests/scansActions.test.js
```diff
@@ -1,15 +1,19 @@
...
 import promiseMiddleware from 'redux-promise-middleware';
-import moxios from 'moxios';
+import { createStore, applyMiddleware, combineReducers } from 'redux';
 import expect from 'expect';
 
-import { scansTypes } from '../../constants';
 import * as actions from '../scansActions';
+import { reducers } from '../../reducers';
+
+// const mockStore = configureMockStore(middlewares);
```
cdcabrera: Note: next round

PR #805, opened 2018-02-22
UI: update scans list and source list with data

jeff-phillips-18:
Updates #690 

Shows correct scan status/time
Shows correct hosts when expanding a source

### DIFF
--- a/client/src/components/scans/scanListItem.js
+++ b/client/src/components/scans/scanListItem.js
@@ -1,3 +1,5 @@
+import _ from 'lodash';
+import * as moment from 'moment';
 import React from 'react';
 import cx from 'classnames';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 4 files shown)


client/src/components/scans/scanListItem.js
```diff
@@ -90,14 +101,17 @@ class ScanListItem extends React.Component {
 
     let sourcesCount = item.sources ? item.sources.length : 0;
 
+    let successHosts = Number.isNaN(item.systems_scanned) ? 0 : item.systems_scanned;
+    let failedHosts = Number.isNaN(item.systems_failed) ? 0 : item.systems_failed;
```
cdcabrera: Note: This works, think this works too... 
``` 
let successHosts = item.systems_scanned || 0;
```

PR #809, opened 2018-02-22
UI: Update list views when items have 0 count

jeff-phillips-18:
Updates #690

### DIFF
--- a/client/src/components/listStatusItem/listStatusItem.js
+++ b/client/src/components/listStatusItem/listStatusItem.js
@@ -1,0 +1,76 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Icon, ListView } from 'patternfly-react';
+
+import { SimpleTooltip } from '../simpleTooltIp/simpleTooltip';
+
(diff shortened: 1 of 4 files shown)


client/src/components/listStatusItem/listStatusItem.js
```diff
@@ -0,0 +1,76 @@
...
+          {` ${text}`}
+        </span>
+      );
+    }
+  };
+
+  if (count > 0) {
+    return (
+      <ListView.InfoItem className="list-view-info-item-icon-count">
+        <SimpleTooltip id={`${id}_tip`} tooltip={`${count}  ${count === 1 ? tipSingular : tipPlural}`}>
```
cdcabrera: Optional: the id string/template literals, looks like the id is always supposed to be a string, we worried about someone jamming a number in there? if so we could just flip ```tip_${id}```

jeff-phillips-18: The propType is string. I could make it accept either, but if you pass an number you do get a console warning but it works OK.

PR #833, opened 2018-02-24
UI: Add Scans Jobs to expansion area

jeff-phillips-18:
Also adds responsive handling to masthead/NavBar

Updates #690


quipucords/client/index.html
```diff
@@ -1,84 +1 @@
-<!DOCTYPE html>
```
cdcabrera: This looks like an accidental left over bit?

jeff-phillips-18: yup, thanks

client/src/components/mastheadOptions/mastheadOptions.js
```diff
@@ -3,33 +3,15 @@ import PropTypes from 'prop-types';
...
 
 import { Dropdown, Icon, MenuItem } from 'patternfly-react';
-
-import Store from '../../redux/store';
-import { getUser, logoutUser } from '../../redux/actions/userActions';
-import { aboutTypes } from '../../redux/constants';
-import helpers from '../../common/helpers';
+import { getUser } from '../../redux/actions/userActions';
 
 class MastheadOptions extends React.Component {
```
cdcabrera: Note: At some point we could probably go full stateless with this component, turn it into a ```const``` instead of an ecma ```class```.

client/src/components/app.js
```diff
@@ -51,6 +57,15 @@ class App extends React.Component {
     });
   }
 
+  renderMenuActions() {
+    const { logoutUser } = this.props;
+
+    return [
```
cdcabrera: Note: Moving the masthead pieces into here was a good start. Not sure how feel about all the authentication bits spread out, maybe at some point a single function/filter inside of ```App``` to handle it would be nice.

Note: Missed this before, but tracking the ```props.wasLoggedIn``` value was interesting. It's only mentioned in one place, and not in the initial userState. Might be a good idea for consistency to place it in the initial userState.

client/src/components/scans/scanJobsList.js
```diff
@@ -0,0 +1,132 @@
...
+        icon = <Icon className="scan-job-status-icon fa-spin" type="fa" name="spinner" />;
+        scanTime = _.get(job, 'start_time');
+        break;
+      case 'paused':
+        scanDescription = 'Paused';
+        icon = <Icon className="scan-job-status-icon" type="pf" name="warning-triangle-o" />;
+        break;
+      default:
+        return null;
+    }
```
cdcabrera: Note: the switch might be an opportunity for some clean up next round, little repetition

client/src/components/scans/scanListItem.js
```diff
@@ -263,10 +280,12 @@ class ScanListItem extends React.Component {
...
       case 'sources':
         return <ScanSourceList scan={item} />;
-      case 'scans':
-        return <JSONPretty json={item} />;
+      case 'jobs':
+        return (
+          <ScanJobsList scan={item} onSummaryDownload={onSummaryDownload} onDetailedDownload={onDetailedDownload} />
+        );
       default:
-        return <JSONPretty json={item} />;
```
cdcabrera: Note: clean up next round, remove ```JSONPretty``` from package.json

PR #856, opened 2018-02-27
Check for sources that reference credential before delete

kholdaway:


quipucords/api/messages.py
```diff
@@ -30,6 +30,9 @@
 HC_NAME_ALREADY_EXISTS = 'Host credential with name=%s already exists'
 CRED_TYPE_REQUIRED_CREATED = 'cred_type is required for credential creation'
 CRED_TYPE_NOT_ALLOWED_UPDATE = 'cred_type is invalid for credential update'
+CRED_DELETE_NOT_VALID_W_SOURCES = 'Credential %s cannot be deleted '\
+    'because it is used by 1 or more sources.  '\
+    'First edit sources to remove credential %s.'
```
abaiken: Should the user be informed of which sources are using the cred?

kholdaway: It does.
```
{
    "message": "Credential 5 cannot be deleted because it is used by 1 or more sources.  First edit sources to remove credential 5.",
    "sources": [
        {
            "id": "7",
            "name": "NSingleSource"
        },
        {
            "id": "8",
            "name": "NSmallSource"
        },
        {
            "id": "9",
            "name": "NLargeSource"
        }
    ]
}
```

PR #861, opened 2018-02-27
Handle delete for source

### LINKED ISSUE #779: Make sure deletion of a cred, source, or scan leave the model in a working state
# Opening a new issue
___

## Specify type:
- Enhancement

## Description:
Deletion of items may leave the model in an invalid state.  For example, removing a cred will cascade the delete to the source foreign key (remove cred from the source list of creds), but that could leave the source without any creds.  When creating a source, we require at least one cred so this cred delete creates an invalid state for the source.

This issue should cover the deletion/update of a cred, source, scan and ensure the model works or surfaces an action to take to fix it.

kholdaway:


quipucords/api/messages.py
```diff
@@ -55,6 +55,9 @@
 SOURCE_TYPE_INV = 'A source_type must not be provided when updating a source.'
 SOURCE_CRED_IDS_INV = 'Credential identifiers must be integer values.'
 SOURCE_MIN_CREDS = 'Source must have at least one set of credentials.'
+SOURCE_DELETE_NOT_VALID_W_SCANS = 'Source %s cannot be deleted '\
+    'because it is used by 1 or more scans.  '\
+    'First edit scans to remove source %s.'
```
chambridge: `First edit scans to remove source %s.` --> `First edit the following scans to remove the source: %s`

kholdaway:
Comments addressed.

PR #866, opened 2018-02-27
Cast the job id as a string. Closes #864

### LINKED ISSUE #864: Scan Job needs to cast the job id as a string for the path.
## Specify type:
- Bug

- Medium
___

## Description:
In job.py the _build_req_params method is grabbing the id (an int) without casting it as a string, causing a test failure. 

### Steps to reproduce:
run `pattern=tests_scan_job.py make test-case`

abaiken:
When making the path in job.py, we need to cast the id as a string.

### DIFF
--- a/qpc/scan/job.py
+++ b/qpc/scan/job.py
@@ -72,7 +72,7 @@ class ScanJobCommand(CliCommand):
             found, scan_object_id = _get_scan_object_id(self.parser,
                                                         self.args.name)
             if found:
-                self.req_path += str(scan_object_id) + 'jobs/'
+                self.req_path += scan_object_id + 'jobs/'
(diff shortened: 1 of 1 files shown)


qpc/scan/job.py
```diff
@@ -72,11 +72,11 @@ def _build_req_params(self):
             found, scan_object_id = _get_scan_object_id(self.parser,
```
chambridge: maybe change it in the utility instead?

abaiken: It is already being cast as a string in the utility. Turns out only line 79 needs to be cast as a string.

kholdaway:
One to pad your stats

Issue #839, opened 2018-02-26
UI front-end sprint: API, view, & testing refinement


cdcabrera:
## Specify type:
- Enhancement
- UI

## Description:
- [x] Build & framwork updates #999 #1020 
- [x] Bug fixes #912 #931 
- [x] Testing updates - Jest snapshots carried over
  - Travis
  - Unit
- [x] Download Details & Summary Actions in Scans #926 
- [x] Form updates #931 #932 #972 #992 #999 #1019 
  - Wizard `931, 932, 972, 999 1019`
  - Credentials `992`
- [x] View refinement #873 #966 #970 #972 #980 #988 #992 #998 #1020 
  - About `980`
  - Login `972`
  - Sources `873`
    - Form updates
  - Scans `873`
  - Credentials `873`

Issue #827, opened 2018-02-23
UI: Add Source Wizard issues


jeff-phillips-18:
## Type:
- Bug

### Bug severity:
- Medium
___

## Description:
A couple of issues with the Add Source Wizard:

- Second page should auto focus on the source name field
- After adding a credential, the credential does not appear in the list
- Selecting a credential doesn't show it as selected when I bring up the list again. It does if I select another credential. But then that one doesn't show up.
- The last page of the wizard says I will get notified, I never do. We should probably just refresh the sources page on an add, it shows up then.

cdcabrera:
Relates to #690

cdcabrera:
Note: Issues found with the wizard

- Endpoint for editing a source needs to be fixed
- Wizard isn't limiting the credential types... they are specific and need to be filtered for "network", "satellite", and "vcenter"... I think this is the issue @bclarhk  came across.
- Looks like the validation on the wizard screws up after adding a credential, it needs to be reset for some reason, after blanking the fields it then fires again... 
- Need to fire a refresh when closing the wizard down, dependent on the view you're in

cdcabrera:
relates to #931

PR #873, opened 2018-02-28
UI: Add source wizard credentials & view styling

cdcabrera:
* Add source wizard available to Scans and Credentials views
* Make the dropdown button same size across Empty-States
* Source wizard step 2 patch
* Source wizard initial step 3 prep for stateless
* Credentials now refresh when added from the Add Source Wizard
* Relocated add source wizard to avoid refresh when components shift
* Design updates for source & scan views
* Popover added for hosts on source view
* Helper for generating an id

What's currently not in this PR, based on the design review.
- top/bottom equal padding for list row items. the padding appears equal already per the computed styling in dev tools
- equal padding around the popover due to scrolling, will continue to look at modifying it. and smaller screen review of the popover

updates #839 & #827


client/src/redux/actions/credentialsActions.js
```diff
@@ -15,10 +16,15 @@ const getCredential = id => dispatch => {
   });
 };
 
-const getCredentials = (query = {}) => dispatch => {
+const getCredentials = (query = {}) => (dispatch, getState) => {
```
jeff-phillips-18: It might be better to have a separate action for this `getAllCredentials` that could then dispatch a different type which wouldn't be picked up by the credentialsReducer and we wouldn't need to track wizard state here.

cdcabrera: Coincidentally, we did give a separate action a go as one of the first attempts to avoid having to pass the additional state with a limit on it, feels hacky along with the Reducer condition. We can take this approach, but a compare/contrast from what we saw...

The simple approach we add another action, a single constant, avoid creating a separate Action.js file. From there 2 possibilities pop up...

Either we place a new case in the Credentials reducer, or we place it in the Wizard reducer. 

1. Adding the case to the Credentials reducer means we can cause the Credentials view to refresh when it comes time to updating it. This is nice because...

2. Adding the case to the Wizard reducer means we have to specifically place logic just for the scenario of closing the Wizard in the Credentials view during the initial Empty State. There's a slim chance that could happen since we've placed all the necessarily buttons in there... jokingly we could remove the Add Source button from the Credentials view empty state and that would resolve the issue too =)

The part we tried was the second approach, and that didn't appear necessarily cleaner or an improvement on readability, just shifted the logic around. In working through the first approach, we may be able to eliminate a few bits (like the condition check in the Reducer) and make it more readable, slightly. We're game for that, we'll give it go, readability for the win!

PR #875, opened 2018-02-28
Add extended product search options to ScanOptions

### LINKED ISSUE #838: Provide API option for scans to enable extended product search
## Specify type:
- Enhancement

## Description:
Similar to the disable optional product search we will add a `enable_extended_product_search` and `extended_product_search_directories` to the Scan Options:
```
"enable_extended_product_search": {
      "jboss_eap": true,
      "jboss_fuse": true,
      "jboss_brms": true
    },
"extended_product_search_directories": [ "dirA", "dirB"]
```
The model should be default populated to `None` for both as this should only be added if specified from the CLI or API.  In code default scan directories will be `'/', '/opt', '/app', '/home', '/usr'` which is the same as `Rho`. https://github.com/quipucords/rho/blob/master/rho/scancommand.py#L165

## Acceptance Criteria: 
- [ ] Verify that API users can create a scan with a list of products for extended product search
- [ ] Verify that API users can edit a scan with a list of products for extended product search (this options edit is not additive; must supply full list of products)
- [ ] Verify that API users can create a scan with a list of scan directories only if extended product search is provided
- [ ] Verify that API users can edit a scan with a list of scan directories only if extended product search is provided or previously existed.

___
Associated with: https://github.com/quipucords/quipucords/issues/836

kholdaway:
Closes #838

### DIFF
--- a/docs/swagger.yml
+++ b/docs/swagger.yml
@@ -1306,7 +1306,7 @@ definitions:
         description: "The maximum concurrent host scans. default is 50."
       disabled_optional_products:
         $ref: "#/definitions/DisableOptionalProducts"
-      enable_extended_product_search:
+      enabled_extended_product_search:
         $ref: "#/definitions/ExtendedSearchProducts"
   Scan:
     type: "object"
--- a/docs/swagger.yml
+++ b/docs/swagger.yml
@@ -1349,16 +1349,6 @@ definitions:
             items:
               $ref: "#/definitions/ScanOut"
             description: "The paginated scan"
-
-  ScanJobOptions:
-    type: "object"
-    properties:
-      max_concurrency:
-        type: "integer"
-        format: "int64"
-        description: "The maximum concurrent host scans. default is 50."
-      disabled_optional_products:
-        $ref: "#/definitions/DisableOptionalProducts"
   ScanJob:
     type: "object"
     properties:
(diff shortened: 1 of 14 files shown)


docs/swagger.yml
```diff
@@ -1289,6 +1306,8 @@ definitions:
         description: "The maximum concurrent host scans. default is 50."
       disabled_optional_products:
         $ref: "#/definitions/DisableOptionalProducts"
+      enable_extended_product_search:
```
chambridge: `enable_extended_product_search` --> `enabled_extended_product_search`
Need to keep consistent with `disabled_optional_products`

quipucords/api/scan/serializer.py
```diff
@@ -16,31 +16,76 @@
...
+        """Validate search directories."""
+        try:
+            search_directories_list = json.loads(search_directories)
+            if not isinstance(search_directories_list, list):
+                raise ValidationError(
+                    _(messages.SCAN_OPTIONS_EXTENDED_SEARCH_DIR_NOT_LIST))
+            for directory in search_directories_list:
+                if not isinstance(directory, str):
+                    raise ValidationError(
+                        _(messages.SCAN_OPTIONS_EXTENDED_SEARCH_DIR_NOT_LIST))
```
chambridge: Worth adding a check like the following:
https://github.com/quipucords/rho/blob/master/rho/utilities.py#L578

Makes sure what was given was at minimum a "good" path structure.

docs/swagger.yml
kholdaway: Done.

quipucords/api/scan/serializer.py
kholdaway: Done.

PR #879, opened 2018-02-28
Add network postprocessing for EAP 5 facts

noahl:
In addition, this PR
  - Moves some common postprocessing code into `quipucords.scanner.network.processing.util`
  - Marks two tasks as internal, since we don't use their output outside of Ansible.

Here's some sample output from a scan of a host with EAP 5 installed. This is what's logged when the scanner creates a fact collection just before sending the fact collection to the fingerprinter. The output is pretty-printed and only has the EAP 5 facts, but otherwise is the raw debug logging:
```
{
  "eap5_home_candidates": [
    "/opt/jboss/jbosseap-5.2.0/jboss-eap-5.2"
  ],
  "eap5_home_version_txt": {
    "/opt/jboss/jbosseap-5.2.0/jboss-eap-5.2": true
  },
  "eap5_home_readme_html": {
    "/opt/jboss/jbosseap-5.2.0/jboss-eap-5.2": true
  },
  "eap5_home_ls_jboss_as": {
    "/opt/jboss/jbosseap-5.2.0/jboss-eap-5.2": [
      "JBossEULA.txt"
    ]
  },
  "eap5_home_run_jar_version": {
    "/opt/jboss/jbosseap-5.2.0/jboss-eap-5.2": true
  }
}
```

### DIFF
--- a/quipucords/scanner/network/inspect.py
+++ b/quipucords/scanner/network/inspect.py
@@ -42,6 +42,7 @@ DEFAULT_ROLES = [
     'etc_release',
     'file_contents',
     'jboss_eap',
+    'jboss_eap5',
     'jboss_brms',
     'jboss_fuse',
     'jboss_fuse_on_karaf',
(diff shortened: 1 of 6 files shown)


noahl:
@chambridge thanks!

PR #915, opened 2018-03-02
Add API to create a fact collection from two or more completed  ScanJobs

kholdaway:


quipucords/api/messages.py
```diff
@@ -108,6 +108,17 @@
 SJ_EXTRA_VARS_KEY = 'Extra vars keys must be jboss_eap, jboss_fuse, '\
     'or jboss_brms.'
 
+SJ_MERGE_JOB_REQUIRED = 'This field is required'
+SJ_MERGE_JOB_NOT_LIST = 'This field must be a list of job ids.'
+SJ_MERGE_JOB_TOO_SHORT = 'Two or more job ids are required.'
+SJ_MERGE_JOB_NOT_INT = 'Job ids must be integers.'
+SJ_MERGE_JOB_NOT_UNIQUE = 'Set of ids must be unique.'
+SJ_MERGE_JOB_NOT_FOUND = 'Not all job ids exist. Jobs note found: %s'
```
abaiken: Jobs note found --> jobs not found

quipucords/api/fact/util.py
```diff
@@ -33,6 +33,26 @@
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
 
+def build_sources_from_tasks(tasks):
```
chambridge: Not a great name. I know what you mean here, but its more like `build_fact_collection_data_from_tasks()`

quipucords/api/messages.py
chambridge: let's use scan job just to be specific.... everywhere!

kholdaway: Done for both

quipucords/api/fact/util.py
kholdaway: Its not quite that either.

PR #920, opened 2018-03-02
Redirect error output to /dev/null to avoid spurious output in raw facts. Closes #882.

### LINKED ISSUE #882: Ansible tasks using `ls` should guard against error related to access
## Specify type:
- Bug

- Low
___

## Description:
Currently you run a network scan against a system where the user has limited authority and access to different parts of the file system or the files you look for as indicators are not present you can see errors.

`ls: cannot access /etc/rhsm/facts: No such file or directory`

We should be sending error output to /dev/null:
`raw: ls /etc/rhsm/facts 2> /dev/null | grep .facts | wc -l`
___

### Expected behavior:
You should not be processing error output from commands.

### Actual behavior:
Potential spurious output may make its way into the raw facts.

chambridge:


roles/jboss_fuse_on_karaf/tasks/main.yml
```diff
@@ -64,15 +64,15 @@
...
+  raw: ls -1 "{{ item }}"/bin/fuse 2>/dev/null
   register: karaf_home_bin_fuse
   ignore_errors: yes
   with_items: "{{ karaf_homes }}"
   when: 'jboss_fuse'
 
 
 - name: check system/org/jboss/fuse
-  raw: ls -1 "{{ item }}/system/org/jboss"
+  raw: ls -1 "{{ item }}"/system/org/jboss 2>/dev/null
```
kdelee: I see on most of these you moved the quotes to just be around `"{{item}"`. Do they disappear after the templating has done its job?

roles/jboss_brms/tasks/main.yml
```diff
@@ -61,7 +61,7 @@
   when: 'jboss_brms'
 
 - name: look for kie-api files inside candidate directories
-  raw: ls -1 {{ item }}/WEB-INF/lib/kie-api*
+  raw: ls -1 {{ item }}/WEB-INF/lib/kie-api* 2>/dev/null
```
kdelee: Most of the other items have quotes around the "{{item}}", you think you need that here?

roles/jboss_fuse_on_karaf/tasks/main.yml
chambridge: yes the quotes disappear.

roles/jboss_brms/tasks/main.yml
chambridge: I'm gonna update this. It seems to work, but I prefer it to be consistent.

PR #922, opened 2018-03-02
Add report id as a lookup option for report summary/details

### LINKED ISSUE #851: Ability to combine multiple scans into a single report
## Specify type:
- Enhancement

## Description:
Right now a report can only be generated from a single scan job but it would be really helpful to enable a report generation from multiple scan jobs. 

Either have a merge task or be able to provide multiple scan job IDs to the report object.

kholdaway:
Closes #851


docs/source/man.rst
```diff
@@ -471,11 +471,15 @@ Viewing the Detail Report
...
 
-``--id=scan_job_identifier``
+``--scan-job=scan_job_identifier``
+
+  Required. Contains the scan job identifier for the scan that is used to generate the report. Mutually exclusive with the ``--report`` option.
+
+``--report=report_identifier``
 
-  Required. Contains the scan job identifier for the scan that is used to generate the report.
+  Required. Contains the report identifier to retrieve.  Mutually exclusive with the ``--scan-job`` option.
```
abaiken: Possibly remove `Required` from lines 478 & 482. In other places in the man.rst where we have mutually exclusive groups we don't put required (ie. the --json/--csv for report or viewing scan jobs with --name/--id)

docs/source/man.rst
```diff
@@ -495,11 +499,15 @@ The ``qpc report summary`` command generates a report that contains the processe
...
```
abaiken: Same comment as above.

docs/source/man.rst
kholdaway: Great suggestion!

docs/source/man.rst
kholdaway: Same

PR #926, opened 2018-03-02
UI: Scans & Download for Reports

cdcabrera:
PR...
* Setup reports service with dotenv config
* Reports Redux action, constants, reducer
* Reports unit and constant updates
* Download CSV Report
* API updates for scans service
* Copy update related to notification, started vs created scan
* Fix for icon component throwing error when empty string passed for "type"
* Fix for credential dialog and non-unique ids

Notes:
* the Reports unit test is a placeholder, that work should be in a follow up PR, shortly after this PR

updates #839


client/src/components/createCredentialDialog/createCredentialDialog.js
```diff
@@ -313,7 +314,7 @@ class CreateCredentialDialog extends React.Component {
           {this.renderFormLabel('Become Method')}
           <Grid.Col sm={7}>
             <div className="form-split-button">
-              <SplitButton className="form-control" bsStyle="default" title={becomeMethod} id="become-method-button">
+              <SplitButton className="form-control" bsStyle="default" title={becomeMethod} id={helpers.generateId('become-method-button')}>
```
jeff-phillips-18: Future - generateId should take a prefix string to reduce any chance of duplicates and to give the id some kind of context.

client/src/components/sources/sources.js
```diff
@@ -73,7 +73,7 @@ class Sources extends React.Component {
         alertType: 'success',
         message: (
           <span>
-            Started new scan <strong>{_.get(results, 'data.name')}</strong>.
+            Created new scan <strong>{_.get(results, 'data.name')}</strong>.
```
jeff-phillips-18: Good for now, but I think we need to start the scan upon successful creation.

cdcabrera: Agreed!

PR #928, opened 2018-03-02
Fix partial update of scan. Closes #901

### LINKED ISSUE #901: Partial update of scan is not correct (loses values not specified in request)
# Opening a new issue
___

## Specify type:
- Bug

- High

___

## Description:
Source and other partial updates in sonar assume that a missing field should not be updated.  Scan does not work this way.

## Acceptance Criteria: 
_(for enhancements)_

- [ ] Verify that CLI users can start a network scan
- [ ] Verify that CLI users can pause a network scan
- [ ] Verify that CLI users can restart a paused network scan

___

### Expected behavior:
Only values provided in request should be updated.

### Actual behavior:
Values not specified are accidentally cleared for disabled_optional_products and enabled_extended_product_search

### Steps to reproduce:
Create a scan with enabled_extended_product_search or disabled_optional_products set.  Run a partial update with options but not disabled_optional_products and enabled_extended_product_search.  Notice you lose these.

abaiken:
I am sure that there is a better way to reset the search_directories. Throw suggestions my way.


qpc/scan/utils.py
```diff
@@ -113,7 +113,8 @@ def get_enabled_products(enabled_ext_product_search,
     enabled_products = {}
     enabled_default = {scan.JBOSS_FUSE: False,
                        scan.JBOSS_EAP: False,
-                       scan.JBOSS_BRMS: False}
+                       scan.JBOSS_BRMS: False,
+                       'search_directories': ['/reset/the/dir/']}
```
kholdaway: Let's send the empty array.

PR #932, opened 2018-03-03
UI: Fixes in source view and wizard

### LINKED ISSUE #827: UI: Add Source Wizard issues

jeff-phillips-18:
Implement Delete source
Update sources list after Add, Edit, Delete
Do no lost updates in wizard when a new credential is added

Fixes #827

### DIFF
--- a/client/src/components/addSourceWizard/addSourceWizardStepTwo.js
+++ b/client/src/components/addSourceWizard/addSourceWizardStepTwo.js
@@ -76,7 +76,7 @@ class AddSourceWizardStepTwo extends React.Component {
     }
 
     let udpatedSource;
-    if (nextProps.source !== _.get(this.state, 'source')) {
+    if (!_.isEqual(nextProps.source, _.get(this.state, 'source'))) {
       udpatedSource = {
         source: _.get(nextProps, 'source', {}),
         sourceType: _.get(nextProps.source, apiTypes.API_SOURCE_TYPE, ''),
@@ -85,7 +85,7 @@ class AddSourceWizardStepTwo extends React.Component {
         hosts: _.get(nextProps.source, apiTypes.API_SOURCE_HOSTS, []),
         port: _.get(nextProps.source, apiTypes.API_SOURCE_PORT, ''),
         satelliteVersion: _.get(nextProps.source, apiTypes.API_SOURCE_SAT_VERSION, ''),
-        sslProtocol: _.get(nextProps.source, apiTypes.API_SOURCE_SSL_PORT, ''),
+        sslProtocol: _.get(nextProps.source, apiTypes.API_SOURCE_SSL_PROTOCOL, ''),
         sslCertVerify: _.get(nextProps.source, apiTypes.API_SOURCE_SSL_CERT, ''),
         disableSsl: _.get(nextProps.source, apiTypes.API_SOURCE_SSL_DISABLE, ''),
         singleHostPort: _.get(nextProps.source, apiTypes.API_SOURCE_HOSTS, []).length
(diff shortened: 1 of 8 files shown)


client/src/components/addSourceWizard/addSourceWizardStepTwo.js
```diff
@@ -75,19 +75,31 @@ class AddSourceWizardStepTwo extends React.Component {
       remappedCredentials = credentials.map(val => val.id);
     }
 
+    let udpatedSource;
+    if (nextProps.source !== _.get(this.state, 'source')) {
```
cdcabrera: There may be some limitations around this check, think using lodash's `_.isEqual(a, b)` might be better

client/src/constants/apiConstants.js
```diff
@@ -16,6 +16,6 @@ export {
   API_SOURCE_SAT_VERSION,
   API_SOURCE_SSL_CERT,
   API_SOURCE_SSL_DISABLE,
-  API_SOURCE_SSL_PROT,
+  API_SOURCE_SSL_PORT,
```
cdcabrera: yeahhhh, I should have named this one better, it was actually "prot" not "port"... it's an enum based on the API swagger doc...  `SSLv2, SSLv23, SSLv3, TLSv1, TLSv1_1, TLSv1_2`... maybe flip it to `PROTOCOL`

jeff-phillips-18: ah....

client/src/components/addSourceWizard/addSourceWizardStepTwo.js
jeff-phillips-18: I'll update to use _.isEqual for now. I think we need a deeper dive on the use of props and state for this component.

client/src/components/addSourceWizard/addSourceWizardStepTwo.js
```diff
@@ -125,19 +122,22 @@ class AddSourceWizardStepTwo extends React.Component {
...
-
+      let updatedSource = { id: source.id };
+      _.set(updatedSource, apiTypes.API_SOURCE_TYPE, sourceType);
+      _.set(updatedSource, apiTypes.API_SOURCE_NAME, sourceName);
+      _.set(updatedSource, apiTypes.API_SOURCE_HOSTS, hosts);
+      _.set(updatedSource, apiTypes.API_SOURCE_SAT_VERSION, satelliteVersion);
+      _.set(updatedSource, apiTypes.API_SOURCE_SSL_PROTOCOL, sslProtocol);
+      _.set(updatedSource, apiTypes.API_SOURCE_SSL_CERT, sslCertVerify);
+      _.set(updatedSource, apiTypes.API_SOURCE_SSL_DISABLE, disableSsl);
+      _.set(updatedSource, apiTypes.API_SOURCE_CREDENTIALS, credentials);
```
cdcabrera: This partially occurred to me before, viewing this update I've had a reoccurrence... I think we can definitely look at focusing the steps so they only update the properties they're responsible for, similar to using the former `_.merge` that was on `line 140`. If we handle it that way we might help alleviate some muddled confusion around 

> Why are we aiming to re-map the source object, both in resetting the state and in passing it back to Redux... what if we narrowed the scope even more, or at least until step 2 ends up handling the remaining properties. That ties into only having to observe each prior steps updated properties, similar to the "step valid" property.

I think if we combined this concept, with some form of a validation component that simply returns a value towards the state source object/model, and internally handles the validation, we could cut the wizard steps down even further.

jeff-phillips-18: Yeah, basically what this is doing. I can remove the fields that are not updated by this page and merge instead.

jeff-phillips-18:
Added the /, thanks!

PR #937, opened 2018-03-05
Add last checkin data to network and vcenter and provide in fingerprint. Closes #919.

### LINKED ISSUE #919: Capture last_check_in value within Fingerprints
## Specify type:
- Enhancement

## Description:
Satellite systems already contain a raw fact for last check in (date timestamp). We should similarly have these facts added to vCenter systems and Network systems raw facts.  vCenter systems should capture this value at the time of creating the System inspection result using python. Network systems should capture this value similar to the connection information using a Ansible task to set the fact.

The fingerprinting engine should be updated to capture these values.

## Acceptance Criteria: 
- [ ] Verify that a network scan include raw facts for last check in
- [ ] Verify that a vcenter scan includes raw facts for last check in
- [ ] Verify that the summary fingerprints include the last check in value

chambridge:


quipucords/fingerprinter/__init__.py
```diff
@@ -649,6 +649,18 @@ def _process_network_fact(source, fact):
                                 fingerprint,
                                 fact_value=system_creation_date.date())
 
+    if fact.get('connection_timestamp'):
```
kholdaway: This feels like it could be put into a helper function.  Perhaps there would be too many parameters.

PR #939, opened 2018-03-05
Improve QPC error response for 500 codes. Closes #898

### LINKED ISSUE #898: Make sure qpc prints correct error message for 500 errors.
See #897 for context.

@chambridge says the way to fix is to change `handle_token_errors` (request.py, line 40) to `handle_general_errors`.

abaiken:


abaiken:
@noahl The old mocked responses were wrong -- I think. They were returning successful json data and when it hit the 'handle_error_response_' (which it hadn't previously been hitting because we weren't catching the 500s explicitly) it was breaking because it expected either one error message, or multiple.

kholdaway:
Build failed.

PR #944, opened 2018-03-05
UI: Display proper error messages

jeff-phillips-18:
Updates #839


client/src/components/sources/createScanDialog.js
```diff
@@ -20,20 +25,39 @@ class CreateScanDialog extends React.Component {
...
+      Store.dispatch({
+        type: toastNotificationTypes.TOAST_ADD,
+        alertType: 'success',
+        message: (
+          <span>
+            Created new scan <strong>{this.state.scanName}</strong>.
+          </span>
+        )
+      });
+      this.props.onClose(true);
```
cdcabrera: Note: The callback setup works and the Redux works. Now that we've been handling Redux a bit more it might good if we come up with scenarios/guidelines of when to use a callback versus a Redux dispatch. 

One possibility for a guide is what's demonstrated here, something around updates and data happening across views or components that are shared, like toast notifications. This might also help us clean up a bit. Or maybe we just fully embrace Redux and everything is dispatched =)

PR #956, opened 2018-03-06
UI: Update error message determination methods.

jeff-phillips-18:
Updates #839

### DIFF
--- a/client/src/common/helpers.js
+++ b/client/src/common/helpers.js
@@ -1,3 +1,5 @@
+import _ from 'lodash';
+
 const bindMethods = (context, methods) => {
   methods.forEach(method => {
     context[method] = context[method].bind(context);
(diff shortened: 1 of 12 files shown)


client/src/common/helpers.js
```diff
@@ -149,6 +169,59 @@ const createViewQueryObject = (viewOptions, queryObj) => {
...
+    let errorContext = 'Error';
+    if (key !== 'non_field_errors' && key !== 'detail' && key !== 'options') {
+      errorContext = key;
+    }
+    if (responseData[key] instanceof String) {
+      if (message !== '') {
+        message += '\n';
+      }
+      message += `${errorContext}: ${responseData[key]}`;
+    } else if (responseData[key].forEach !== undefined) {
```
cdcabrera: Same comment about being explicit on the check as above...

PR #971, opened 2018-03-07
Add source information into summary report

### LINKED ISSUE #952: Provide source_type and sources information summary report
## Specify type:
- Enhancement

## Description:

For each source type tell whether they system was present.
(network, satellite, vcenter)

List of source names the system was found in

## Acceptance Criteria: 

- [ ] Verify that report users can determine what kinds of sources types the system was inspected with
- [ ] Verify that report users can see the list of sources the system was found in.

____
Associated with: https://github.com/quipucords/quipucords/issues/947

kholdaway:
- JSON includes sources field for each system.  
- CSV contains column for network/vcenter/satellite and source names.

Closes #952

### DIFF
--- a/quipucords/api/report/renderer.py
+++ b/quipucords/api/report/renderer.py
@@ -179,7 +179,8 @@ class ReportCSVRenderer(renderers.BaseRenderer):
         csv_content = report_buffer.getvalue()
         return csv_content
 
-    def _compute_source_info(self, sources):
+    @staticmethod
+    def _compute_source_info(sources):
         """Detect scan source types."""
         result = {
             NETWORK_DETECTION_KEY: False,
--- a/quipucords/api/report/renderer.py
+++ b/quipucords/api/report/renderer.py
@@ -110,6 +110,15 @@ class ReportCSVRenderer(renderers.BaseRenderer):
     """Class to render Deployment report as CSV."""
 
     # pylint: disable=too-few-public-methods
+    NETWORK_DETECTION_KEY = 'detection-network'
+    VCENTER_DETECTION_KEY = 'detection-vcenter'
+    SATELLITE_DETECTION_KEY = 'detection-satellite'
+    SOURCES_KEY = 'sources'
+    source_headers = {NETWORK_DETECTION_KEY,
+                      VCENTER_DETECTION_KEY,
+                      SATELLITE_DETECTION_KEY,
+                      SOURCES_KEY}
+
     media_type = 'text/csv'
     format = 'csv'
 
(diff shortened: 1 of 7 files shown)


quipucords/api/report/renderer.py
```diff
@@ -138,19 +147,51 @@ def render(self,
...
+            if system_sources is not None:
+                sources_info = self._compute_source_info(system_sources)
+            else:
+                sources_info = None
             for header in headers:
-                fact_value = system.get(header)
+                fact_value = None
+                if header in self.source_headers:
+                    if sources_info is not None:
+                        fact_value = sources_info.get(header)
```
chambridge: Is there a case here where fact value doesn't get set if source_info is None? Is None a good default or is empty string (just want to make sure we have the appropriate commas being generated)

quipucords/api/report/renderer.py
```diff
@@ -138,19 +147,51 @@ def render(self,
...
 
         csv_writer.writerow([])
 
         csv_content = report_buffer.getvalue()
         return csv_content
+
+    def _compute_source_info(self, sources):
+        """Detect scan source types."""
+        result = {
+            self.NETWORK_DETECTION_KEY: False,
```
chambridge: is `self.` necessary?

quipucords/fingerprinter/__init__.py
```diff
@@ -226,6 +251,7 @@ def _process_source(report_id, source):
 
         if fingerprint is not None:
             fingerprint['report_id'] = report_id
+            fingerprint[SOURCES_KEY] = [int(source_id)]
```
chambridge: A nit and fake, but could throw an error if source_id wasn't an int.

quipucords/api/report/renderer.py
kholdaway: None is good. Leaves column blank.

quipucords/api/report/renderer.py
chambridge: sure

quipucords/api/report/renderer.py
chambridge: ok

quipucords/api/report/renderer.py
kholdaway: Yeah, we probably have the same constants in many files.  It may be worth making a constants file at some time.

quipucords/fingerprinter/__init__.py
kholdaway: yep.

PR #980, opened 2018-03-07
UI: Add version and build info to about box

jeff-phillips-18:
Updates #839


client/src/services/statusService.js
```diff
@@ -0,0 +1,14 @@
+import axios from 'axios';
+
+class StatusService {
+  static getStatus() {
+    return axios({
+      method: 'get',
+      url: process.env.REACT_APP_STATUS_SERVICE,
+      xsrfCookieName: process.env.REACT_APP_AUTH_TOKEN,
+      xsrfHeaderName: process.env.REACT_APP_AUTH_HEADER
```
cdcabrera: So this is going sound weird, but GETs dont need the Cookie or Header with the way the server is setup, anything that modifies the data you do. And it defaults to GET, so you could trim it down it a bit
`axios({ url: ... })`

client/src/redux/constants/tests/index.test.js
```diff
@@ -1,19 +1,20 @@
...
-import * as factsTypes from './factsConstants';
-import * as reportsTypes from './reportsConstants';
-import * as scansTypes from './scansConstants';
-import * as sourcesTypes from './sourcesConstants';
-import * as toastNotificationTypes from './toasNotificationConstants';
-import * as userTypes from './userConstants';
-import * as viewTypes from './viewConstants';
-import * as viewPaginationTypes from './viewPaginationConstants';
-import * as viewToolbarTypes from './viewToolbarConstants';
+import reduxTypes from '..';
```
cdcabrera: Note: =) it's like eyes staring at me

jeff-phillips-18: Redux is watching you!

cdcabrera: heh

PR #981, opened 2018-03-07
Detect satellite version. Closes #767.

### LINKED ISSUE #767: Discover the satellite version in connection flow
## Specify type:
- Enhancement

## Description:
Determine the version of Satellite API (5 or 6.2+) by attempting login to each of the servers.

## Acceptance Criteria: 
- [ ] Verify that users can create a satellite source without specifying a version
- [ ] Verify that can scan a satellite source without a version and it will determine a connection API

chambridge:
@quipucords/ux-ui 

FYI, this removes the need to specify a satellite version.

### DIFF
--- a/quipucords/scanner/satellite/connect.py
+++ b/quipucords/scanner/satellite/connect.py
@@ -13,7 +13,7 @@ from requests import exceptions
 from api.models import (ScanTask)
 from scanner.task import ScanTaskRunner
 from scanner.satellite import utils
-from scanner.satellite.api import SatelliteException
+from scanner.satellite.api import SatelliteException, SatelliteAuthException
 from scanner.satellite.factory import create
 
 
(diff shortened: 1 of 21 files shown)


quipucords/scanner/satellite/connect.py
```diff
@@ -38,50 +38,45 @@ def __init__(self, scan_job, scan_task):
...
-            error_message += 'Connect scan failed for %s.' % self.scan_task
+            error_message = 'Satellite error encountered: %s. ' % conn_error
+            error_message += 'Connect scan failed for source %s.' \
+                % self.source.name
             return error_message, ScanTask.FAILED
         except TimeoutError as timeout_error:
-            error_message = 'Satellite error encountered: %s\n' % timeout_error
-            error_message += 'Connect scan failed for %s.' % self.scan_task
-            return error_message, ScanTask.FAILED
-        except Exception as unknown_error:  # pylint: disable=broad-except
```
kholdaway: I think the general exception could be useful now that we don't know for sure the version of Satellite we are talking too.  The job handle the exception this way:
```
message = 'FATAL ERROR. %s' % str(error)
self.scan_job.fail(message)
raise error
```

Would it be useful to log critical information and then raise again?

chambridge: I think that is a good point, but maybe we should open an issue that captures the job information in the Job's general exception capture.

PR #992, opened 2018-03-08
UI: use common createCredentialDialog, refreshTimeButton

jeff-phillips-18:
- Both CreateCredentialDialogs were being shown simultaneously
- Move forceUpdate to only redraw the refresh button not entire view

Updates #839

### DIFF
--- a/client/src/components/createCredentialDialog/createCredentialDialog.js
+++ b/client/src/components/createCredentialDialog/createCredentialDialog.js
@@ -323,7 +323,12 @@ class CreateCredentialDialog extends React.Component {
                 multiselect={false}
               >
                 {this.becomeMethods.map((nextMethod, index) => (
-                  <MenuItem key={index} eventKey={`become${index}`} onClick={() => this.setBecomeMethod(nextMethod)}>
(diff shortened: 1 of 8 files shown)


client/src/components/createCredentialDialog/createCredentialDialog.js
```diff
@@ -313,19 +315,19 @@ class CreateCredentialDialog extends React.Component {
...
+            <div className="quipucords-dropdownselect">
+              <DropdownSelect
                 title={becomeMethod}
-                id={helpers.generateId('become-method-button')}
+                id="become-method-select"
+                className="form-control"
+                multiselect={false}
               >
                 {this.becomeMethods.map((nextMethod, index) => (
                   <MenuItem key={index} eventKey={`become${index}`} onClick={() => this.setBecomeMethod(nextMethod)}>
```
cdcabrera: Note: added a class for highlighting the menu items, but you'd have to conditionally display them. Also, I kinda like it without the class: 
` className={{ 'quipucords-dropdownselect-menuitem-selected': ... }}`

PR #1007, opened 2018-03-10
UI: Undo change to  get task_results from connection_results

jeff-phillips-18:
Swagger structure is incorrect. Put back to prod/stage results format


jeff-phillips-18:
Filed issue #1016

PR #1018, opened 2018-03-12
Create API endpoint to get connection results. Closes #916

### LINKED ISSUE #916: API to get the connection results for a scan job
## Specify type:
- Enhancement

## Description:
Provide an API endpoint to get the connection results for a scan job. The results should be paginated.  System results should be ordered by status with `Failed` first.

### API
**Method:** GET
**URI:** `/api/v1/jobs/<job_id>/connection/`
**Response:**
400 -- bad id format
404 -- if not found
200 -- Body:
```
{
"count": <count>,
"next": <link>,
"previous": <link>,
"results:": [
    {
     "name": <system_name>,
     "status": <status>,
     "source": { "id": <source_id>, "name": <source_name>},
     "credential": { "id": <cred_id>, "name": <cred_name>}
    }
]
}
```

## Acceptance Criteria:
- [ ] Verify that API users can obtain paginated connection results.
- [ ] Verify that API users gets a 404 if no connection results exist or if the job does not exist.
- [ ] Verify that API users gets a 400 if no connection results ID or if the job ID is not an integer.

abaiken:


quipucords/api/scanjob/tests_scanjob.py
```diff
@@ -287,25 +287,127 @@ def test_details(self):
...
+                    'results': [{'name': 'Foo',
+                                 'status': 'success',
+                                 'credential': {'id': 1,
+                                                'name': 'cred1'},
+                                 'source': {'id': 1,
+                                            'name': 'source1',
+                                            'source_type': 'network'}}]}
+        self.assertEqual(json_response, expected)
+
+    def test_connection_failed_success(self):
```
kholdaway: This test currently has 2 results with a single task.  

I think we should have a test with two scan tasks:
 - test total ordering cross tasks
- test case where one task has no results the other has results

PR #1019, opened 2018-03-12
UI: Add source wizard fixes

cdcabrera:
What is it
* Fix for network hosts regex
* Initial automation for setting up a connection scan

What's not in the PR
* Fixes and filters for connection vs inspection scan within the Scans view.

updates #839

### DIFF
--- a/client/src/components/addSourceWizard/addSourceWizard.js
+++ b/client/src/components/addSourceWizard/addSourceWizard.js
@@ -7,6 +7,8 @@ import Store from '../../redux/store';
 import { confirmationModalTypes, sourcesTypes } from '../../redux/constants';
 import { addSourceWizardSteps, editSourceWizardSteps } from './addSourceWizardConstants';
 import { addSource, updateSource } from '../../redux/actions/sourcesActions';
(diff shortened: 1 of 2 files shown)


client/src/components/addSourceWizard/addSourceWizard.js
```diff
@@ -92,6 +94,33 @@ class AddSourceWizard extends React.Component {
...
+      const srcId = _.get(source, 'id');
+      const srcName = _.get(source, 'name');
+
+      if (typeof srcId === 'number') {
+        let data = {
+          name: `Connection Scan ${srcName}`,
+          sources: [srcId]
+        };
+
+        return addScan(data).then(response => startScan(_.get(response, 'value.data.id')));
```
jeff-phillips-18: Do we need to inform the user if either of these calls fail?

cdcabrera: We might, was looking at that... we could throw a similar error as below

cdcabrera: Added some additional catches for console warnings around the connection scans failing. For the most part though the app should probably continue unhindered.

PR #1020, opened 2018-03-12
UI Refactor: handle select/expand actions in common viewOptions reducer

jeff-phillips-18:


client/src/components/credentials/credentials.js
```diff
@@ -136,17 +136,17 @@ class Credentials extends React.Component {
   }
```
cdcabrera: Around line 87 looks like a left over `DESELECT_CREDENTIAL` in a dispatch. Fires when trying to delete multiple creds.

Issue #1017, opened 2018-03-12
UI front-end sprint: Bugs, design, API & testing refinement


cdcabrera:
## Specify type:
- Enhancement
- UI

## Description:
- [x] API #1036 #1044 #1047 #1107 
  - API calls & response structure updates
- [x] App optimization #1050 
- [x] Bugs/issues #1028 #1107 
- [x] Build updates #1031 #1035 #1071 
  - Django routing against React router `1031 1035`
- [x] Testing updates #1125 
  - Jest snapshots for components `1125`
- [x] Design & view refinement #1024 #1036 #1044 #1047 #1070 #1099 #1123 #1132 
  - About
  - Login
  - Sources `1024`
  - Scans `1024 1044 1070 1099`
  - Credentials

PR #1036, opened 2018-03-14
UI: Add Merge Scan Results Feature

jeff-phillips-18:
Updates #1017


client/src/components/mergeReportsDialog/mergeReportsDialog.js
```diff
@@ -0,0 +1,196 @@
...
+    super();
+
+    helpers.bindMethods(this, ['mergeScanResults']);
+  }
+
+  getValidScans() {
+    const { scans } = this.props;
+    return _.filter(scans, scan => {
+      return _.get(scan, 'most_recent.id') && _.get(scan, 'most_recent.status') === 'completed';
+    });
```
cdcabrera: Had trouble seeing when `most_recent.id` might not return when combined with the status `completed`?

client/src/components/mergeReportsDialog/mergeReportsDialog.js
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  getInvalidScans() {
+    const { scans } = this.props;
+    return _.filter(scans, scan => {
+      return !_.get(scan, 'most_recent.id') || _.get(scan, 'most_recent.status') !== 'completed';
```
cdcabrera: Same question about the combo for id and status.

client/src/components/mergeReportsDialog/mergeReportsDialog.js
```diff
@@ -0,0 +1,196 @@
...
+      </Modal>
+    );
+  }
+}
+
+MergeReportsDialog.propTypes = {
+  getMergeScanReults: PropTypes.func,
+  show: PropTypes.bool.isRequired,
+  scans: PropTypes.array,
+  onClose: PropTypes.func
```
cdcabrera: Note:
Callback for onClose, but there's also a prop for `show`. Seems to make sense at first, but feels off a little since we're using Redux. Might be a good candidate for looking at some clarity between using callbacks vs local state vs redux level

jeff-phillips-18: What I've been going on thus far is: if it is a common dialog we use redux for application state management but if it is a single dialog, pertinent only to a view, then we include it in the view and use the `show` prop to show/hide the dialog (and thus the close callback). I can change this to use redux if preferred.

client/src/components/mergeReportsDialog/mergeReportsDialog.js
jeff-phillips-18: true, if it completed it would have an id. I can simplify.

client/src/components/scans/scans.js
```diff
@@ -29,13 +36,21 @@ class Scans extends React.Component {
...
       'addSource',
       'refresh',
-      'notifyActionStatus'
+      'notifyActionStatus',
+      'hideMergeDialog'
     ]);
 
+    this.okToDelete = false;
+    this.scansToDelete = [];
+    this.deletingScan = null;
```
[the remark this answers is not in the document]
jeff-phillips-18: they are handled internally and don't effect the rendering so no real need to be state variables. Not opposed to using state if preferred, but just not necessary.

client/src/services/reportsService.js
```diff
@@ -65,16 +63,29 @@ class ReportsService {
...
-      }
-    );
+    return this.getReportDeployments(id, Object.assign(query, { format: 'csv' })).then(success => {
+      return this.downloadCSV(success.data, `report_${id}_deployments_${this.getTimeStampFromResults(success)}.csv`);
+    });
+  }
+
+  static getMergeScanResults(jobIds) {
+    let apiPath = process.env.REACT_APP_REPORTS_MERGED_JOBS;
+
```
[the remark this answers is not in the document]
jeff-phillips-18: Can do. Will need to call ReportsService's downloadCSV method from the ScansService or duplicate it.

client/src/components/mergeReportsDialog/mergeReportsDialog.js
cdcabrera: Let's iterate on this one... 
As for converting to Redux,  the explanation works "view based" vs "app level" 

If the component is maintained as is, thinking it can go the other direction, more stateless. Looks like it would work since it's embracing stateless aspects already. The only rethink might be around how the action is handled, but a callback would work too

client/src/components/mergeReportsDialog/mergeReportsDialog.js
```diff
@@ -0,0 +1,204 @@
...
+};
+
+const mapDispatchToProps = (dispatch, ownProps) => ({
+  mergeScans: data => dispatch(mergeScans(data)),
+  getMergedScanReportDetailsCsv: id => dispatch(getMergedScanReportDetailsCsv(id))
+});
+
+const mapStateToProps = function(state) {
+  return {};
+};
```
cdcabrera: we could probably just remove the `mapStateToProps`

client/src/components/scans/scans.js
```diff
@@ -179,6 +322,30 @@ class Scans extends React.Component {
     return null;
   }
 
+  renderScansActions() {
+    const { viewOptions } = this.props;
+
+    let mergeAllowed = _.size(viewOptions.selectedItems) > 1;
+
+    let deleteAction = null;
+    if (this.okToDelete) {
```
cdcabrera: this condition resolve? or a placeholder

client/src/redux/constants/reportsConstants.js
```diff
@@ -1,6 +1,3 @@
-const GET_REPORT_DEPLOYMENTS = 'GET_REPORT_DEPLOYMENTS';
-const GET_REPORT_DEPLOYMENTS_CSV = 'GET_REPORT_DEPLOYMENTS_CSV';
-const GET_REPORT_DETAILS = 'GET_REPORT_DETAILS';
-const GET_REPORT_DETAILS_CSV = 'GET_REPORT_DETAILS_CSV';
+const GET_REPORT = 'GET_REPORT';
 
-export { GET_REPORT_DEPLOYMENTS, GET_REPORT_DEPLOYMENTS_CSV, GET_REPORT_DETAILS, GET_REPORT_DETAILS_CSV };
+export { GET_REPORT };
```
cdcabrera: Nice! 

Only thing that might come back around, a second constant with the `_CSV` suffix, but definitely leave this as is  for now, since as you pointed out we don't really use a bunch of distinction with state for reports

client/src/components/mergeReportsDialog/mergeReportsDialog.js
jeff-phillips-18: I could pass noop maybe?

client/src/components/scans/scans.js
jeff-phillips-18: are you ok with just setting okToDelete as a constant or I can remove it. If i remove it I have to remove all the delete handling or comment it alllllll out.

cdcabrera: Ahhhh it's a placeholder. If we're going to leave it in place then at minimum I think a comment/todo that explains what's happening might be in order since it doesn't appear to resolve. Otherwise, yeahhhh a constant would work, but again a comment should probably be placed in there.

PR #1040, opened 2018-03-15
Add local pkg install. Add overrides and options to skip components. Added usage.

chambridge:


install/roles/docker/tasks/main.yml
```diff
@@ -52,7 +52,7 @@
 
 - name: Install Docker from local path
   shell: rpm -Uvh --force "{{find_docker_rpm_local_path}}"
-  ignore_errors: yes
+  # ignore_errors: yes
```
kholdaway: Is this still needed?

kholdaway: There were more of these commented out lines.  I just noted this one.

PR #1049, opened 2018-03-16
UI: Build, animation for scans, & clean up

cdcabrera:
What's included
* Build, add clean up to staging start up
* Scans service minor missing param
* Export as default clean up, update imports
* Confirmation modal state-to-props clean up
* Clean up, let vs const
* Aria label fix
* Animated spinner for scans view

updates #1017

### DIFF
--- a/client/src/components/confirmationModal/confirmationModal.js
+++ b/client/src/components/confirmationModal/confirmationModal.js
@@ -1,9 +1,7 @@
 import { connect } from 'react-redux';
 import React from 'react';
 import PropTypes from 'prop-types';
-
 import { Modal, Button, Icon } from 'patternfly-react';
-
 import helpers from '../../common/helpers';
 import Store from '../../redux/store';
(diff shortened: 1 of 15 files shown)


client/src/components/confirmationModal/confirmationModal.js
```diff
@@ -71,35 +60,26 @@ class ConfirmationModal extends React.Component {
...
-    confirmBody: state.confirmationModal.body,
-    confirmButtonText: state.confirmationModal.confirmButtonText,
-    cancelButtonText: state.confirmationModal.cancelButtonText,
-    onConfirm: state.confirmationModal.onConfirm,
-    onCancel: state.confirmationModal.onCancel
-  };
-}
+const mapStateToProps = function(state) {
+  return { ...state.confirmationModal };
+};
```
cdcabrera: Acknowledging the naming was probably done to avoid conflicts with html elements... since the overlap wasn't within this component went ahead and converted it

PR #1050, opened 2018-03-16
UI: Helper and constant updates

cdcabrera:
What's included

* Clean up and organize for helpers
* Removed unused service worker JS
* Casing adjustment for the Redux constant helper
  * similar to DEV_MODE
  * casing call out for being constant related

Note:
Separated the casing adjustment for the constant helper on purpose. We probably need to discuss a little before this gets merged in.


client/src/common/helpers.js
```diff
@@ -190,27 +192,19 @@ const getErrorMessageFromResults = results => {
...
 
-const rejectedAction = base => {
-  return `${base}_REJECTED`;
-};
+const DEV_MODE = process.env.REACT_APP_ENV === 'development';
 
-const pendingAction = base => {
-  return `${base}_PENDING`;
-};
+const FULFILLED_ACTION = base => `${base}_FULFILLED`;
```
jeff-phillips-18: Not thrilled about a function being all caps. Is there a particular reason for this change?

cdcabrera: Indeed
* Part of it had to do with what the helper is doing, substituting a constant string. Which isn't necessarily obvious at first function glance
* Another part was readability, in general, at a glance. The original constants it's replacing related clearly what the action was, just a giant bunch of all caps. Going into the reducers now, requires a bit more effort having to read instead of glance at string suffixes. The current lower camel-cased function prefixes just happen to be close in character length and are all cased in similar spots... `pendingAction`, `rejectedAction`, `fulfilledAction` vs `PENDING_ACTION`, `REJECTED_ACTION`, `FULFILLED_ACTION`.

The alternative had considered was the string/template literals. It adds a bit more overhead in the code, but it accomplished the readability bit while still keeping boilerplate constants out of the constants files... so they might look something like
```js
case `${SOME_CONSTANT}_FULFILLED`:
```
If it's more preferential to use the template literal instead of an uppercased function, shouldn't be a problem

At minimum do feel something needs to shift towards the readability whichever direction that happens to take, even something hadn't considered

jeff-phillips-18: OK, I'm fine with the change.

PR #1069, opened 2018-03-16
Update Jenkinsfile. Add documentation. Convert to 1.0.0 tag. Closes #786.

### LINKED ISSUE #786: Investigate reducing the number of commands required to install Sonar
# Opening a new issue
___

## Specify type:
- Enhancement

- Low
___

## Description:
Since we have an RPM installer for qpc would it be possible to minimize the number of commands needed to get Sonar setup? For example could it create some default dirs for logs, sshkeys etc? Also could qpc come configured with a server of 127.0.0.1 as the default so that command is not required? Steps like this that we can make optional if customers are using it in a non-default manner but the RPM install could take care of a lot of the defaults.

chambridge:


docs/source/install.rst
```diff
@@ -1,42 +1,145 @@
 Installing Quipucords
 =====================
-Quipucords is delivered in two parts, a command line tool as an RPM package and a server as a container image. The following instructions describe how to install the parts of Quipucords.
+Quipucords is delivered in two parts, a command line tool as an RPM package and a server as a container image. The following instructions describe how to install the parts of Quipucords using an scripted installation or step by step commands.
```
abaiken: parts of Quipucords using an scripted installation --> using a scripted installation

docs/source/install.rst
```diff
@@ -1,42 +1,145 @@
...
+
+Scripted Installation
+----------------------
+The scripted installation utilizes Ansible to install the command line and container image and any associated dependencies.
+
+Installing the Ansible Prerequisite
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Follow the `Ansible installation documentation <http://docs.ansible.com/ansible/latest/intro_installation.html#installing-the-control-machine>`_ to prepare the your system to install the software.
```
abaiken: to prepare the your system --> to prepare the system

docs/source/install.rst
```diff
@@ -1,42 +1,145 @@
...
+In order to install the command line you will need the following items, see `Installing the Quipucords Command Line Tool`_ for download locations for specific operating systems:
+
+- EPEL package
+- QPC package
+
+1. Change into the installer directory with the following command::
+
+    # cd install/
+
+2. Create a directory, default is ``packages``, to install downloaded packages from with the following command::
```
abaiken: default is ``packages`` --> the recommended default is ``packages`` --> This line isn't incorrect, it just reads weird to me so feel free to leave it as it is.

docs/source/install.rst
chambridge: done

docs/source/install.rst
chambridge: done

docs/source/install.rst
chambridge: I took this out since the `pkg_install_dir` option tries to explain how you could use a different directory with the installer.

chambridge:
Fixed the rendering on Github. Some wierdness with the section title handling.

PR #1081, opened 2018-03-19
Change product name to WildFly if version indicates WildFly

### LINKED ISSUE #1055: Don't mark EAP present when version is only Wildfly
Right now it's possible for Sonar to say EAP is present when the version is `Wildfly-10` (or presumably other Wildfly versions too). We should only say that EAP is present when we see an actual EAP version.

kholdaway:
Closes #1055

### DIFF
--- a/quipucords/fingerprinter/jboss_eap.py
+++ b/quipucords/fingerprinter/jboss_eap.py
@@ -19,6 +19,7 @@ logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
 NAME = 'name'
 PRODUCT = 'JBoss EAP'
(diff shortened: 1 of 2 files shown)


kholdaway:
I realized this solution is not great.  I am going to re-work some of it.

chambridge:
What would happen if they had JBoss EAP and Wildfly on the same system? Would we loose the EAP present?

Maybe we should extract out some code so that we can continue to have the JBoss EAP product show up (as absent in some cases) and also have Wildfly (essentially just adding to the list of products we are "fingerprinting")

kholdaway:
Right. That is why I marked it as not ready to merge. I realized it needed more.

PR #1099, opened 2018-03-20
UI: Use pagination for scan results, get failures from connection scan.

jeff-phillips-18:
Updates #1017


client/src/components/sources/sourceHostList.js
```diff
@@ -1,84 +1,188 @@
...
+      ordering: 'name'
+    };
+
+    if (_.get(source, 'connection.id')) {
+      this.props
+        .getConnectionScanResults(source.connection.id, queryObject)
+        .then(results => {
+          console.log('ConnectionScanResults');
+          console.dir(results);
+          console.log(_.get(results.value, 'data.next'));
```
cdcabrera: There's a few of these hanging around... see also `client/src/components/sources/sourceListItem.js`

jeff-phillips-18: Oops. Removed. Thanks.

client/src/components/scans/scanFailedHostList.js
```diff
@@ -0,0 +1,205 @@
...
+    allResults.sort((item1, item2) => {
+      if (this.isIpAddress(item1.name) && this.isIpAddress(item2.name)) {
+        const value1 = this.ipAddressValue(item1.name);
+        const value2 = this.ipAddressValue(item2.name);
+
+        return value1 - value2;
+      }
+
+      return item1.name.localeCompare(item2.name);
+    });
```
cdcabrera: Optional: Moving the IP related functionality to the helper/common area
Note: Getting weird results on the sort with the `localeCompare`, still looks alphabetized on mine... trying to figure it out, can't tell if it's my settings or the method. Running the sort off to the side independently works fine

screen shot 2018-03-21 at 3 00 00 pm

jeff-phillips-18: Are you looking in the Scans view at failed hosts? The sources view failed hosts do not sort because we are using server side paging/sorting. No sense sorting the local results when it could be just a subset.

jeff-phillips-18:
With the success hosts showing only results from the inspection scan and using paging, and the failed hosts showing both connection and inspection scan results and not paging, and local sorting, I think it would be too difficult to reuse. I could reuse just the components part. I'll take a crack at that.

client/src/components/scans/scanHostList.js
```diff
@@ -2,74 +2,273 @@ import _ from 'lodash';
...
+          scanResultsError: helpers.getErrorMessageFromResults(error),
+          disableNext: true,
+          disablePrevious: true
+        });
+      });
+  }
+
+  refresh(useProps, page) {
+    let { useConnectionResults, useInspectionResults, status } = useProps;
+    page = page || 1;
```
cdcabrera: Iteration: could set a default param/argument instead... `refresh(useProps, page = 1)`

PR #1115, opened 2018-03-22
Order and filter connection and inspection results. Closes #1101.

### LINKED ISSUE #1101: Ability to sort/filter connection and inspection job results
## Type:
- Enhancement

### Priority:
- High
___

## Description:
The results of the call to get a scan job's connection and inspection results are paged, therefore sorting on the client is not accurate. Requesting we sort/filter the results in the API:
```
/jobs/2/connection/?page=1&page_size=100&ordering=name&status=failed
```
would sort by the system 'name' field and filter by status (failed) results.

Are there other statuses returned other than 'success' and 'failed'? The UI only distinguishes the two so if there are others it could be an issue we need to address (filter by failed returns all statuses that would fall into the failure category).
___
@

chambridge:
FYI
@quipucords/ux-ui

### DIFF
--- a/quipucords/api/scanjob/tests_scanjob.py
+++ b/quipucords/api/scanjob/tests_scanjob.py
@@ -279,6 +279,80 @@ class ScanJobTest(TestCase):
                                             'source_type': 'network'}}]}
         self.assertEqual(json_response, expected)
 
+    def test_connection_bad_ordering_filter(self):
+        """Test ScanJob connection results with bad ordering filter."""
+        # pylint: disable=no-member
+        scan_job, scan_task = create_scan_job(
+            self.source, ScanTask.SCAN_TYPE_INSPECT)
+
+        # Create a connection system result
+        sys_result = SystemConnectionResult(name='Foo',
+                                            source=self.source,
+                                            credential=self.cred,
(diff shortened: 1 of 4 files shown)


quipucords/api/scanjob/tests_scanjob.py
```diff
@@ -328,6 +402,56 @@ def test_connection_failed_success(self):
...
+        url = reverse('scanjob-detail', args=(scan_job.id,)) + 'connection/'
+        url += '?ordering=-name'
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        json_response = response.json()
+        expected = {'count': 2,
+                    'next': None,
+                    'previous': None,
+                    'results': [
+                        {'name': 'Foo',
```
kdelee: So this is supposed to be ordering these items on the name value? Why would `Foo` come before `Bar`?

kdelee: I guess you are saying that descending is `Z X Y ... A` instead of `A B C D....`?

chambridge: correct `-name` means sort the strings in reverse alphabetical order.

PR #1122, opened 2018-03-22
Detect BRMS version in non-invasive scan

### LINKED ISSUE #1014: Improve non-invasive BRMS inspection to determine version
## Specify type:
- Enhancement

### Priority:
- Medium
___

## Description:
Currently, we don't determine the version of the BRMS product using the non-invasive scan. We need to improve our data collection to gather the necessary information to determine the version of BRMS.
___

## Acceptance Criteria:

- [ ] Verify that we can determine the version of BRMS a non-invasive scan

noahl:
We were already collecting the necessary data. This change plumbs it
through to the fingerprinter and update the version classification
logic.

Closes #1014

### DIFF
--- a/quipucords/fingerprinter/jboss_brms.py
+++ b/quipucords/fingerprinter/jboss_brms.py
@@ -11,9 +11,8 @@
 
 """Ingests raw facts to determine the status of JBoss BRMS for a system."""
 
+import itertools
 import logging
-import os
-import re
 from api.models import Product, Source
 from fingerprinter.utils import (product_entitlement_found,
                                  generate_raw_fact_members)
@@ -74,27 +73,6 @@ def classify_version_string(version_string):
     return None
 
 
-def classify_kie_file(pathname):
-    """Classify a kie-api-* file.
-
-    :param pathname: the path to the file
-    :returns: a BRMS version string, or None if not a Red Hat kie file.
-    """
-    # os.path.basename behaves differently if the last part of the
-    # path ends in a /, so normalize.
-    if pathname.endswith('/'):
-        pathname = pathname[:-1]
-
-    basename = os.path.basename(pathname)
-
-    # Filename format is 'kie-api-' + version + '.jar' + stuff.
-    match = re.match(r'kie-api-(.*)\.jar.*', basename)
-    if match:
-        return classify_version_string(match.group(1))
-
-    return None
-
-
 # pylint: disable=R0912,R0914,R0915
 def detect_jboss_brms(source, facts):
     """Detect if JBoss BRMS is present based on system facts.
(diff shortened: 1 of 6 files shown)


quipucords/fingerprinter/jboss_brms.py
```diff
@@ -138,25 +127,42 @@ def detect_jboss_brms(source, facts):
     }
     product_dict = {'name': PRODUCT}
 
+    versions = set()
```
chambridge: Do we have issues if they have BRMS installed at the same version multiple times on the same system?

quipucords/fingerprinter/jboss_brms.py
```diff
@@ -138,25 +127,42 @@ def detect_jboss_brms(source, facts):
...
+    found_kie_version = False
+    for filename in kie_files:
+        category = classify_kie_file(filename)
```
chambridge: Are we able to tweak this raw fact to capture home directory so we can differentiate between multiple installs?

noahl:
@chambridge Are you saying that you want to be able to provide a list of the different installs on a system and what versions they are? I can do that, just want to be sure it's what you're thinking of. You're right, I made this version not distinguish between different installations on the same host.

quipucords/fingerprinter/jboss_brms.py
noahl: Can do.

quipucords/fingerprinter/jboss_brms.py
noahl: I designed this to deduplicate versions on a single system, and it sounds like you want to keep the version information for each installation. Is that right? I will go ahead and work on that change.

noahl:
@chambridge just pushed a major update; please take a look. It now includes the installation directory with the BRMS installation. The output looks like this:
```
{
  "name": "JBoss BRMS",
  "presence": "present",
  "version": [
    "/opt/jboss/jboss-eap-6.4/standalone/deployments/business-central.war:BRMS 6.4.0",
    "/root/jboss-eap-6.4/standalone/deployments/kie-server.war:BRMS 6.4.0",
    "/root/jboss-eap-6.4/standalone/deployments/business-central.war:BRMS 6.4.0",
    "None:BRMS 6.4.0",
    "/opt/jboss/jboss-eap-6.4/standalone/deployments/kie-server.war:BRMS 6.4.0"
  ],
  "metadata": {
    "source_id": 5,
    "source_name": "sonar-jbossbrms-rhel-6-vc",
    "source_type": "network",
    "raw_fact_key": "jboss_brms_kie_in_business_central/jboss_brms_locate_kie_api/jboss_brms_manifest_mf"
  }
}
```

The one thing I haven't figured out is how I want to handle the fact that we get business-central.war and kie-server.war directories for each BRMS installation. I was thinking of using /root/jboss-eap-6.4 as the identifier instead, but that is less specific about where we found BRMS. What do you think? Other than that, I think this works.

noahl:
@chambridge okay, I think I see what you're asking. Just pushed a new version with this output:
```
{
  "name": "JBoss BRMS",
  "presence": "present",
  "version": [
    "BRMS 6.4.0"
  ],
  "metadata": {
    "source_id": 5,
    "source_name": "sonar-jbossbrms-rhel-6-vc",
    "source_type": "network",
    "raw_fact_key": "jboss_brms_kie_in_business_central/jboss_brms_locate_kie_api/jboss_brms_manifest_mf"
  }
}
```

but the raw facts still have the directories. (Except for one particular raw fact that can't gather them, but that's a special case and there's a comment about why. It should only affect compressed war files, and I'm not even sure Red Hat distributes those.)

noahl:
@chambridge not yet, will test

noahl:
@chambridge I had to push a fix based on the results with extended search, so please take another look before I merge. Here are the full test results:

Results on an EAP system with no BRMS, extended search off:
```
{
  "name": "JBoss BRMS",
  "presence": "absent",
  "metadata": {
    "source_id": 2,
    "source_name": "sonar-jbosseap-rhel-7-vc",
    "source_type": "network",
    "raw_fact_key": null
  }
}
```

On a fuse machine, no extended search:
```
{
  "name": "JBoss BRMS",
  "presence": "absent",
  "metadata": {
    "source_id": 10,
    "source_name": "sonar-jbosseap-fuse-rhel-6-vc",
    "source_type": "network",
    "raw_fact_key": null
  }
}
```

On a BRMS system with extended search turned on:
```
{
  "name": "JBoss BRMS",
  "presence": "present",
  "version": [
    "BRMS 6.4.0"
  ],
  "metadata": {
    "source_id": 5,
    "source_name": "sonar-jbossbrms-rhel-6-vc",
    "source_type": "network",
    "raw_fact_key": "jboss_brms_drools_core_ver/jboss_brms_kie_api_ver/jboss_brms_kie_in_business_central/jboss_brms_kie_war_ver/jboss_brms_locate_kie_api/jboss_brms_manifest_mf"
  }
}
```

Fuse system, extended search on:
```
{
  "name": "JBoss BRMS",
  "presence": "absent",
  "metadata": {
    "source_id": 10,
    "source_name": "sonar-jbosseap-fuse-rhel-6-vc",
    "source_type": "network",
    "raw_fact_key": null
  }
}
```

chambridge:
Changes look good.

Issue #1135, opened 2018-03-27
UI front-end sprint: Bugs, design, testing, & packaging


cdcabrera:
## Specify type:
- Enhancement
- UI

## Description
- [x] API calls & response structure updates #1177 
- [x] Bugs/issues #1164 #1165 #1184 #1186 
- [x] Build/packaging updates #1180 #1186 
- [x] Copy, Design & View updates, #1144 #1160 #1177 #1194 
  - Infinite scroll, `1144`
  - Copy `1194`
  - Credentials, `1160`
  - Unreachable hosts, `1177`
- [x] Testing updates, #1150 #1172

PR #1144, opened 2018-03-28
UI: Use infinite scroll for host lists.

jeff-phillips-18:
Updates #1135

### DIFF
--- a/client/src/components/scanHostList/scanHostList.js
+++ b/client/src/components/scanHostList/scanHostList.js
@@ -1,29 +1,29 @@
 import _ from 'lodash';
 import React from 'react';
-import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
+import InfiniteScroll from 'react-infinite-scroller';
 
 import helpers from '../../common/helpers';
 import { connect } from 'react-redux';
-import { Button, EmptyState, Grid } from 'patternfly-react';
+import { EmptyState, Grid } from 'patternfly-react';
 import { getConnectionScanResults, getInspectionScanResults } from '../../redux/actions/scansActions';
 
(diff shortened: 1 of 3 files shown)


client/src/components/scanHostList/scanHostList.js
```diff
@@ -102,15 +80,14 @@ class ScanHostList extends React.Component {
...
 
         this.setState({
           inspectionScanResultsPending: morePages,
           scanResults: this.addResults(results),
-          prevResults: morePages || this.state.connectionScanResultsPending ? this.state.prevResults : [],
-          disableNext: !usePaging || _.get(results.value, 'data.next') === null,
-          disablePrevious: !usePaging || _.get(results.value, 'data.previous') === null
+          prevResults: morePages || this.state.connectionScanResultsPending ? this.state.prevResults : []
         });
+        this.loading = this.state.connectionScanResultsPending;
```
cdcabrera: Note: we probably need to start looking at being consistent, for tracking, and continue declaring class level properties in the constructor. they tend to get lost a little

client/src/components/scanHostList/scanHostList.js
```diff
@@ -190,62 +163,64 @@ class ScanHostList extends React.Component {
...
-              <Grid.Row>
-                <Button bsStyle="link" className="show-next-button" onClick={this.onNextPage}>
-                  Show next {nextSetSize} results
-                </Button>
-              </Grid.Row>
-            </Grid>
-          )}
+        <Grid fluid className="host-list">
+          <InfiniteScroll
+            key="infinite-scroller"
```
cdcabrera: Iterate: it was looking like `key` could be removed

PR #1164, opened 2018-04-02
UI: Allow download of reports for failed scans with report IDs.

### LINKED ISSUE #1162: UI doesn't show report if 1 source had an error
## Specify type:
- Bug

### Priority:
- Medium

___

## Description:
UI doesn't present "Download" box for report if a scan had a source/task that failed even though report ID is generated and CLI allows for successful report to be downloaded. users may need to download the report to see exactly what failed or wasn't generated. 

___

## Acceptance Criteria:

- [ ] Verify that users can download a report if report ID is generated regardless if the scan failed or not.

___

### Version Information:
CLI Server Version or commit information.

### Expected behavior:
Describe the expected behavior for the program.

### Actual behavior:
Provide the actual behavior for the program.

### Steps to reproduce:
List clear steps to reproduce the bug.

### Environment information:
Provide detailed information on the *scanning* and *target* systems (if applicable); see example below.

Scanning System | RHEL             | 7.4        | 3.6            | Bare Metal
Target System   | RHEL             | 5.9        | 2.4.6          | RHEV

jeff-phillips-18:
Fixes #1162

### DIFF
--- a/client/src/components/scans/scanJobsList.js
+++ b/client/src/components/scans/scanJobsList.js
@@ -106,7 +106,7 @@ class ScanJobsList extends React.Component {
           {job.systems_scanned > 0 ? job.systems_scanned : '0'}
         </Grid.Col>
         <Grid.Col xs={3} sm={2} smPush={5}>
-          {job.status === 'completed' && (
+          {job.report_id > 0 && (
             <Dropdown id={helpers.generateId()} className="pull-right" pullRight>
               <Dropdown.Toggle useAnchor>
                 <Icon type="fa" name="download" />
--- a/client/src/components/scans/scanListItem.js
+++ b/client/src/components/scans/scanListItem.js
@@ -205,6 +205,20 @@ class ScanListItem extends React.Component {
   renderActions() {
     const { item, onSummaryDownload, onDetailedDownload, onPause, onCancel, onStart, onResume } = this.props;
 
+    let downloadActions = null;
+    if (_.get(item, 'most_recent.report_id')) {
+      downloadActions = (
+        <DropdownButton key="downLoadButton" title="Download" pullRight id={`downloadButton_${item.id}`}>
+          <MenuItem eventKey="1" onClick={() => onSummaryDownload(_.get(item, 'most_recent.report_id'))}>
+            Summary Report
+          </MenuItem>
+          <MenuItem eventKey="2" onClick={() => onDetailedDownload(_.get(item, 'most_recent.report_id'))}>
+            Detailed Report
+          </MenuItem>
+        </DropdownButton>
+      );
+    }
+
     switch (_.get(item, 'most_recent.status')) {
       case 'completed':
         return (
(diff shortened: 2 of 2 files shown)


client/src/components/scans/scanJobsList.js
```diff
@@ -106,7 +106,7 @@ class ScanJobsList extends React.Component {
           {job.systems_scanned > 0 ? job.systems_scanned : '0'}
         </Grid.Col>
         <Grid.Col xs={3} sm={2} smPush={5}>
-          {job.status === 'completed' && (
+          {job.report_id > 0 && (
```
cdcabrera: It probably shouldn't happen a zero ID, buttttt can totally see it happening... maybe just a `>=`. 

This check along with the one on `line 209, client/src/components/scans/scanListItem.js` seem like they should be doing the same set of checks if we're just detecting the property exists, or that it has a value to "report" off of.

client/src/components/scans/scanListItem.js
```diff
@@ -205,6 +205,20 @@ class ScanListItem extends React.Component {
...
+        <DropdownButton key="downLoadButton" title="Download" pullRight id={`downloadButton_${item.id}`}>
+          <MenuItem eventKey="1" onClick={() => onSummaryDownload(_.get(item, 'most_recent.report_id'))}>
+            Summary Report
+          </MenuItem>
+          <MenuItem eventKey="2" onClick={() => onDetailedDownload(_.get(item, 'most_recent.report_id'))}>
+            Detailed Report
+          </MenuItem>
+        </DropdownButton>
+      );
+    }
```
cdcabrera: Iterate: Like centralizing it works well. Doing this points out we've got some pretty heavy HTML repetition going on, something to take a look at as clean up maybe

client/src/components/scans/scanListItem.js
```diff
@@ -205,6 +205,20 @@ class ScanListItem extends React.Component {
   renderActions() {
     const { item, onSummaryDownload, onDetailedDownload, onPause, onCancel, onStart, onResume } = this.props;
 
+    let downloadActions = null;
+    if (_.get(item, 'most_recent.report_id')) {
```
cdcabrera: This check and the one from `line 109 scanJobsList.js` seem like they should be the same

jeff-phillips-18: The problem is in the JSX if the value is not a boolean then the number would be displayed. 1 and true are treated differently.

client/src/components/scans/scanJobsList.js
jeff-phillips-18: 0 is not a valid id, but I can change to >= 0 or a comparison against undefined if you prefer.

cdcabrera: Expanding out, looks like it's in a couple of spots, can avoid it

client/src/components/scans/scanListItem.js
cdcabrera: Can see the truthy check, and it being passed on. Unsure follow the rest, we may be talking from different points. Regardless, looks like the same `_.get` is called a few times, something to iterate on, along with with the HTML... looking good

cdcabrera:
tracking #1135

PR #1170, opened 2018-04-03
Catch additional invalid host range format. Closes #1161.

### LINKED ISSUE #1161: Network Source Accepts Invalid Hosts Range
## Specify type:
- Enhancement

### Priority:
- Low
___

## Description:
When creating a network source you can input invalid host ranges like 192.168.56.1-254 without an error until you run the scan.

e.g. qpc source add --type network --name local_vms --hosts 192.168.56.0-254 --cred local_root
___

## Acceptance Criteria:

- [ ] Verify that users receive an error when they input an invalid hosts range/value

___

chambridge:


quipucords/api/source/serializer.py
```diff
@@ -353,6 +353,7 @@ def validate_hosts(hosts):
         relaxed_cidr_pattern = r'[0-9]*\.[0-9]*\.[0-9]*\.[0-9]*\/[0-9]*'
         relaxed_invalid_ip_range = r'[0-9]*\.[0-9]*\.[0-9]*\.[0-9]*-' \
                                    r'[0-9]*\.[0-9]*\.[0-9]*\.[0-9]*'
+        relaxed_invalid_ip_range_alt = r'[0-9]*\.[0-9]*\.[0-9]*\.[0-9]*-[0-9]*'
```
kholdaway: Maybe we should check for a - in the other 3 groups.  For example: 1-2.56.6.3 would also be invalid.

chambridge: done

PR #1175, opened 2018-04-04
Deduplicate EAP versions

### LINKED ISSUE #1174: Deduplicate JBoss EAP versions
This issue is part of #1102 .

Right now the EAP fingerprinter can report multiple versions that are logically the same, like for instance `'7.0.8'` and `'EAP-7.0'`. We want to only report genuinely unique versions.

noahl:
Closes #1174

### DIFF
--- a/quipucords/fingerprinter/jboss_eap.py
+++ b/quipucords/fingerprinter/jboss_eap.py
@@ -11,6 +11,7 @@
 
 """Ingests raw facts to determine the status of JBoss EAP for a system."""
 
+import bisect
 import logging
 
 from api.models import Product, Source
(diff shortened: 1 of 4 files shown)


chambridge:
Looks like linting failed. You need to rerun `pip install` to pull the import-order flake8 plugin.

noahl:
@chambridge thanks! Just uploaded a new version with the correct import groups.

chambridge:
I see how this solves the version from classification & version from manifest into a single version issue, but did we do anything to handle knowing whether there are multiple installs similar to the work you did with BRMS?

Such that it would be possible for a customer to install EAP on the system twice and we would show a resulting ["6.x.y", "6.x.x"] in the versions section to declare the two separate installations?

noahl:
@chambridge it should handle that situation in the same way the BRMS fingerprints do, where you get two version numbers if your two installs are different versions and one version number if they are the same version.

The one caveat is that sometimes we only get two digits of version from an installation, because our version table doesn't have all three. I've asked Mark about it, but I don't think we know how that table was generated, and I couldn't find the originals to make a more detailed one. So it could happen that you had 6.x.y and 6.x.z, but we only identify 6.x.y and 6.x, and then we deduplicate them to just 6.x.y.

That seemed like a pretty unlikely situation to me, which is why I left it like that, but I can fix it if you want to. I'd rather do that in a separate PR though, to keep the size manageable.

chambridge:
Ok. Just couldn't clearly tell without an example detail report.

PR #1184, opened 2018-04-06
UI: Show source associated with system in scans results lists.

### LINKED ISSUE #1030: Multi-Source Scan host list
## Specify type:
- Bug

### Priority:
- High
___

## Description:
Scanning multiple networks returns a combined list of hosts for an individual network.

___

## Bug Report
Create 2 network sources
- Net#1: 10.10.181.[98:178]
- Net#2:  10.10.181.[0:98]

Select them both and click "Scan"

The interface should return the correct number of hosts (6) for each source, but the list of hosts is combined. The 10.10.181.98 address is also listed twice.

Going to the "Scans" page and expanding the list of hosts also fails. See second attachment.

### Version Information:
#1028 

image

image

jeff-phillips-18:
Fixes #1030

image

@bclarhk

### DIFF
--- a/client/src/components/scans/scanListItem.js
+++ b/client/src/components/scans/scanListItem.js
@@ -280,12 +280,17 @@ class ScanListItem extends React.Component {
 
   renderHostRow(host) {
     return (
-      <Grid.Col xs={12}>
-        <span>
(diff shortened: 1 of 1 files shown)


cdcabrera:
tracking #1135

PR #1214, opened 2018-04-13
Allow true/false for boolean values in source options

### LINKED ISSUE #1207: Make qpc boolean options' value to ignore case
## Specify type:
- Enhancement

### Priority:
- Medium
___

## Description:
There are qpc commands that accepts either `True` or `False` (boolean value). It would be better, in terms of usability, if the value could be all lower case or even better to ignore the case. This specially because in the output the value are presented all lower case (JSON notation).
___

## Acceptance Criteria:

- [ ] Verify that qpc boolean options can accept lower case values. For example the source `--ssl-cert-verify` option.
___

### Version Information:
```console
$ qpc --version
0.0.41
```

Commit 0656187badf382a59ecc975fc7c3e7d6665cb643

### Expected behavior:
Boolean qpc options should accept true or false in and character case combination

### Actual behavior:
Boolean qpc options values are required to have the first character uppercase and all the following lower case.

### Steps to reproduce:
For example try to set the --ssl-verify-option or a source, it will force the case sensitive values.

### Environment information:
N/A

kholdaway:
Closes #1207

### DIFF
--- a/qpc.spec
+++ b/qpc.spec
@@ -56,7 +56,7 @@ install -D -p -m 644 build/qpc.1 $RPM_BUILD_ROOT%{_mandir}/man1/qpc.1
 %{_mandir}/man1/qpc.1.gz
 
 %changelog
-* Fri Apr 13 2018 Kevan Holdaway <kholdawaß@redhat.com> 0.0.42-1
+* Fri Apr 13 2018 Kevan Holdaway <kholdawa@redhat.com> 0.0.42-1
 - Allow true/false for boolean values in source options
 * Wed Mar 14 2018 Ashley Aiken <aaiken@redhat.com> 0.0.41-1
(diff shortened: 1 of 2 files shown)


kholdaway:
spec updated

qpc.spec
```diff
@@ -56,6 +56,8 @@ install -D -p -m 644 build/qpc.1 $RPM_BUILD_ROOT%{_mandir}/man1/qpc.1
 %{_mandir}/man1/qpc.1.gz
 
 %changelog
+* Fri Apr 13 2018 Kevan Holdaway <kholdawaß@redhat.com> 0.0.42-1
```
chambridge: not your email.

chambridge:
🥉

PR #1219, opened 2018-04-13
Fix false identification of Wildfly as EAP

### LINKED ISSUE #1217: Detect Wildfly 11
## Specify type:
- Bug

### Priority:
- High
___

## Description:
When scanning our system `10.10.181.205` we fail to detect Wildfly version 11

Report shows:
```
{
    "name": "JBoss EAP",
    "version": null,
    "presence": "present",
    "metadata": {
        "source_id": 15,
        "source_name": "WildFly",
        "source_type": "network",
        "raw_fact_key": "jboss_eap_running_paths/jboss_eap_locate_jboss_modules_jar/eap_home_ls/eap_home_version_txt"
    }
}
```

This causes us to report EAP because the version is not known.  In this case, we know the test machine has Wildfly 11 so we can improve this.

noahl:
There were several separate issues that lead to this:

  - We considered jboss_eap_running_paths to be conclusive evidence of
    EAP presence. This commit changes that to be potential, because
    that fact can also find Wildfly.
  - We considered jboss-modules.jar to be conclusive evidence of EAP,
    when it is present for Wildfly too.
  - We used several other indicator files as evidence for EAP, even
    though they are present for Wildfly too.
  - We had a typing error that lead to us treating version.txt as
    present (which is evidence for EAP) when it was in fact absent.

We don't detect the version of Wildfly with this PR because we specifically added code to this fingerprinter to disable detecting Wildfly versions. It should be possible to detect the Wildfly version if we remove that, but that's a separate discussion.

Closes #1217 .

### DIFF
--- a/quipucords/fingerprinter/jboss_eap.py
+++ b/quipucords/fingerprinter/jboss_eap.py
@@ -80,6 +80,7 @@ EAP_CLASSIFICATIONS = {
     '1.5.0.Final': 'WildFly-10',
     '1.5.1.Final': 'WildFly-10',
     '1.5.2.Final': 'WildFly-10',
+    '1.6.0.Final': 'WildFly-11',
     '1.6.1.Final': 'WildFly-11',
     '1.7.0.Final': 'WildFly-12',
     'JBPAPP_4_2_0_GA': '4.2',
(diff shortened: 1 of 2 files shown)


noahl:
Just added a new commit to detect this version of WildFly. This test VM has a point release of jboss-modules that we've never seen before, so we need to add it to our list.

chambridge:
What is the final result against the specified test machine for the JBoss EAP product?

kholdaway:
```
"products": [
    {
        "name": "JBoss EAP",
        "version": [],
        "presence": "absent",
        "metadata": {
            "source_id": 15,
            "source_name": "WildFly",
            "source_type": "network",
            "raw_fact_key": "jboss_eap_running_paths/jboss_eap_locate_jboss_modules_jar"
        }
    }
```

PR #1222, opened 2018-04-13
Add validation to reject updates that add SSL options to network source

kholdaway:


kholdaway:
@abaiken This was a change in the API not CLI.  No need to update spec file.

PR #1242, opened 2018-04-17
Add logging to determine if scan manager is alive

### LINKED ISSUE #1236: RFE: Heartbeat logging for Scan Manager
## Specify type:
- Enhancement

### Priority:
- High
___

## Description:
We are running into an issue with the automated jobs where it appears that the server can sometimes enter into a broken state where scanjobs get stuck in the `created` state and never proceed. It has been postulated that the "Scan Manager" has died in these cases. 

It would be very helpful to understand the state of the server if the Scan Manger logged its status at some regular interval.
___

## Acceptance Criteria:

- [ ] Verify that additional logging is added that makes it clear whether the Scan Manger is alive

abaiken:
Closes #1236 
I know that this is a silly thing to not know but what file are these logs being saved too (I have always relied on reading the server terminal print outs) and when I looked in the settings.py file it seemed like the default was app.log (but mine is empty)? 

Anyways, as an example, this is what the output looks like. The annoying thing (or maybe it's helpful) is that this will continue to log if there are no jobs as long as the scan manager is still alive:

ie. While running scan jobs:
```
INFO Task 18 (inspect, network, NSmallSource, elapsed_time: 5s) - PROCESSING 10.10.182.81 - ANSIBLE ROLE virt
INFO Scan Manager Stats: current_scan_job=3. queue_length=3.
INFO Task 18 (inspect, network, NSmallSource, elapsed_time: 8s) - PROCESSING 10.10.181.132 - ANSIBLE ROLE cpu
```

After finishing (with two minutes gone by):
```
INFO Job 2 (inspect, elapsed_time: 120s) - STATE UPDATE (completed)
INFO Scan Manager Stats: current_scan_job=None. queue_length=0.
INFO Scan Manager Stats: current_scan_job=None. queue_length=0.
```

### DIFF
--- a/quipucords/scanner/manager.py
+++ b/quipucords/scanner/manager.py
@@ -12,6 +12,7 @@
 """Queue Manager module."""
 
 import logging
+import os
 from threading import Thread, Timer
 from time import sleep
 
@@ -41,11 +42,14 @@ class Manager(Thread):
 
     def log_info(self):
         """Log the status of the scan manager."""
-        interval = 60
+        try:
+            interval = int(os.environ.get('QUIPUCORDS_MANAGER_HEARTBEAT', '60'))
+        except ValueError:
+            interval = 60
         heartbeat = Timer(interval, self.log_info)
-        logger.debug('Scan manager running: %s. '
-                     'Current task: %s. '
-                     'Number of scans in queue: %s.',
+        logger.info('Scan manager running: %s. '
+                    'Current task: %s. '
+                    'Number of scan jobs in queue: %s.',
                      self.running, self.current_task, len(self.scan_queue))
         if self.running:
             heartbeat.start()
--- a/quipucords/scanner/manager.py
+++ b/quipucords/scanner/manager.py
@@ -48,15 +48,15 @@ class Manager(Thread):
         except ValueError:
             interval = 60
         heartbeat = Timer(interval, self.log_info)
-        logger.info('Scan manager running: %s. '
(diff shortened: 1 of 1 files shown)


quipucords/scanner/manager.py
```diff
@@ -39,6 +39,20 @@ def __init__(self):
         self.running = True
         logger.debug('Scan manager created.')
 
+    def log_info(self):
+        """Log the status of the scan manager."""
+        interval = 60
```
chambridge: Could we make this configurable via an environment variable too?
```
try
    interval = int(os.environ.get('QUIPUCORDS_MANAGER_HEARTBEAT', '60'))
except ValueError:
    interval = 60
```

quipucords/scanner/manager.py
```diff
@@ -39,6 +39,20 @@ def __init__(self):
...
+        heartbeat = Timer(interval, self.log_info)
+        logger.debug('Scan manager running: %s. '
+                     'Current task: %s. '
+                     'Number of scans in queue: %s.',
+                     self.running, self.current_task, len(self.scan_queue))
+        if self.running:
+            heartbeat.start()
+        else:
+            if heartbeat.is_alive():
+                heartbeat.cancel()
```
chambridge: I don't understand this else condition. How could heartbeat be alive? Its a local variable.

kholdaway: Why do we need a new timer? Can't the manager just log in the run loop?

PR #1254, opened 2018-04-19
Modify fact endpoint source validation & update JSON to have source info

### LINKED ISSUE #1246: Modify fact endpoint so important source information is in JSON and relax source validation
## Specify type:
- Enhancement

## Description:
Currently, the fact endpoint validates that a source primary key is in the DB.  Additionally, the primary key is used to lookup additional source information during fingerprinting.  The enhancement should:
- Relax validation so that DB looks are not done
- Enhance details JSON to include key source JSON fields
- Update fingerprint engine's code so that it uses JSON instead of db lookups

abaiken:
Closes #1246


quipucords/fingerprinter/__init__.py
```diff
@@ -297,7 +293,11 @@ def _process_source(report_id, source):
 
         if fingerprint is not None:
             fingerprint['report_id'] = report_id
-            fingerprint[SOURCES_KEY] = [source_id]
+            fingerprint[SOURCES_KEY] = \
+                {source_name: {'source_type': source_type,
+                               'name': source_name}}
+            fingerprint['sources_info'] = [{'source_type': source_type,
```
kholdaway: Let's convert that to a dictionary merge --> https://github.com/quipucords/quipucords/blob/master/quipucords/fingerprinter/__init__.py#L539-L542

Issue #1277, opened 2018-04-24
UI front-end sprint: Bugs, features, and testing


cdcabrera:
## Specify type:
- Enhancement
- UI

## Description
- Bugs/issues
- Build/packaging/framework updates
- [x] Feature work #1280 
- [x] Copy, Design & View updates #1295 
- [x] Internationalization research and base
- Documentation/help component research
- Testing updates (ongoing)

cdcabrera:
Closing, stale

PR #1295, opened 2018-04-26
UI: Add Source Wizard, SSL Cert option

### LINKED ISSUE #1256: UI: Allow specification of SSL strictness
## Specify type:
- Enhancement

### Priority:
- Medium

___

## Description:
the UI provides no option to specify the --ssl-cert-verify option, forcing users to the default of "True" . The use of self-signed certs which require "--ssl-cert-verify False" is proving common.

As a QPC user who applies self-signed certificates, I would like a creation option added to the UI for Satellite and vCenter source types.  This option would be a checkbox, checked by default, and would bear the label "Require CA signed security certificates" .  When a source is submitted, its value for "--ssl-cert-verify" will be driven by the value of the checkbox.
___

## Acceptance Criteria:

- [ ] Verify that users can use the UI to create a source with SSL verification set to False
- [ ] Verify that users can use the UI to create a source with SSL verification set to True
- [ ] Verify that users can use the UI to edit a source and change the SSL verification value

___

cdcabrera:
* SSL cert option for vCenter and Satellite sources

  
screen shot 2018-04-25 at 11 58 12 pm

closes #1256
updates #1277 

@bclarhk @dozdowski

### DIFF
--- a/client/src/components/addSourceWizard/addSourceWizardStepTwo.js
+++ b/client/src/components/addSourceWizard/addSourceWizardStepTwo.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
-import { MenuItem, Button, Icon, Form, Grid } from 'patternfly-react';
+import { MenuItem, Button, Icon, Form, Grid, Checkbox } from 'patternfly-react';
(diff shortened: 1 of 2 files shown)


bclarhk:
alignment and wording look good

Issue #1285, opened 2018-04-24
Inconsistent JBoss EAP/Fuse/BRMS detection between root and sudo


kholdaway:
## Specify type:
- Bug

### Priority:
- High
___

## EAP Description:
System Information:
```
            "os_name": "Red Hat Enterprise Linux Server",
            "os_release": "Red Hat Enterprise Linux Server release 5.9 (Tikanga)",
            "os_version": "5.9 (Tikanga)",
            "infrastructure_type": "virtualized",
            "mac_addresses": [
                "00:50:56:9E:1E:CD"
            ],
            "ip_addresses": [
                "10.10.181.13"
            ],
```

Root finds:
```
{
    "name": "JBoss EAP",
    "version": [],
    "presence": "absent",
    "metadata": {
        "server_id": "b72e3509-8219-4044-8cad-b610c240eec3",
        "source_name": "NLargeSource",
        "source_type": "network",
        "raw_fact_key": "jboss_eap_id_jboss/jboss_eap_common_files/jboss_eap_locate_jboss_modules_jar/jboss_eap_chkconfig"
    }
},
```

Sudo finds:
```
{
    "name": "JBoss EAP",
    "version": null,
    "presence": "potential",
    "metadata": {
        "server_id": "b72e3509-8219-4044-8cad-b610c240eec3",
        "source_name": "NBecomeSource",
        "source_type": "network",
        "raw_fact_key": "jboss_eap_id_jboss/jboss_eap_common_files"
    }
},
```
___

## Fuse Description:
System information:
```
"name": "dhcp181-14.gsslab.rdu2.redhat.com",
            "os_name": "Red Hat Enterprise Linux Server",
            "os_release": "Red Hat Enterprise Linux Server release 5.9 (Tikanga)",
            "os_version": "5.9 (Tikanga)",
            "infrastructure_type": "virtualized",
            "mac_addresses": [
                "00:50:56:9E:27:35"
            ],
            "ip_addresses": [
                "10.10.181.14"
            ],
```

Root:
```
{
    "name": "JBoss Fuse",
    "version": [
        "Fuse-6.2.0"
    ],
    "presence": "present",
    "metadata": {
        "server_id": "b72e3509-8219-4044-8cad-b610c240eec3",
        "source_name": "NLargeSource",
        "source_type": "network",
        "raw_fact_key": "fuse_activemq_version/fuse_camel_version/fuse_cxf_version"
    }
},
```

Sudo:
```
{
[...]

PR #1311, opened 2018-05-02
Detect consistent presence of fuse between root & sudo user

abaiken:
2/3 of #1285

### DIFF
--- a/quipucords/scanner/network/processing/karaf.py
+++ b/quipucords/scanner/network/processing/karaf.py
@@ -124,7 +124,7 @@ class FuseVersionProcessor(process.Processor):
         results = []
         for item in output['results']:
             result = {}
-            item_name = item['item']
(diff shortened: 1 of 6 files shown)


quipucords/scanner/network/processing/karaf.py
```diff
@@ -124,15 +125,18 @@ def process(output, dependencies=None):
...
-            if item.get('rc', True):
-                pass
-            else:
-                if item['stdout'] != '':
-                    value = str(item['stdout']).splitlines()
-                    result['install_home'] = item_name
-                    result['version'] = list(set(value))
-            if result:
-                results.append(result)
+            if item_name:
```
kholdaway: Should we change line 128 to a item.get('item')? I realized you didn't add that code but I think we should move away for dict['key'].

PR #1333, opened 2018-05-10
Vary error messages to help debug errors

kholdaway:


quipucords/scanner/satellite/connect.py
```diff
@@ -75,27 +75,32 @@ def run(self, manager_interrupt):
                     % self.source.name
                 return error_message, ScanTask.FAILED
         except SatelliteAuthException as auth_error:
-            error_message = 'Satellite error encountered: %s. ' % auth_error
+            error_message = 'Satellite auth error encountered: %s. ' \
```
abaiken: Not a blocker but you could move these into a common messages file since they're used in multiple places.

kholdaway: True

PR #1344, opened 2018-05-23
Various fixes

### LINKED ISSUE #1343: Default options for JBoss products incorrect when options not include in scan
## Specify type:
- Bug

### Priority:
- High
___

## Description:
When a scan is created, omitting the options from the JSON body will cause JBoss products to be excluded from scan.  The default should be to include them.

This is not a problem via the CLi, but is a problem with the UI or API (e.g. Postman)

___

## Acceptance Criteria:

- [ ] Verify that JBoss is enabled by default when options is missing
___

kholdaway:
- Fix default for scan options when options missing
- Add env to allow internal facts to be part of details report
- Add environment variables to startup log and status endpoint

Closes #1343

### DIFF
--- a/quipucords/api/scan/tests_scan.py
+++ b/quipucords/api/scan/tests_scan.py
@@ -423,7 +423,7 @@ class ScanTest(TestCase):
                          status.HTTP_204_NO_CONTENT)
 
     def test_get_extra_vars_missing_options(self):
-        """Tests the get_extra_vars with extended search None."""
+        """Tests the get_default_extra_vars."""
         extra_vars = ScanOptions.get_default_extra_vars()
 
         expected_vars = {'jboss_eap': True,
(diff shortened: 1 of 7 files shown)


abaiken:
Not a blocker but you could add a new test for when no options are passed through.

quipucords/api/scan/tests_scan.py
```diff
@@ -421,6 +422,18 @@ def test_delete(self):
         self.assertEqual(response.status_code,
                          status.HTTP_204_NO_CONTENT)
 
+    def test_get_extra_vars_missing_options(self):
+        """Tests the get_extra_vars with extended search None."""
```
abaiken: Docstring feels copied.

kholdaway: Good catch

PR #1356, opened 2018-06-07
Added ability to exlude ip addresses from CLI

### LINKED ISSUE #1351: Add ability to exclude IP addresses to the REST Source CLI for network scans
## Specify type:
- Enhancement

## Description:
Modify the `qpc source add` and `qpc source edit` CLI to allow exclusion of IP addresses and ranges.  The focus of this story is the CLI.  Other issues will capture API and scanning.  This issue is dependent on #1350.

Add an option to both add/update called `--exclude-hosts`.  The command should have syntax, help, and documentation that is similar to the `--hosts` parameter.
___

## Acceptance Criteria:

- [ ] `qpc source add` with --exclude-hosts passes values to the create source REST API
- [ ] `qpc source edit` with --exclude-hosts passes values to the partial update source REST API
- [ ] Verify that omitting this --exclude-hosts attribute is allow
- [ ] Verify that the --exclude-hosts attribute will accept more than one ip, host, or range.
- [ ] Man page and user documentation is updated for the `qpc source add/edit` commands
- [ ] Unit tests cover newly created function

cixil:
closes #1351


docs/source/man.rst
```diff
@@ -245,6 +245,10 @@ To create a source, supply the type of source with the ``type`` option, one or m
 
     ``--hosts /home/user1/hosts_file``
 
+``--exclude-hosts ip_address``
+
+  Sets the host name, IP address, or IP address range to exclude when running a scan. Follows the same formatting options as ``--hosts`` shown above.
```
abaiken: I would put `Optional.` at the beginning of the description so that we follow the flow of the other source options that aren't required.

docs/source/man.rst
```diff
@@ -211,9 +211,9 @@ When you configure a scan, it contains references to one or more sources, includ
...
 
-**qpc source add --name=** *name*  **--type=** *(network | vcenter | satellite)* **--hosts** *ip_address* **--cred** *credential* **[--port=** *port* **]** **[--ssl-cert-verify=** *(True | False)* **]** **[--ssl-protocol=** *protocol* **]** **[--disable-ssl=** *(True | False)* **]**
+**qpc source add --name=** *name*  **--type=** *(network | vcenter | satellite)* **--hosts** *ip_address* **[--exclude-hosts** *ip_address* **]** **--cred** *credential* **[--port=** *port* **]** **[--ssl-cert-verify=** *(True | False)* **]** **[--ssl-protocol=** *protocol* **]** **[--disable-ssl=** *(True | False)* **]**
```
abaiken: Not a blocker and Kevan may have a different opinion but since `type`, `hosts`, and `cred` are required for a source, I feel like they should come before the `exclude_hosts` option.

docs/source/man.rst
```diff
@@ -211,9 +211,9 @@ When you configure a scan, it contains references to one or more sources, includ
 Creating and Editing Sources
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-To create a source, supply the type of source with the ``type`` option, one or more host names or IP addresses to connect to with the ``--hosts`` option, and the credentials needed to access those systems with the ``--cred`` option. The ``qpc source`` command allows multiple entries for the ``hosts`` and ``cred`` options. Therefore, a single source can access a collection of servers and subnets as needed to create an accurate and complete scan.
+To create a source, supply the type of source with the ``type`` option, one or more host names or IP addresses to connect to with the ``--hosts`` option, and the credentials needed to access those systems with the ``--cred`` option. When creating a ``network type`` source, to specify a host or range of hosts to be excluded from the scan use the ``--exclude-hosts`` option. The ``qpc source`` command allows multiple entries for the ``hosts`` and ``cred`` options. Therefore, a single source can access a collection of servers and subnets as needed to create an accurate and complete scan.
```
abaiken: Since this is an optional argument, I don't think that we should include it in this paragraph because it seems like the purpose of this paragraph is giving the essentials to creating a source. 

If we do want to include this I would change it from ``network type`` to ``network``,  and make it clear that this is an optional flag (not required like type/cred/hosts). I would also add ``exclude_hosts`` in the sentence below where it says ```allows multiple entries for the hosts and cred options``` since `exclude_hosts` also allows multiple entries.

docs/source/working_with_sources.rst
```diff
@@ -56,6 +56,12 @@ To create a network source, use the following steps:
 
    # qpc source add --type network --name mynetwork --hosts /home/user1/hosts_file --cred roothost1 roothost2
 
+ 
+   **TIP:** You can optionally supply an exclude-hosts field that will omit hosts from a scan.
```
abaiken: I would say `For network sources, you can optionally . . .` or you can word it how you want, we just need to specify that this option is only available for network sources.

qpc/source/tests_source_add.py
```diff
@@ -213,6 +217,88 @@ def test_add_source_net(self):
...
+                                            'my-rhel[a:d].company.com'],
+                             type='network',
+                             port=22)
+            with redirect_stdout(source_out):
+                nac.main(args)
+                self.assertEqual(source_out.getvalue(),
+                                 messages.SOURCE_ADDED % 'source1' + '\n')
+
+    ##################################################
+    # Vcenter Source Test
```
abaiken: I like the separators that you added 😄

quipucords/api/source/model.py
```diff
@@ -83,7 +83,7 @@ class Source(models.Model):
         SourceOptions, null=True, on_delete=models.CASCADE)
     credentials = models.ManyToManyField(Credential)
     hosts = models.TextField(unique=False, null=False)
-    exclude_hosts = models.TextField(unique=False, null=True)
+    exclude_hosts = models.TextField(unique=False, null=False)
```
abaiken: I'm pretty sure this should be set to `null=True`.

docs/source/working_with_sources.rst
cixil: This is already under the 'Creating a Network Source' section.

quipucords/api/source/model.py
cixil: If it is set to true the server returns 
`django.db.utils.IntegrityError: NOT NULL constraint failed: api_source.exclude_hosts`
after running a `qpc source add .... ` (This isn't covered in the code coverage tests)
From django model reference https://docs.djangoproject.com/en/2.0/ref/models/fields/:  
If null=True, Django will store empty values as NULL in the database. Default is False.

docs/source/working_with_sources.rst
abaiken: No need! I didn't see that. You can leave it how it is 👍

quipucords/api/migrations/0001_initial.py
```diff
@@ -1,358 +0,0 @@
-# -*- coding: utf-8 -*-
```
kholdaway: Do not remove this file.

quipucords/api/migrations/0002_source_exclude_hosts.py
```diff
@@ -1,20 +0,0 @@
-# -*- coding: utf-8 -*-
```
kholdaway: Do not remove this file.

kholdaway:
🎈

PR #1359, opened 2018-06-11
added scanner capability to exclude hosts

cixil:
I didn't see anything that needed to be added to the docs.
Also the doc changes I added in the last merge aren't showing up yet on https://quipucords.readthedocs.io/en/0.0.42-pilot/man.html

(which they shouldn't really until this PR is merged and they become true)

### DIFF
--- a/quipucords/scanner/network/connect.py
+++ b/quipucords/scanner/network/connect.py
@@ -26,7 +26,6 @@ from django.db import transaction
 
 import pexpect
 
-
 from scanner.network.connect_callback import ConnectResultCallback
 from scanner.network.utils import (_construct_error,
                                    _construct_vars,
@@ -55,10 +54,19 @@ class ConnectResultStore(object):
         # Sources can contain patterns that describe multiple hosts,
         # like '1.2.3.[4:6]'. Expand the patterns so hosts is a list
         # of single hosts we can try to connect to.
-        hosts = []
+        hosts, exclude_hosts = [], []
+
         hosts_list = source.get_hosts()
+        exclude_hosts_list = source.get_exclude_hosts()
+
         for host in hosts_list:
             hosts.extend(expand_hostpattern(host))
+        for host in exclude_hosts_list:
+            exclude_hosts.extend(expand_hostpattern(host))
+
+        # Removes excluded ip addresses from the hosts list.
+        hosts = [host for host in hosts if host not in exclude_hosts]
+
         self._remaining_hosts = set(hosts)
 
         scan_task.update_stats('INITIAL NETWORK CONNECT STATS.',
@@ -200,7 +208,9 @@ class ConnectTaskRunner(ScanTaskRunner):
 
 
 # pylint: disable=too-many-arguments
-def connect(hosts, callback, credential, connection_port, forks=50):
+# pylint: disable=dangerous-default-value
+def connect(hosts, callback, credential, connection_port, forks=50,
+            exclude_hosts=[]):
     """Attempt to connect to hosts using the given credential.
 
     :param hosts: The collection of hosts to test connections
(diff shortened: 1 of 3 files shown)


quipucords/scanner/network/connect.py
```diff
@@ -55,10 +54,19 @@ def __init__(self, scan_task):
...
         hosts_list = source.get_hosts()
+        exclude_hosts_list = source.get_exclude_hosts()
+
         for host in hosts_list:
             hosts.extend(expand_hostpattern(host))
+        for host in exclude_hosts_list:
+            exclude_hosts.extend(expand_hostpattern(host))
+
+        # Removes excluded ip addresses from the hosts list.
+        hosts = [host for host in hosts if host not in exclude_hosts]
```
kholdaway: you can just convert these to sets and perform a diff.  

```
s1 = set([1,2,3])
s2 = set([3,4,5])

s1_not_in_s2 =  s1 - s2
```

quipucords/scanner/network/connect.py
```diff
@@ -200,18 +208,22 @@ def run_with_result_store(self, result_store):
 
 
 # pylint: disable=too-many-arguments
-def connect(hosts, callback, credential, connection_port, forks=50):
+# pylint: disable=dangerous-default-value
+def connect(hosts, callback, credential, connection_port, forks=50,
```
kholdaway: Probably should use None and not [].  Also remove the disable=dangerous-default-value

quipucords/scanner/network/connect.py
```diff
@@ -254,18 +266,21 @@ def _handle_ssh_passphrase(credential):
             pass
 
 
-def construct_connect_inventory(hosts, credential, connection_port):
+# pylint: disable=dangerous-default-value
```
kholdaway: same

quipucords/scanner/network/connect.py
```diff
@@ -254,18 +266,21 @@ def _handle_ssh_passphrase(credential):
...
     :param connection_port: The connection port
-    :returns: A dictionary of the ansible invetory
+    :param exclude_hosts: Optional. Hosts to exclude test connections
+    :returns: A dictionary of the ansible inventory
     """
     inventory = None
     hosts_dict = {}
 
-    for host in hosts:
+    for host in [host for host in hosts if host not in exclude_hosts]:
```
kholdaway: Use set subtraction here too.  However, do we need this if we performed the set difference in the constructor?

kholdaway:
Maybe add a test case where exclude contains something not in the hosts.  Make sure meaningless excludes are ignored.

quipucords/scanner/network/connect.py
cixil: Yes, because this function is defined outside the scope of the class with the constructor.

quipucords/scanner/network/tests_network_connect.py
```diff
@@ -124,11 +124,12 @@ def setUp(self):
 
         self.source = Source(
             name='source1',
+            hosts='["1.2.3.4", "1.2.3.5"]',
```
abaiken: We probably should set up an additional source that does not have `exclude_hosts` to ensure that the we have tested the case where it is not specified. IE. We need to make sure that everything has been tested with & without `exclude_hosts`.

For example when we call the `get_exclude_hosts` from the source model, we need to make sure that it will return [] if the source was created without `exclude_hosts`.

quipucords/scanner/network/connect.py
```diff
@@ -254,19 +266,25 @@ def _handle_ssh_passphrase(credential):
...
-    :returns: A dictionary of the ansible invetory
+    :param exclude_hosts: Optional. Hosts to exclude test connections
+    :returns: A dictionary of the ansible inventory
     """
     inventory = None
     hosts_dict = {}
 
-    for host in hosts:
-        hosts_dict[host] = None
+    if exclude_hosts is None:
```
abaiken: I feel like we should be calculating what our finalized hosts are before we initializing them inside the `hosts_dict` so that we are not repeating code (`hosts_dict[host]=None`). 

1. Calculate hosts (do the set subtraction)
2. Loop through the finalized hosts/ add them to the `hosts_dict`

kholdaway: 1. Would be do the set subtraction if excludes is not None.

abaiken:
Great job!!!!!

PR #1371, opened 2018-06-28
Use RetrieveProperties to get all VM names

agrare:
Replace the loop over all VMs and individual calls to retrieve the VM names with a single RetrieveProperties to get all names in one shot.

This reduces the time it takes to run the get_vm_names method from 12min 53sec to 22sec on a 10,240VM vCenter.


agrare:
@kholdaway okay this is ready to review

agrare:
Thanks @kholdaway !

PR #1380, opened 2018-07-17
Detect JBoss Web Server/JBoss Enterprise Web Server installations

### LINKED ISSUE #1373: Detect JBoss Web Server/JBoss Enterprise Web Server installations
## Specify type:
- Enhancement

### Priority:
- High
___

## Description:
Detect JBoss Web Server/JBoss Enterprise Web Server installations.  Fingerprint document provided by SEAP -> https://docs.google.com/document/d/1red3mS7NVvrbk5NYkEGD8JTJbi5CFh9oLo4GEkvfMyk/edit

___

cixil:
WIP
Current methods of detecting jws (other than definitively checking with yum) attempt to find the directory above the tomcat home directory (which is jws_home for the default zip installation) and then find signs that tomcat was installed with a redhat/jboss product.

TO DO:

- [x] Detect versions prior to 3.0.0
~~- [ ] Ensure detection is accurate when all packages for JWS are installed via rpm but not as a group (ie yum has groups jws3 and jws5 to make installation easier)~~ - If valid combinations of versions of tomcat and apache webserver are found, product will be marked as potential.
- [x] Change presence status of jws from potential to present where possible (see if presence of JBossEULA.txt in "JWS_HOME" ensures that jws is present, etc)
- [x] Fix errors found during our full 26 system scan
- [x] Move code that parses ansible response to a scanner post processor instead of doing this in the fingerprint engine. 
- [x] Add JWS to the disable optional productions feature (API and CLI)
- [x] Add JWS to the extended production search feature (API and CLI)
- [x] Update swagger to reflect new API
closes #1373


roles/jboss_ws/tasks/main.yml
```diff
@@ -0,0 +1,93 @@
...
+
+- name: ls JWS_HOME
+  raw: ls -d1 "{{ item }}" 2>/dev/null
+  register: find_jws_home
+  with_items: "{{ jws_home_candidates }}"
+  ignore_errors: yes
+
+- name: set jws_home based on find_jws_home
+  set_fact:
+      jws_home: "{{ find_jws_home['results'][0]['stdout_lines'][0] }}"
```
chambridge: I'd suggest adding some guards to this in case you are trying to index into the dictionary or list and its not there.
i.e.
```
when:
  - 'results' in find_jws_home
  - find_jws_home['results'] | length > 0
  - 'stdout_lines' in find_jws_home['results'][0]
  - find_jws_home['results'][0]['stdout_lines'] | length > 1

File: roles/jboss_ws/tasks/main.yml [hunk 2]
```diff
@@ -0,0 +1,93 @@
...
+# ------------------------------ DETECT JWS/EWS -------------------------------
+- name: check jws presence with yum
+  raw: yum grouplist jws3 jws3plus jws5 | grep -A1 'Installed Groups:' | grep 'Red Hat JBoss Web Server [0-99]'
+  register: check_installed_with_rpm
+  ignore_errors: yes
+
+- name: determine if tomcat is installed from upstream or as part of redhat product
+  raw: \[ $(grep -iR redhat "{{ jws_home }}"/tomcat[0-99] 2>/dev/null| wc -l) != 0 \] && echo True || echo False
+  register: tomcat_is_not_upstream
+  ignore_errors: yes
```
chambridge: if you add the when conditions above you would need to add a check that `jws_home` is defined.
```
when: jws_home is defined
```

roles/jboss_ws/tasks/main.yml
```diff
@@ -0,0 +1,93 @@
...
+
+- name: set tomcat_is_part_of_redhat_product based on tomcat_is_not_upstream 
+  set_fact:
+      tomcat_is_part_of_redhat_product: "{{ tomcat_is_not_upstream['stdout_lines'][0] }}"
+  ignore_errors: yes
+
+- name: set installed_with_rpm based on check_installed_with_rpm 
+  set_fact:
+      installed_with_rpm: "{{ check_installed_with_rpm['stdout_lines']}}"
```
chambridge: Use `get` method instead assuming 'stdout_lines' is there.
`check_installed_with_rpm.get('stdout_lines', [])`

roles/jboss_ws/tasks/main.yml
```diff
@@ -0,0 +1,93 @@
...
+  ignore_errors: yes
+
+- name: determine if tomcat is installed from upstream or as part of redhat product
+  raw: \[ $(grep -iR redhat "{{ jws_home }}"/tomcat[0-99] 2>/dev/null| wc -l) != 0 \] && echo True || echo False
+  register: tomcat_is_not_upstream
+  ignore_errors: yes
+
+- name: set tomcat_is_part_of_redhat_product based on tomcat_is_not_upstream 
+  set_fact:
+      tomcat_is_part_of_redhat_product: "{{ tomcat_is_not_upstream['stdout_lines'][0] }}"
```
chambridge: You should add guards here or handle this with a post processor.

roles/jboss_ws/tasks/main.yml
```diff
@@ -0,0 +1,93 @@
...
+
+  # detect jws if installed with zip even when no tomcat process is running
+- name: see if JBossEULA.txt exists 
+  raw: ls "{{ jws_home }}"/JBossEULA.txt
+  register: jboss_eula_txt_file
+  ignore_errors: yes
+
+- name: set jboss_eula_location based on jboss_eula_txt_file 
+  set_fact:
+      jboss_eula_location: "{{ jboss_eula_txt_file['stdout_lines'][0] }}"
```
chambridge: You should add guards here or handle this with a post processor.

Will need a check for `jboss_eula_txt_file is defined` along with indexing.

roles/jboss_ws/tasks/main.yml
```diff
@@ -0,0 +1,93 @@
...
+- name: set installed_with_rpm based on check_installed_with_rpm 
+  set_fact:
+      installed_with_rpm: "{{ check_installed_with_rpm['stdout_lines']}}"
+  ignore_errors: yes
+
+  # detect jws if installed with zip even when no tomcat process is running
+- name: see if JBossEULA.txt exists 
+  raw: ls "{{ jws_home }}"/JBossEULA.txt
+  register: jboss_eula_txt_file
+  ignore_errors: yes
```
chambridge: if you add the when conditions above you would need to add a check that `jws_home` is defined.
```
when: jws_home is defined
```

roles/jboss_ws/tasks/main.yml
```diff
@@ -0,0 +1,93 @@
...
+- name: find JWS version
+  raw: "{{ item }}"
+  register: jws_version
+  ignore_errors: yes
+  with_items: 
+    - ."{{ jws_home }}"/httpd/sbin/httpd -v | grep Apache 2>/dev/null
+    - awk -F 'id=' '/JWS/ {print substr($2,16,9)}' "{{ jws_home }}"/tomcat?/webapps/docs/changelog.html | grep JWS | head -n1 2>/dev/null
+    - cat "{{ jws_home }}"/version.txt | grep "Red Hat JBoss Web Server" 2>/dev/null
+    - systemctl status jws5-tomcat.service | grep "Apache Tomcat Web" | grep -o jws5
+    - (echo -n $("{{ jws_home }}"/httpd/sbin/httpd -v | grep Apache); echo $(./"{{ jws_home }}"/tomcat*/bin/version.sh | grep "Server version")) | sed -e 's/Server version: //g'
```
chambridge: if you add the when conditions above you would need to add a check that `jws_home` is defined.
```
when: jws_home is defined
```

Looks like you could still to the `systemctl` check.

You should check that they have `systemctl` older systems wont they will have `chkconfig` with the `internal_have_systemctl` fact:
https://github.com/quipucords/quipucords/blob/master/roles/jboss_eap/tasks/main.yml#L160

I assume you should also add a check using chkconfig as well.
https://github.com/quipucords/quipucords/blob/master/roles/jboss_eap/tasks/main.yml#L167

Each of these items should have the `2>/dev/null` before the first pipe.

roles/jboss_ws/tasks/main.yml
```diff
@@ -39,52 +38,31 @@
...
+    - "jws_find_home['results'] | length > 0"
+    - "'stdout_lines' in jws_find_home['results'][0]"
+    - "jws_find_home['results'][0]['stdout_lines'] | length > 0"
 
 # ------------------------------ DETECT JWS/EWS -------------------------------
 - name: check jws presence with yum
-  raw: yum grouplist jws3 jws3plus jws5 | grep -A1 'Installed Groups:' | grep 'Red Hat JBoss Web Server [0-99]' 2>/dev/null
-  register: jws_check_installed_with_rpm
+  raw: yum grouplist jws3 jws3plus jws5 2>/dev/null | grep -A1 'Installed Groups:' | grep 'Red Hat JBoss Web Server [0-99]' 
+  register: jws_installed_with_rpm
```
chambridge: Didn't catch this last time, but `yum` will require privilege escalation. You should add `become: True` so sudo, etc will be used.

roles/jboss_ws/tasks/main.yml
```diff
@@ -103,7 +81,7 @@
   ignore_errors: yes
   with_items:
     - ."{{ jws_home }}"/httpd/sbin/httpd -v | grep Apache 2>/dev/null
-    - awk -F 'id=' '/JWS/ {print substr($2,16,9)}' "{{ jws_home }}"/tomcat?/webapps/docs/changelog.html 2>/dev/null | grep JWS | head -n1
+    - awk -F 'id=' '/JWS/ {print substr($2,16,9)}' "{{ jws_home }}"/tomcat?/webapps/docs/changelog.html 2>/dev/null | grep JWS | head -n1 
     - cat "{{ jws_home }}"/version.txt  2>/dev/null| grep "Red Hat JBoss Web Server"
     - systemctl status jws5-tomcat.service 2>/dev/null | grep "Apache Tomcat Web" | grep -o jws5
```
chambridge: Did you decide against pulling out the `systemctl` check to its own task (along with creating one for `chkconfg`?

roles/jboss_ws/tasks/main.yml
cixil: yum grouplist doesn't require sudo

quipucords/fingerprinter/tests_product_jws.py
```diff
@@ -0,0 +1,121 @@
...
+from unittest.mock import patch
+
+from api.models import ServerInformation
+
+from django.test import TestCase
+
+from fingerprinter.jboss_web_server import detect_jboss_ws, get_version
+
+
+class ProductFuseTest(TestCase):
```
abaiken: The class & docstring should be updated to reflect JWS instead of Fuse

quipucords/fingerprinter/jboss_web_server.py
```diff
@@ -0,0 +1,103 @@
...
+
+"""Ingests raw facts to determine the status of JBoss Web Server on system."""
+
+import logging
+
+from api.models import Product
+
+from fingerprinter.utils import product_entitlement_found
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
```
abaiken: Are we ever actually logging anything?

quipucords/fingerprinter/tests_product_jws.py
abaiken: Update this class & docstring to reflect JWS instead of Fuse

quipucords/scanner/network/processing/jws.py
```diff
@@ -0,0 +1,63 @@
...
+# along with this software; if not, see
+# https://www.gnu.org/licenses/gpl-3.0.txt.
+
+"""Initial processing of raw shell output from Ansible commands."""
+
+import logging
+
+from scanner.network.processing import process
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
```
abaiken: Are we logging anything?

quipucords/scanner/network/processing/jws.py
```diff
@@ -0,0 +1,63 @@
...
+    """Process output of search for redhat string in tomcat files."""
+
+    KEY = 'tomcat_is_part_of_redhat_product'
+
+    @staticmethod
+    def process(output, dependencies=None):
+        """Return either True or False."""
+        result = output.get('stdout_lines', False)
+        if result:
+            result = result[0]
```
abaiken: instead of reassigning `result` here, you could say:
```
if result:
    if result[0] == 'True':
        return True
return False
``` 
but this is no biggie, so if you want to leave it the way it is, I'm cool with that.

kholdaway:
Swagger needs to be updated.  No changes have been made.

roles/jboss_ws/tasks/main.yml
```diff
@@ -0,0 +1,128 @@
...
+# Find directory above tomcat home with find
+
+- name: combine JWS_HOME candidates into a list
+  set_fact:
+    jws_home_candidates: "{{
+        (internal_jws_find_home_from_search.get('stdout_lines', []) +
+         internal_jws_find_home_from_tomcat_file.get('stdout_lines', []) +
+         internal_jws_find_home_from_tomcat_process.get('stdout_lines', []))|difference(['']) | unique }}"
+  ignore_errors: yes
+  when: 'jboss_ws'
```
kholdaway: Doesn't this require JBoss Web Server to be running?  Can't we detect it using extended search when it is not running?

roles/jboss_ws/tasks/main.yml
```diff
@@ -0,0 +1,128 @@
...
+    - "internal_jws_find_home['results'] | length > 0"
+    - "'stdout_lines' in internal_jws_find_home['results'][0]"
+    - "internal_jws_find_home['results'][0]['stdout_lines'] | length > 0"
+
+# ------------------------------ DETECT JWS/EWS -------------------------------
+- name: check jws presence with yum
+  raw: yum grouplist jws3 jws3plus jws5 2>/dev/null | grep -A1 'Installed Groups:' | grep 'Red Hat JBoss Web Server [0-99]' 
+  register: jws_installed_with_rpm
+  ignore_errors: yes
+  when: 'jboss_ws'
```
kholdaway: Same.  Is there a reason we need the server to be running for this yum check?

roles/jboss_ws/tasks/main.yml
```diff
@@ -0,0 +1,128 @@
...
+
+  # If tomcat was installed with a Redhat product, a 'redhat' string can be found in certain files
+- name: determine if tomcat is installed from upstream or as part of redhat product
+  raw: \[ $(grep -iR redhat "{{ jws_home }}"/tomcat[0-99] 2>/dev/null| wc -l) != 0 \] && echo False || echo True
+  register: tomcat_is_part_of_redhat_product
+  ignore_errors: yes
+  when: 'jws_home is defined and jboss_ws'
```
kholdaway: Same.

roles/jboss_ws/tasks/main.yml
cixil: Yeah sorry, added the result from extended for home and forgot to push the commit. I'll add a check for the JWS_HOME environment variable too, in case they have it set.

PR #1389, opened 2018-07-27
enable connection through paramiko rather than ssh

### LINKED ISSUE #1364: Allow users to specify paramiko for the connection type used by ansible
## Specify type:
- Enhancement

### Priority:
- Medium
___

## Description:
Open-ssh is the default ansible connection method but certain customers have versions of open-ssh which aren't compatible so we should add an option to QPC to allow them to use paramiko. 

The option when running ansible is `-c paramiko` appended to the ansible or ansible-playbook command. 

The default should always be to not append -c paramiko but if there is a flag specified on the scan command like `--use_paramiko` then we should add the `-c paramiko` option to anywhere Ansible is used (discovery and fact collection). 

___

## Acceptance Criteria:

- [ ] Verify that users can specify an option and it forces ansible to use

cixil:
To Do:
Write tests

closes #1364


qpc/source/utils.py
```diff
@@ -61,6 +61,8 @@ def build_source_payload(args, add_none=True):
         req_payload['hosts'] = args.hosts
     if hasattr(args, 'exclude_hosts') and args.exclude_hosts:
         req_payload['exclude_hosts'] = args.exclude_hosts
+    if hasattr(args, 'use_paramiko') and args.use_paramiko:
```
abaiken: since `args.use_paramiko` is a boolean, we want to check that it is not None here so it would be like: 
```
if hasattr(...) and args.use_paramiko is not None:
```

quipucords/api/migrations/0002_source_exclude_hosts.py
```diff
@@ -1,20 +0,0 @@
-# -*- coding: utf-8 -*-
```
abaiken: why are we deleting this?

quipucords/scanner/network/connect.py
```diff
@@ -162,11 +162,12 @@ def run_with_result_store(self, result_store):
...
 
-        if self.scan_job.options is None:
-            forks = ScanOptions.get_default_forks()
-        else:
+        if self.scan_job.options is not None:
             forks = self.scan_job.options.max_concurrency
+        else:
+            forks = ScanOptions.get_default_forks()
         connection_port = source['port']
+        use_paramiko = source['use_paramiko']
```
abaiken: Does this throw an error if `use_paramiko` is not specified when creating the source?

quipucords/scanner/network/connect.py
```diff
@@ -162,11 +162,12 @@ def run_with_result_store(self, result_store):
         serializer = SourceSerializer(self.scan_task.source)
         source = serializer.data
 
-        if self.scan_job.options is None:
-            forks = ScanOptions.get_default_forks()
-        else:
+        if self.scan_job.options is not None:
```
abaiken: Out of curiosity, is there a reason we switched this if..else?

quipucords/scanner/network/utils.py
```diff
@@ -92,23 +92,33 @@ def write_inventory(inventory):
...
                                            'forks', 'become', 'become_method',
                                            'become_user', 'check'])
-    options = named_options(connection='ssh',
-                            module_path=C.DEFAULT_MODULE_PATH,
-                            forks=forks, become=False,
-                            become_method='sudo', become_user='root',
-                            check=False)
+    if use_paramiko:
+        # Change connection method from default ssh to paramiko
+        options = named_options(connection='paramiko',
```
abaiken: instead of rebuilding the options based on whether or not we use paramiko, can we just have a variable for connection that defaults to ssh, but if `use_paramiko` is True, we change it to `paramiko`?

quipucords/scanner/network/connect.py
cixil: It was done that way in another file so I switched it for consistency. Doesn't really matter.

docs/source/man.rst
```diff
@@ -213,7 +213,7 @@ Creating and Editing Sources
 
 To create a source, supply the type of source with the ``type`` option, one or more host names or IP addresses to connect to with the ``--hosts`` option, and the credentials needed to access those systems with the ``--cred`` option. The ``qpc source`` command allows multiple entries for the ``hosts`` and ``cred`` options. Therefore, a single source can access a collection of servers and subnets as needed to create an accurate and complete scan.
 
-**qpc source add --name=** *name*  **--type=** *(network | vcenter | satellite)* **--hosts** *ip_address* **--cred** *credential* **[--exclude-hosts** *ip_address* **]** **[--port=** *port* **]** **[--ssl-cert-verify=** *(True | False)* **]** **[--ssl-protocol=** *protocol* **]** **[--disable-ssl=** *(True | False)* **]**
+**qpc source add --name=** *name*  **--type=** *(network | vcenter | satellite)* **--hosts** *ip_address* **--cred** *credential* **[--exclude-hosts** *ip_address* **]** **[--port=** *port* **]** **[--use-paramiko=** *(True | False)* **]** **[--ssl-cert-verify=** *(True | False)* **]** **[--ssl-protocol=** *protocol* **]** **[--disable-ssl=** *(True | False)* **]**
```
kholdaway: Let's add `use-paramiko` to the `qpc source edit` docs (line 284)

quipucords/api/migrations/0002_source_exclude_hosts.py
kholdaway: Yeah, let's put the entire file back.

quipucords/api/migrations/0001_initial.py
```diff
@@ -1,358 +0,0 @@
-# -*- coding: utf-8 -*-
```
kholdaway: Put this file back.

quipucords/api/source/serializer.py
```diff
@@ -79,6 +79,7 @@ class SourceSerializer(NotEmptySerializer):
     source_type = ValidStringChoiceField(
         required=False, choices=Source.SOURCE_TYPE_CHOICES)
     port = IntegerField(required=False, min_value=0, allow_null=True)
+    use_paramiko = NullBooleanField(required=False)
```
kholdaway: Two things.  Move to SourceOptionSerializer and document in swagger.

quipucords/api/source/model.py
```diff
@@ -79,6 +79,7 @@ class Source(models.Model):
         null=False
     )
     port = models.IntegerField(null=True)
+    use_paramiko = models.NullBooleanField(default=False)
```
kholdaway: I think this should be part of the SourceOptions.

quipucords/scanner/network/connect.py
kholdaway: Setting should be handled if options is None.

quipucords/api/migrations/0002_source_exclude_hosts.py
cixil: It has to do with the migrations, I think it was accidentally added earlier.

cixil: I think this was added by accident.

quipucords/api/migrations/0001_initial.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Generated by Django 1.11.5 on 2018-04-19 15:22
```
kholdaway: Please restore the exact migration from master

quipucords/api/migrations/0002_source_exclude_hosts.py
kholdaway: This was deleted again

kholdaway: Additionally, there will need to be a migration 3 for the use-paramiko.

quipucords/scanner/network/inspect.py
```diff
@@ -257,8 +262,8 @@ def _inspect_scan(self, connected, roles=DEFAULT_ROLES,
         inventory_file = write_inventory(inventory)
 
         error_msg = ''
-        log_message = 'START PROCESSING GROUPS with %d forks' \
-            ' and extra_vars=%s' % (forks, extra_vars)
+        log_message = 'START PROCESSING GROUPS with use_paramiko: %d, \
```
kholdaway: Let's change this log to say with `openssh` or `paramiko` instead of use_paramiko.

qpc/source/utils.py
```diff
@@ -89,6 +87,10 @@ def build_source_payload(args, add_none=True):
             options = {'ssl_protocol': args.ssl_protocol}
         else:
             options['ssl_protocol'] = args.ssl_protocol
+    if hasattr(args, 'use_paramiko') and args.use_paramiko is not None:
+        # Don't have to to check if options is None because paramiko is the
+        # only valid option for network source
+        options = {'use_paramiko': args.use_paramiko}
```
kholdaway: Let's add a None check here so that the ssl options are not removed.  The server will respond with a validation error, but that is okay.  It will tell the user what they should change.

kholdaway:
Looks great @cixil!  Nice job!  This is a great feature.

PR #1394, opened 2018-08-01
Fixed yum history date issue

kholdaway:


roles/date/tasks/main.yml
```diff
@@ -48,8 +48,5 @@
 
 - name: extract result value for date.yum_history
   set_fact:
-    date_yum_history: "{{  internal_date_yum_history_cmd['stdout_lines'] | select | first if internal_have_yum else '' }}"
+    date_yum_history: "{{  internal_date_yum_history_cmd }}"
```
chambridge: Can we just pull `internal` and `_cmd` off and not have the `set_fact` task?

kholdaway: Done.

PR #1428, opened 2018-09-05
Remove joins from scan manager to fix scan job cancel

### LINKED ISSUE #1416: Canceling a scan job when QPC is running in gunicorn restarts server
## Bug Description:
Canceling a scan job when QPC is running in gunicorn (how it runs in docker) restarts server.  The restart occurs before the job state is updated to a canceled state.  This causes the job to restart too.

In digging in, it seems this is a side effect of running with gunicorn.  

### Expected behavior:
Our algorithm to cancel is job is:
1) Kill job process
2) Update job state to cancelled

### Actual behavior:
Job cannot be canceled.  The following sequence occurs:
1) Kill job process
2) gunicorn worker recreated and server restarts
3) Rerun job

### Steps to reproduce:
Run QPC in the docker container and cancel a running job.
___

## Acceptance Criteria:

- [ ] A job can be canceled in the docker container

___

### Version Information:
QPC 0.0.43

kholdaway:
This work intends to ensure that a job will:
- Cause gunicorn to timeout while canceling a job
- Really cancel a job (at least by the end of the inventory group)
- Everything else should work the same

Closes #1416

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -122,10 +122,13 @@ You can run the server locally inside of gunicorn.  This can be a useful way to 
 3. Make symbolic link to ansible roles::
     ln -s ../roles/ roles
 
-4. Start gunicorn::
+4. Install gunicorn::
+    pip install gunicorn==19.7.1
+
+5. Start gunicorn::
     gunicorn quipucords.wsgi -c ./local_gunicorn.conf.py
 
-5. Configure the CLI by using the following commands::
+6. Configure the CLI by using the following commands::
     qpc server config --host 127.0.0.1 --port 8000
     qpc server login
 
--- a/README.rst
+++ b/README.rst
@@ -120,15 +120,19 @@ You can run the server locally inside of gunicorn.  This can be a useful way to 
     cd quipucords
 
 3. Make symbolic link to ansible roles::
+
     ln -s ../roles/ roles
 
 4. Install gunicorn::
+
     pip install gunicorn==19.7.1
 
 5. Start gunicorn::
+
     gunicorn quipucords.wsgi -c ./local_gunicorn.conf.py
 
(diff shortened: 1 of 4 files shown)


quipucords/scanner/manager.py
```diff
@@ -52,29 +64,35 @@ def log_info(self):
             interval = DEFAULT_HEARTBEAT
         heartbeat = Timer(interval, self.log_info)
         current_scan_job = None
-        if self.current_task:
-            current_scan_job = self.current_task.identifier
-        logger.info('Scan Manager Stats: '
-                    'current_scan_job=%s. '
-                    'queue_length=%s.',
-                    current_scan_job, len(self.scan_queue))
+        scan_job_message = None
```
abaiken: Does this variable need to be initialized to None since it is set in either the if or the else?

kholdaway: Removed.

README.rst
```diff
@@ -180,6 +180,36 @@ If you intend to run on Mac OS, there are several more steps that are required.
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+You can run the server locally inside of gunicorn.  This can be a useful way to debug.
+
+1. Clone the repository::
+
+    git clone git@github.com:quipucords/quipucords.git
+
+2. Switch to quipucords directory::
+
+    cd quipucords
```
abaiken: After testing this manually, this line is a little misleading because I feel like I am just supposed to cd to the repo I just cloned, but I actually am going to `quipucords/quipucords`

myersCody: cd /quipucords/quipucords/quipucords/quipucords/quipucords

PR #1452, opened 2018-09-11
Use only DNF for Fedora

kholdaway:


install/roles/docker/tasks/main.yml
```diff
@@ -69,13 +89,22 @@
...
   yum:
     name: libselinux-python
     state: present
   become: true
   when:
-    - is_rhel_centos_6 or is_fedora
+    - is_rhel_centos_6
+    - selinux_on
+
+- name: Install libselinux-python on Fedora
```
abaiken: Small nitpick: some places say`on RHEL 6` some places just have `RHEL6`, `Fedora`, etc. We should probably make these uniform possibly like: ``` - name: Install libselinux-python (Fedora) ```. We can do this at a later time because I'm sure this inconsistency extends beyond your PR.

kholdaway: Let's do this on master.  I think we should normalize all the role names.

kholdaway: See https://github.com/quipucords/quipucords/issues/1453

PR #1459, opened 2018-09-17
Improve date parsing and increase gunicorn timeout

### LINKED ISSUE #1457: Merging job reports causes 'Connection aborted' error inside of docker container
## Bug Description:
When merging 7 reports, the QPC server fails with the following error:
A connection error occurred while attempting to communicate with the server. To troubleshoot this problem, check the configuration and the status of the server.
('Connection aborted.', RemoteDisconnected('Remote end closed connection without response',))

### Expected behavior:
Reports would be merged

### Actual behavior:
'Connection aborted.', RemoteDisconnected('Remote end closed connection without response'

### Steps to reproduce:
Merge several large reports
___

kholdaway:
Closes #1457

### DIFF
--- a/quipucords/fingerprinter/jboss_eap.py
+++ b/quipucords/fingerprinter/jboss_eap.py
@@ -201,8 +201,9 @@ def is_eap_manifest_version(manifest_dict):
                     if classification and 'EAP' in classification:
                         return Product.PRESENT
         else:
-            logger.warning('Expected a dictionary of strings for %s',
-                           EAP_HOME_MODULES_VERSION)
+            logger.warning('Expected a dictionary of strings for %s, '
+                           'but found %s.  Invalid value: %s',
+                           EAP_HOME_MODULES_VERSION, type(manifest), manifest)
     return Product.ABSENT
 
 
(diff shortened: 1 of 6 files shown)


quipucords/fingerprinter/jboss_eap.py
```diff
@@ -218,8 +219,9 @@ def get_eap_manifest_version(manifest_dict):
                     if classification:
                         versions.add(classification)
         else:
-            logger.warning('Expected a dictionary of strings for %s',
-                           EAP_HOME_MODULES_VERSION)
+            logger.warning('Expected a dictionary of strings for %s, '
+                           'but found %s.  Invalid value: %s',
+                           EAP_HOME_MODULES_VERSION, type(manifest), manifest)
```
abaiken: Not related to this PR but we might consider going through this file and improving the logging in the rest of the functions. (There are a lot of log statements that don't say what was found, only what was expected)

abaiken:
Also not a blocker, but you should consider adding tests for the new date formats that are now accepted.

quipucords/fingerprinter/jboss_eap.py
kholdaway: Yeah.  Agree. :(

PR #1461, opened 2018-09-18
Add env to control gunicorn timeout value

kholdaway:


docs/source/configure.rst
```diff
@@ -86,6 +86,22 @@ To view the status of the server after it is running, enter the following comman
 
   # sudo docker ps
```
abaiken: Do we need to provide a way for them to set this in the install script too? That way they can just run the install script and pass it in using `-e QPC_SERVER_TIMEOUT=30` in addition to being able to manually do it starting the docker container?

kholdaway: Done

docs/source/configure.rst
```diff
@@ -86,6 +86,22 @@ To view the status of the server after it is running, enter the following comman
...
 
+Changing Other Quipucords Server Settings
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+You may also configure the quipucords server by setting the following environment variables:
+
+  - QPC_SERVER_TIMEOUT
+    - The default http request timeout.  For example, this can be raised if you are merging a large number of jobs.
```
abaiken: Maybe change jobs to reports?

install/install.sh
```diff
@@ -51,6 +51,8 @@ usage() {
          -e QPC_DBMS_HOST=host
     * If using postgres, specify the db port (if not specified the default value is 5432):
          -e QPC_DBMS_PORT=5432
+    * Override default server timeout for HTTP requests (if not specified the default value is 120):
+         -e QPC_SERVER_TIMEOUT=120
```
abaiken: We should make our instructions consistent with either QPC_SERVER_TIMEOUT or server_timeout

Issue #1412, opened 2018-08-29
User Story: Enhance QPC installer to install a Postgres DB


kholdaway:
## Description:
Epic https://github.com/quipucords/quipucords/issues/1410 focuses on allowing a user to configure QPC to run with a Postgres DB they setup.  This epic will modify the QPC installation script to setup a Postgres DB.  Additionally, the script should make Postgres the default DB. 
___

## Acceptance Criteria:

- [x] Installation script successfully setup a Postgres DB
- [x] The QPC server is configured to use the Postgres DB that was setup
- [x] The default DB is Postgres when running the qpc docker image (dev can still be sqlite)
- [x] A user can specify SQLite via a flag in the installation script
- [x] The user documentation reflects these changes

kholdaway:
Nice job @abaiken and @myersCody

PR #1465, opened 2018-09-19
Fixed issue #1412: Modified the installation scripts to include postgres

myersCody:


myersCody:
- [ ] Update README.rst

- [ ] Edit the make file to include postgres

install/roles/qpc_server/tasks/main.yml
```diff
@@ -22,12 +22,12 @@
 
 - name: Set server url
   set_fact:
-    server_url: "{{ ansible_env['SERVER_URL'] | default('https://github.com/quipucords/quipucords/releases/download/1.0.0/quipucords.1.0.0.tar.gz') }}"
+    server_url: "{{ ansible_env['SERVER_URL'] | default('https://github.com/quipucords/quipucords/releases/download/0.0.44/quipucords.0.0.44.tar.gz') }}"
```
kholdaway: This should be 1.0.0 for master.

install/roles/qpc_server/tasks/main.yml
```diff
@@ -22,12 +22,12 @@
...
   when: server_url is not defined
 
 - name: Set server image and tag name
   set_fact:
-    server_image_and_tag: "{{ ansible_env['SERVER_IMAGE'] | default('quipucords:1.0.0') }}"
+    server_image_and_tag: "{{ ansible_env['SERVER_IMAGE'] | default('quipucords:0.0.44') }}"
```
kholdaway: This should also be 1.0.0 (no change needed here)

install/roles/qpc_server/tasks/main.yml
```diff
@@ -172,16 +172,22 @@
...
-    QPC_DBMS: "{{ ansible_env['QPC_DBMS'] | default('sqlite') }}"
+    QPC_DBMS: "{{ ansible_env['QPC_DBMS'] | default('postgres') }}"
   when:
     - QPC_DBMS is not defined
 
 - name: Set default for use_postgres to False
   set_fact:
     use_postgres: false
 
+- name: Set EXTERNAL_POSTGRES (defaults to false)
```
kholdaway: EXTERNAL_POSTGRES should be described in the usage section of the `install.sh`

kholdaway:
I opened https://github.com/quipucords/quipucords/issues/1470 for the docker volume work and https://github.com/quipucords/quipucords/issues/1471 for the external postgres db work

quipucords/quipucords/settings.py
```diff
@@ -162,13 +162,26 @@ def is_int(value):
 # https://docs.djangoproject.com/en/1.11/ref/settings/#databases
 
 # Database Management System could be 'sqlite' or 'postgresql'
-QPC_DBMS = os.getenv('QPC_DBMS', 'sqlite').lower()
+QPC_DBMS = os.getenv('QPC_DBMS', 'postgres').lower()
 
-if QPC_DBMS == 'postgres':
+if QPC_DBMS == 'sqlite':
+    # If user enters an invalid QPC_DBMS, use default sqlite
```
abaiken: we should remove or move and edit this comment because the default is no longer sqlite

quipucords/quipucords/environment.py
```diff
@@ -115,10 +115,10 @@ def startup():
             logger.info('QPC_DBMS_DATABASE set to "%s"', database)
             logger.info('QPC_DBMS_USER set to "%s"', user)
     elif not qpc_dbms:
-        logger.info('QPC_DBMS not set. Using default of "sqlite".')
+        logger.info('QPC_DBMS not set. Using default of "postgres".')
     else:
         logger.info('QPC_DBMS was set to "%s" which is not a valid option. '
-                    'Using default of "sqlite.',
+                    'Using default of "sqlite".',
```
abaiken: change this to say using the default of `postgres`

install/install.sh
```diff
@@ -41,13 +41,15 @@ usage() {
     * Specify a database management system. Valid options include: postgres, sqlite (if not specified, the default is to use sqlite).
```
abaiken: Change this because the default is no longer to use sqlite

install/install.sh
```diff
@@ -41,13 +41,15 @@ usage() {
...
       If the database management system is specified to use postgres, the following extra variables must also be set: QPC_DBMS_HOST, QPC_DBMS_PASSWORD
```
abaiken: I feel like we should probably reword some of this documentation because everything starts with "If using postgres," . . . and the default is now to use postgres. We need more of a specification between having a preconfigured postgres & creating an internal one.  This is mostly me rambling, but see the comments below.

install/install.sh
```diff
@@ -41,13 +41,15 @@ usage() {
...
          -e QPC_DBMS=sqlite
+    * Specify if you would like to use external preconfigured PostgresSQL.
+         -e EXTERNAL_POSTGRES=true
     * If using postgres, specify the db name (if not specified the default value is postgres):
          -e QPC_DBMS_DATABASE=postgres
     * If using postgres, specify the db user (if not specified the default value is postgres):
          -e QPC_DBMS_USER=postgres
-    * If using postgres, you must specify the db password:
+    * If using postgres, you must specify the db password if EXTERNAL_POSTGRES = true (otherwise default value is password):
```
abaiken: If we have an external postgres set up, they must set the password unless it also happens to be "password". If they are using the internal postgres, do they still have the option to create their own password? If so, it needs to be clear

install/install.sh
```diff
@@ -41,13 +41,15 @@ usage() {
...
          -e QPC_DBMS_PASSWORD=password
-    * If using postgres, you must specify the db host:
+    * If using postgres, you must specify the db host if EXTERNAL_POSTGRES = true (otherwise the default is set to the postgres container):
```
abaiken: Along the same lines as above, what happens if they are using the internal postgres but they set a QPC_DBMS_HOST?  If this will break our flow, maybe it should be clear that they can't set this value if not using an external DB. Conversely, if they are using an external DB and fail to provide the QPC_DBMS_HOST 
but it does not match the default, will it bomb out? Or will it silently not work? What do we want to happen? 

 (This question is also for Kevan not just you. )

abaiken: Perhaps these should be internal passwords/host that don't match the same naming convention as the external variables, so that we don't have to worry about users leaving the defaults for the internal postgres?

ie. When I allowed users to pass in external env variables, if they failed to specify a host or password, it automatically used sqlite to avoid a silent inability to reach the database. << not saying this was a good behavior, but it was in the documentation. We should decide how we want to handle this

install/roles/qpc_server/tasks/main.yml
```diff
@@ -203,6 +203,20 @@
...
+- name: Set QPC_DBMS_HOST
+  set_fact:
+    QPC_DBMS_HOST: "{{ansible_env['QPC_DBMS_HOST'] | default('qpc-db') }}"
+  when:
+    - QPC_DBMS_HOST is not defined
+    - QPC_DBMS == "postgres"
+
 - name: Change value for use_postgres if all env variables are specified
   set_fact:
     use_postgres: true
```
abaiken: maybe take this out, seems irrelevant now

PR #1468, opened 2018-09-20
Improve Dockerfile

elyezer:
Use Fedora 28 as the base image and clean dnf cache to generate smaller
images.


elyezer:
I've run some camayoc tests against the generated image and everything passed.

elyezer:
I will update this PR to just clean some installation cache. Will leave updating the base image for a later time.

PR #1469, opened 2018-09-20
Reverse the order that dependencies are installed (Make dev-requirements install requirements)

abaiken:
Initial changes to upgrading dependencies.


kholdaway:
Looks like CONTRIBUTING.rst has requirements.txt instead of dev-requirements.txt

kholdaway:
Nice one! Looks like a ✊ 🆙 📈

PR #1524, opened 2018-09-25
Edited installation scripts to install Ansible (#1238)

### LINKED ISSUE #1238: Have install.sh check for ansible installation and install it if not present
## Specify type:
- Enhancement

### Priority:
- Medium
___

## Description:
It would be nice if install.sh could check if ansible is installed and install it if not. One pilot customer didn't see the prerequisite to install Ansible so the install.sh script failed. It would be nice to have it fail gracefully if Ansible is not installed or better yet have it install Ansible based on the version of RHEL/Centos that is running.

___

## Acceptance Criteria:

- [ ] Verify that install.sh fails gracefully if Ansible is not installed

___

myersCody:
Closes #1238


install/install.sh
```diff
@@ -53,11 +53,43 @@ then
   usage;
 fi
 
+if grep -q -i "release 6" /etc/redhat-release || grep -q -i "release 7" /etc/redhat-release
```
elyezer: You can simplify this logic, something like:

```sh
if dnf --version; then
    PKG_MGR=dnf
else
    PKG_MGR=yum
fi
```

Later in the code you can just call:

```sh
sudo "${PKG_MGR}" install -y ansible
```

install/install.sh
```diff
@@ -53,11 +53,43 @@ then
...
+then
+  install_with_yum=1
+elif grep -q -i "release 7" /etc/centos-release || grep -q -i "release 6" /etc/centos-release
+then
+  install_with_yum=1
+elif grep -q -i "release 28" /etc/fedora-release || grep -q -i "release 27" /etc/fedora-release
+then
+  install_with_dnf=1
+else
+  echo "Installation failed. Could not find a supported Operating System."
```
elyezer: If you change the package manager logic you can check if this is a redhat OS by running:

```sh

if [ ! -f /etc/redhat-release ]; then
	echo "/etc/redhat-release not found. You need to run this on a Red Hat based OS"
fi

install/install.sh
elyezer: Or you could use `command -v dnf` like you did below for ansible

mdvickst:
Does this enable to required repos? I think they are different for RHEL 6 versus RHEL 7. There are also now Ansible Engine Repositories (e.g. rhel-7-server-ansible-2.5-rpms) which could help us lock on a specific major version of Ansible, would there be any advantage to downloading from there?

PR #1539, opened 2018-09-29
Add QPC CLI bulk merge option (#1535)

### LINKED ISSUE #1535: Add QPC CLI bulk merge option
## Feature description:
Add QPC CLI command that takes a directory and merges the details report into a single deployment report.  Command should use new `/api/reports/merge/jobs/` API

Command:
`qpc report merge --json-directory path`.

CLI should return the job id that can be used to check status using `qpc report job --id JOB_ID` (see #1536 )

myersCody:
Closes #1535

### DIFF
--- a/qpc/messages.py
+++ b/qpc/messages.py
@@ -170,7 +170,7 @@ REPORT_INVALID_JSON_FILE = 'The file %s does not contain a valid json ' \
 REPORT_JSON_DIR_NO_FILES = \
     'No files with extension .json found in the provided path.'
 REPORT_JSON_DIR_CHECK_FILES = \
-    'Checking files for the correct structure. %s'
+    'Checking files for valid json details report. %s'
 REPORT_JSON_DIR_FILE_FAILED = 'Failed: %s is not a details report.'
 REPORT_JSON_DIR_FILE_SUCCESS = 'Success: %s is a details report.'
 REPORT_JSON_DIR_ALL_FAIL = 'No details reports were found.'
--- a/qpc/messages.py
+++ b/qpc/messages.py
@@ -168,9 +168,8 @@ REPORT_JSON_FILES_HELP = 'At least two json details report files are ' \
 REPORT_INVALID_JSON_FILE = 'The file %s does not contain a valid json ' \
                            'details report.'
 REPORT_JSON_DIR_NO_FILES = \
-    'No files with extension .json found in the provided path.'
-REPORT_JSON_DIR_CHECK_FILES = \
-    'Checking files for valid json details report. %s'
+    'No files with extension .json found in %s.'
+REPORT_VALIDATE_JSON = 'Checking files for valid json details report. %s'
 REPORT_JSON_DIR_FILE_FAILED = 'Failed: %s is not a details report.'
 REPORT_JSON_DIR_FILE_SUCCESS = 'Success: %s is a details report.'
 REPORT_JSON_DIR_ALL_FAIL = 'No details reports were found.'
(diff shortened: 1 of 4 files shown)


qpc/messages.py
```diff
@@ -161,10 +161,21 @@
...
 REPORT_JSON_FILES_HELP = 'At least two json details report files are ' \
                          'required to merge.'
 REPORT_INVALID_JSON_FILE = 'The file %s does not contain a valid json ' \
                            'details report.'
+REPORT_JSON_DIR_NO_FILES = \
+    'No files with extension .json found in the provided path.'
+REPORT_JSON_DIR_CHECK_FILES = \
+    'Checking files for the correct structure. %s'
+REPORT_JSON_DIR_FILE_FAILED = \
+    'FAILED: (%s) failed to have the correct structue.'
```
kholdaway: Let's just say (%s) is not a details report.

qpc/messages.py
```diff
@@ -161,10 +161,21 @@
...
+REPORT_JSON_DIR_FILE_SUCCESS = \
+    'Success: (%s) structure passed.'
```
kholdaway: Use details report instead of structure.

qpc/messages.py
```diff
@@ -161,10 +161,21 @@
...
+REPORT_JSON_DIR_ALL_FAIL = 'All json files failed to have the right structure'
```
kholdaway: replace with `No details reports were found` or something like that.

qpc/report/merge.py
```diff
@@ -135,6 +138,53 @@ def _merge_json(self):
...
+
+        for file in json_files:
+            with open(file) as lint_f:
+                error = False
+                json_data = json.load(lint_f)
+                try:
+                    # pylint: disable=pointless-statement
+                    json_data['sources']
+                    # pylint: disable=pointless-statement
+                    json_data['sources'][0]['facts']
```
kholdaway: You will need to check lengths or you could have an index out of bounds. For example, sources could equal `[]` and index 0 will fail.  I actually would just do this:
```
sources = json_data.get('sources')
if sources:
    facts = sources[0].get('facts)
    if facts:
          facts.get('server_id')
```

This avoids KeyErrors being thrown so you can remove the try/catch.  The checks for sources and facts screen for both None/empty.

qpc/messages.py
```diff
@@ -161,10 +161,19 @@
...
+REPORT_JSON_DIR_HELP = 'The path to a directory that contain files of json ' \
+                       'details reports to merge'
 REPORT_JSON_FILES_HELP = 'At least two json details report files are ' \
                          'required to merge.'
 REPORT_INVALID_JSON_FILE = 'The file %s does not contain a valid json ' \
                            'details report.'
+REPORT_JSON_DIR_NO_FILES = \
+    'No files with extension .json found in the provided path.'
+REPORT_JSON_DIR_CHECK_FILES = \
+    'Checking files for the correct structure. %s'
```
abaiken: Change this to say something like `Checking files for valid json details report.`

kholdaway:
Add this command to the man page.

qpc/report/merge.py
```diff
@@ -135,6 +144,54 @@ def _merge_json(self):
...
+                sources = json_data.get('sources')
+                if sources:
+                    facts = sources[0].get('facts')
+                    server_id = sources[0].get('server_id')
+                    if facts and server_id:
+                        error = False
+            if error is not True:
+                print(_(messages.REPORT_JSON_DIR_FILE_SUCCESS % file))
+                try:
+                    report_id, sources = self.get_id_and_sources(file)
```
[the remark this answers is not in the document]
myersCody: I split the logic into a separate method called ```_validate_create_json()```

PR #1551, opened 2018-10-03
Lock Ansible version

abaiken:


requirements.txt
```diff
@@ -1,43 +1,44 @@
-ansible==2.3.2.0
+# Include Locked Server Requirements
+-r locked-requirements.txt
+--editable .
```
kholdaway: I don't think this is required.

PR #1559, opened 2018-10-03
fingerprint incensitive compare for mac addr (#1548)

### LINKED ISSUE #1548: Fingerprint engine should use case insensitive compare for mac_address values
## Bug Description:
System 1: [00:50:56:9E:6E:05]
System 2: [00:50:56:9e:6e:05]

These don't match today.  They should.

### Expected behavior:
Match mac address regardless of case.

### Actual behavior:
Doesn't

myersCody:
Closes #1548


quipucords/fingerprinter/tests_fingerprint_task.py
```diff
@@ -559,6 +559,39 @@ def test_merge_network_and_vcenter_infrastructure_type(self):
                 self.assertEqual(result_fingerprint.get(
                     'infrastructure_type'), 'virtualized')
 
+    def test_merge_mac_address_case_insensitive(self):
+        """Test if fingerprints will be merged with mixed mac addr."""
+        n_mac = ['00:50:56:A3:A2:E8']
```
kholdaway: Add a test case with more than one mac address.

quipucords/fingerprinter/task.py
```diff
@@ -643,6 +643,8 @@ def _add_fact_to_fingerprint(self, source,
             actual_fact_value = fact_value
         elif raw_fact.get(raw_fact_key) is not None:
             actual_fact_value = raw_fact.get(raw_fact_key)
+            if fingerprint_key == 'mac_addresses':
+                actual_fact_value[0] = actual_fact_value[0].lower()
```
kholdaway: That only set's the first mac address to lower-case.  Need to iterate them all.

PR #1568, opened 2018-10-04
Move pycrypto to locked dependencies 

abaiken:


kholdaway:
Builds failed 😬

locked-requirements.txt
```diff
@@ -1,5 +1,9 @@
 ansible==2.3.2.0
+pycrypto==2.6.1
+
+# locking Django until we separate CLI
 Django==2.0.0
 
+# flake8 locks the following dependencies
```
elyezer: Not related to this PR but I would assume the locked requirements being something needed to deploy the server. That said I would suggest to avoid adding dev related requirements when installing production requirements.

abaiken: @elyezer I created a separate `dev-locked-requirements.txt` for the locked pyflakes dependency! Great suggestion 👍

Issue #1520, opened 2018-09-25
Spike: UI Build package updates and tests


cdcabrera:
### Is your feature request related to a problem?
Nope

### Describe the solution you'd like
UI build packages are currently out date. Upgrade them, along with syntax and testing updates. The end result of these updates should not affect using the UI.
___

## Acceptance Criteria:

- [x] Verify that the Quipucords UI build and UI linters compile and pass unit testing
- [x] Verify that users can still perform all UI related actions, and no ui console errors exist

### Activate lint rules round 2
- [x] arrow-body-style
- [x] class-methods-use-this
- [x] consistent-return
- [x] func-names
- [x] import: no-extraneous-dependencies, no-useless-path-segments, order, prefer-default-export
- [x] no-return-assign, no-unused-expressions, no-unused-vars
- [x] object-shorthand
- [x] prefer-arrow-callback, prefer-destructuring, prefer-const, prefer-rest-params
- [x] react/button-has-type
- [x] react/default-props-match-prop-types

### Activate lint rules round 3
- [x] jsx-a11y/label-has-associated-control
- [x] no-param-reassign
- [x] no-shadow
- [x] react/destructuring-assignment
- [x] react/jsx-curly-brace-presence
- [x] react/jsx-wrap-multilines
- [x] react/no-access-state-in-setstate
- [x] react/no-array-index-key
- [x] react/no-unused-state
- [x] react/require-default-props
- [x] react/sort-comp

cdcabrera:
We recognize there may be a possible limitation in the React framework we can upgrade towards.

The next minor version update for React requires us to either prefix with `UNSAFE_` or alter all the React life cycle methods for `componentWillReceiveProps`.

At the moment the plan is to stay on the current version of React and update all the lint checks, then upgrade the components as new features come along. Once that's complete we can upgrade the version.

cdcabrera:
The immediate linting and clean up updates for the GUI are in, marking this issue resolved.

We had some quirkiness around adding React default props. Had a couple of client side errors pop up, we fixed most of them. However we were unable to recreate one of them since it popped up and immediately disappeared on refresh. We'll keep an eye on it, and can look at rolling the change back if it becomes an issue on the GUI.

PR #1570, opened 2018-10-05
UI: Build package updates and tests

cdcabrera:
### What's included

fix(ui build): ui build packages, linter updates
* build package dependencies updated
* linter updates from active eslint rules
* testing snapshots updated
* helper bind comment

### Notes
* This is an incremental client build update avoiding UI behavior changes
* Yarn lock added for future updates
* Linter rules were enabled then the majority disabled. The intent is to update incrementally, relaxing the rules then correcting any related issues, see these disabled rules: 
   - https://github.com/cdcabrera/quipucords/blob/issues/1520/client/.eslintrc#L61-L89

### Updates issue
updates #1520


client/src/components/credentials/credentialListItem.js
```diff
@@ -112,7 +112,7 @@ class CredentialListItem extends React.Component {
     return [
       <ListView.InfoItem
         key="sources"
-        className={'list-view-info-item-icon-count ' + (sourceCount === 0 ? 'invisible' : '')}
+        className={`list-view-info-item-icon-count ${sourceCount === 0 ? 'invisible' : ''}`}
```
jeff-phillips-18: Better would be to use the `classNames` util.

client/src/components/addSourceWizard/addSourceWizard.js
```diff
@@ -226,4 +229,7 @@ const mapStateToProps = function(state) {
   return { ...state.addSourceWizard.view };
 };
 
-export default connect(mapStateToProps, mapDispatchToProps)(AddSourceWizard);
+export default connect(
+  mapStateToProps,
+  mapDispatchToProps
+)(AddSourceWizard);
```
jeff-phillips-18: Is this due to line length lint errors? Doesn't seem to be too long.

cdcabrera: Yeah, it looks like it's an aspect of `prettier` having to do with function calls... I think this is related:
- https://github.com/prettier/prettier/issues/1420

Issue #528, opened 2018-01-28
Externalize error messages for logs


kholdaway:
# Opening a new issue
___

## Specify type:
- Enhancement

## Description:
Many strings are not externalized.  Make sure they are for beta.  This is part of the fit and finish.
___

justinkreft:
closed. stale and addressed

PR #1601, opened 2018-10-11
Updates to ensure partial results are recorded for multiple source scans

### LINKED ISSUE #1600: A failing scan task causes report to fail
## Bug Description:
Currently, if you have 5 sources and 1 fails the connect and inspect both run.  The failed task for inspect is skipped.  However, we waste time running the other 4 tasks and do not report the result.  I think this is a regression.

### Expected behavior:
4 successful inspect tasks should report results.  Scan Job should indicate failure.

### Actual behavior:
Currently, if you have 5 sources and 1 fails the connect and inspect both run.  The failed task for inspect is skipped.  However, we waste time running the other 4 tasks and do not report the result.  I think this is a regression.

### Steps to reproduce:
List clear steps to reproduce the bug.

### LINKED ISSUE #1602: Satellite 5 scan appears to be hang for a long period of time at the beginning of inspection
## Bug Description:
Mysterious hang when inspection starts on satellite 5.  Debugging reveals that the scan is gathering physical hosts, virtual hosts, and virtual guests.  Solution is to add logging so scan does not appear hung.

### Expected behavior:
Indication of what is happening

### Actual behavior:
No progress reported.

### Steps to reproduce:
Run satellite 5 scan

kholdaway:
Closes #1600 
Closes #1602

### DIFF
--- a/quipucords/scanner/network/inspect.py
+++ b/quipucords/scanner/network/inspect.py
@@ -110,8 +110,8 @@ class InspectTaskRunner(ScanTaskRunner):
 
         self.connect_scan_task = self.scan_task.prerequisites.first()
         if self.connect_scan_task.status != ScanTask.COMPLETED:
-            error_message = 'Prerequisites scan task with id %d failed.' %\
-                self.connect_scan_task.id
+            error_message = 'Prerequisites scan task %d failed.' %\
+                self.connect_scan_task.sequence_number
             return error_message, ScanTask.FAILED
 
         try:
--- a/.gitignore
+++ b/.gitignore
@@ -112,6 +112,7 @@ ENV/
 
 # sqlite
 db.sqlite3
+db.sqlite3-journal
 
 # From ide
 .idea/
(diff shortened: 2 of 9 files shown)


quipucords/scanner/network/inspect.py
```diff
@@ -110,8 +110,8 @@ def run(self, manager_interrupt):
 
         self.connect_scan_task = self.scan_task.prerequisites.first()
         if self.connect_scan_task.status != ScanTask.COMPLETED:
-            error_message = 'Prerequisites scan task with id %d failed.' %\
-                self.connect_scan_task.id
+            error_message = 'Prerequisites scan task %d failed.' %\
```
abaiken: We should consider adding this error message to our messages file since it is used in 3 different files.

kholdaway: Great suggestion. Let's do this as part of https://github.com/quipucords/quipucords/issues/528

PR #1612, opened 2018-10-16
Fix URI to perform merge (uses sync)

### LINKED ISSUE #1606: Can't merge reports by report id or scan-job id
## Bug Description:
Trying to merge a report by passing either `--job-ids` or `--report-ids` is not working anymore.

Trying to pass either of the options the following message is returned:

```
No reports found.  Error json: 
{'detail': 'Method "PUT" not allowed.'}
```

Marking this as a blocker because this issue is a regression.

### Expected behavior:
The merge report should work as it used to work for all possible options.

### Actual behavior:
An error message is being raised and the merge does not happen.

### Steps to reproduce:
1) Perform two scans and store their job-ids
2) Generate a report for both scans and store their report-ids
3) Try to create a merge report by job-ids and by report-ids. Both should fail.

The commands should be something like:

```console
$ qpc report merge --report-ids 1 2    
No reports found.  Error json: 
{'detail': 'Method "PUT" not allowed.'}

$ qpc report merge --job-ids 2 3
No reports found.  Error json: 
{'detail': 'Method "PUT" not allowed.'}
```

___

## Acceptance Criteria:

- [ ] Verify that a merge report can be created by passing scan job ids
- [ ] Verify that a merge report can be created by passing report ids

___

### Version Information:
4f1aabfa179217f4daac552cbec9441b47dd2966

kholdaway:


qpc/report/tests_report_merge.py
```diff
@@ -111,8 +111,8 @@ def test_detail_merge_job_ids(self):
                              json_dir=None)
             with redirect_stdout(report_out):
                 nac.main(args)
-                self.assertEqual(report_out.getvalue().strip(),
-                                 messages.REPORT_SUCCESSFULLY_MERGED % '1')
+                self.assertIn(messages.REPORT_SUCCESSFULLY_MERGED % ('1', '1'),
```
abaiken: can we not assert that these are equal instead of asserting that it is in the report_out.getvalue()?

qpc/report/tests_report_merge_status.py
```diff
@@ -67,8 +67,8 @@ def test_job_valid_id(self):
             args = Namespace(job_id='1')
             with redirect_stdout(report_out):
                 nac.main(args)
-                self.assertIn(messages.JOB_ID_STATUS % ('1',
-                                                        'complete'),
+                self.assertIn(messages.MERGE_JOB_ID_STATUS % ('1',
```
abaiken: same question here as above

qpc/report/tests_report_merge.py
kholdaway: Fixed.

qpc/report/tests_report_merge_status.py
kholdaway: The reason for this is that there are multiple lines of output.  I would argue that this test is fine the way it is but I did update it.

abaiken:
Test coverage dropped. Possibly add a test-case for the new PUT url

PR #1616, opened 2018-10-17
Add 32 as a valid subnet mask

dozdowski:
For the test applied to each line's entry in source creation, the previous regexp allowed the "larger" subnet masks as well as 30 and 31.  Technically, 32 is also a valid subnet mask that we should honor.  I have added it to the regexp in question.


cdcabrera:
Looks good @dozdowski I'll squash it down once it passes checks. Went ahead and brought your branch forward with master

PR #1635, opened 2018-10-26
Update install script

elyezer:
Look for the master release and exit with a non-zero return code when
ansible-playbook fail.

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -13,7 +13,7 @@
 
 export PATH=$PATH:$ANSIBLE_HOME/bin
 PLAYBOOKFILE="qpc_playbook.yml"
-RELEASE_TAG='-e RELEASE_TAG=1.0.0'
+RELEASE_TAG='-e RELEASE_TAG=master'
 POSTGRES_VERSION='-e POSTGRES_VERSION=9.6.10'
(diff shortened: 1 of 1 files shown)


install/install.sh
```diff
@@ -13,7 +13,7 @@
 
 export PATH=$PATH:$ANSIBLE_HOME/bin
 PLAYBOOKFILE="qpc_playbook.yml"
-RELEASE_TAG='-e RELEASE_TAG=1.0.0'
+RELEASE_TAG='-e RELEASE_TAG=master'
```
elyezer: This change is to mach the version the master pipeline is creating, see https://github.com/quipucords/quipucords/blob/master/Jenkinsfile#L1

Better to name it master than 1.0.0 to not confuse with releases instead of the dev version

install/install.sh
```diff
@@ -112,5 +112,5 @@ then
   echo "Installation complete."
 else
   echo "Installation failed. Review the install logs."
-  exit $?
+  exit 1
```
elyezer: The `$?` will always be 0 because it will be the exit code of the `echo` command above. This will make the script fail whenever a failure is raised when running ansible-playbook.

PR #1650, opened 2018-10-31
Make life easier for developers debugging container

abaiken:
- add instructions to readme 
- install ps in container
-add volume for /app

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -190,7 +190,7 @@ You must have `Docker installed <https://docs.docker.com/engine/installation/>`_
 
   **NOTE:** The need to use ``sudo`` for this step is dependent upon on your system configuration.
 
-4. There are 3 different options for running the QPC server.
+4. There are many different options for running the QPC server.
     A. Run the Docker image with Postgres container::
 
         docker run --name qpc-db -e POSTGRES_PASSWORD=password -d postgres:9.6.10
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -205,13 +205,26 @@ You must have `Docker installed <https://docs.docker.com/engine/installation/>`_
 
         docker run -d --name quipucords -e "QPC_DBMS=sqlite" -p443:443 -i quipucords:1.0.0
 
+    D. For debugging purposes you may want to run the Docker image with the /app directory mapped to your local clone of quipucords and the logs mapped to a temporary directory::
+
+        docker run -d --name quipucords -e "QPC_DBMS=sqlite" -p443:443 -v /path/to/local/quipucords/:/app -v /tmp:/var/log -i quipucords:1.0.0
```
elyezer: :heart:

myersCody: ❤️

README.rst
```diff
@@ -190,7 +190,7 @@ You must have `Docker installed <https://docs.docker.com/engine/installation/>`_
 
   **NOTE:** The need to use ``sudo`` for this step is dependent upon on your system configuration.
 
-4. There are 3 different options for running the QPC server.
+4. There are many different options for running the QPC server.
```
myersCody: If you view this, you will see that number 4 is in bold and italics. Makes it seem a little awkward compared to the rest of the doc.

kholdaway: Agree.

README.rst
kholdaway: This maps the application code not logs.  Let's change the labels from `1.0.0` to `master`.  @elyezer started us down this path with Jenkins.  Really 1.0.0 doesn't make a whole lot of sense anymore.  

I also don't know if mapping the logs is that important since you have to docker exec to restart gunicorn anyway.  We may want to just focus on mapping the app source code.  The developer will be in the docker container already so tailing the logs doesn't seem difficult.

README.rst
```diff
@@ -205,13 +205,26 @@ You must have `Docker installed <https://docs.docker.com/engine/installation/>`_
 
         docker run -d --name quipucords -e "QPC_DBMS=sqlite" -p443:443 -i quipucords:1.0.0
 
+    D. For debugging purposes you may want to run the Docker image with the /app directory mapped to your local clone of quipucords and the logs mapped to a temporary directory::
```
kholdaway: I would clarify the purpose a bit.  Viewing the logs isn't that difficult.  Rebuilding the container is painful.  Clarify that the purpose is to allow you to rapidly change server code without rebuilding the container.

README.rst
```diff
@@ -205,13 +205,26 @@ You must have `Docker installed <https://docs.docker.com/engine/installation/>`_
...
 
 6.  You can work with the APIs, the CLI, and UI (visit `<https://127.0.0.1/>`_ if you installed the UI in step 2 above).
 
+7. To enter the container use the following command::
+
+    docker exec -it quipucords bash
+
+8. If you need to restart the server inside of the container, run the following after entering the container to get the server PIDs and restart::
+
+    ps -ef | grep gunicorn
```
kholdaway: Additionally, there were usually more than 1 PIDs.  I would make the example show that so people are not confused.

README.rst
elyezer: :+1: for changing from `1.0.0` to `master`

abaiken: This does map the logs @kevan with ```-v /tmp:/var/log```. I agree with @elyezer that the this can be useful. Even though it is not difficult to docker exec into the container to look at the logs, I found it easier to just have it tailing the logs on my host filesystem so that when I was inside the container, it was solely to restart the server.

README.rst
abaiken: Super weird. It has been this way in master 😂 🤔

abaiken: Fixed

README.rst
abaiken: see the note that I added and let me know if you think it is sufficient

PR #1665, opened 2018-11-04
Upgrade django to 2.1.3 to remove security issues

kholdaway:


elyezer:
If I understood correctly requirements are ones that we should be updating as new fixes and security fixes are released, right? Then locked requirements are the one that we need to stay on a specific release because compatibility or the need to stay on an specific release, right?

kholdaway:
Correct. So I moved django out.  Additionally, I used pip-check-reqs to located pip modules that we don't need.

After rooming all the extra pip dependencies, I ran a full scan for network, satellite 5, 61, 62, and vcenter.  Additionally, I ran with default, sudo, root (key/password access).  I also ran with postgres and sqlite.

elyezer:
@kholdaway excellent, I think we are good to merge this then

PR #1686, opened 2018-11-07
Added troubleshooting for backslashes in cli inputs. (#1649)

### LINKED ISSUE #1649: ActiveDirectory login fails when connecting to vCenter
## Bug Description:

When Active Directory is used to connect into vCenter and the service account that is used to login into the vCenter client is "Domain\SVC_satadmin". I am finding that the credential fails when running a scan with QPC.

When creating a credential with QPC using the domain user login "Domain\SVC_admin" with a fwd slash. QPC does not save this slash when it stores it in the cred. The fwd slash is missing when verifying the cred in QPC - "DomainSVC_satadmin". When adding 2 fwd slashes, both slashes are saved within the cred. "Domain\\SVC_satadmin"

### Expected behavior:

QPC should authenticate with vCenter as long as the credential is valid when logging into the vCenter client

### Actual behavior:

QPC provides an error stating that the credential is invalid when reviewing the scan job

### Steps to reproduce:

Create a credential
qpc cred add --type vcenter --name vcred1 --username Domain\SVC_satadmin --password

View credential 

qpc cred list
___

## Acceptance Criteria:

- [ ] Users can create a credential, create a source, create a scan, run the scan with QPC
- [ ] User cannot authenticate with vCenter using the provided credentials, preventing the scan from completing.

___

### Version Information:
CLI version, server version or git commit information.

version 0.0.44

### Environment information:
Provide detailed information on the *scanning* and *target* systems (if applicable); see example below.

Scanning System | RHEL             | 7.4        | 3.6            | Bare Metal
Target System   | RHEL             | 5.9        | 2.4.6          | RHEV

myersCody:
Closes #1649


docs/source/troubleshooting.rst
```diff
@@ -45,3 +45,9 @@ Removing the volume for RHEL 6 or Centos 6
 Removing the volume for RHEL 7, Centos 7, Fedora 27, or Fedora 28
 
 # docker volume rm qpc-data
+
+Adding Active Directory Credentials
```
kholdaway: Let's not reference Active Directory.  There could be other reasons to have a `\`.

How about?
- CLI input with `\`

docs/source/troubleshooting.rst
```diff
@@ -45,3 +45,9 @@ Removing the volume for RHEL 6 or Centos 6
...
+-----------------------------------
+Active directory accounts may structure their usernames to include the domain separated by a backslash e.g. ``Domain\Username``. However, this structure will result in issues due to the backslash being used as an escape character in both Shell and Python. In order to add a username with a backslash present, the backslash will need to be escaped with another backslash ``\\``::
```
kholdaway: Again this could apply to any value (not just username).  We could username and Active directory as a concrete example.

Modify this so it addresses the general problem then ends with a concrete example containing ActiveDirectory and username.

PR #1698, opened 2018-11-12
Added insights client id to network scan (#1670)

### LINKED ISSUE #1670: User Story:  Gather Insights client id
## Feature description:
Add the insights client id to the QPC deployments report.

This story should do the following:
1. Check for the presence of the insights client
2. If the insights client is present, collect the insights client id for the system
3. Add the `insights_client_id` fact to the QPC deployments report
4. Use the insights client id during network system deduplication

```
/etc/insights-client/machine_id
/etc/redhat-access-insights/machine_id
```

myersCody:
Closes #1670

### DIFF
--- a/roles/insights/tasks/main.yml
+++ b/roles/insights/tasks/main.yml
@@ -1,0 +1,15 @@
+---
+
+- name: internal_host_started_processing_role
+  set_fact:
+    internal_host_started_processing_role: "insights"
+
+- name: gather insights.machine-id
+  raw: if [ -f /etc/insights-client/machine-id ] ; then cat /etc/insights-client/machine-id; fi || if [ -f /etc/redhat-access-insights ] ; then cat /etc/redhat-access-insights/machine-id; fi
+  register: internal_cat_insights_machine_id
+  ignore_errors: yes
+
+- name: set instnum
+  set_fact:
+    insights_machine_id: "{{ internal_cat_insights_machine_id.get('stdout') | default('') }}"
+  ignore_errors: yes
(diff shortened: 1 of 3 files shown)


roles/insights/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
+---
+
+- name: internal_host_started_processing_role
+  set_fact:
+    internal_host_started_processing_role: "insights"
+
+- name: gather insights.machine-id
+  raw: if [ -f /etc/insights-client/machine-id ] ; then cat /etc/insights-client/machine-id; fi || if [ -f /etc/redhat-access-insights ] ; then cat /etc/redhat-access-insights/machine-id; fi
+  register: internal_cat_insights_client_id
```
abaiken: I would change this to `internal_get_insights_client_id`

roles/insights/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
...
+  ignore_errors: yes
+
+- name: set instnum
```
abaiken: change this name to `extract insights client id from internal_get_insights_client_id`

roles/insights/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
...
+  set_fact:
+    insights_client_id: "{{ internal_cat_insights_client_id.get('stdout') | default('') }}"
+  ignore_errors: yes
+  when: '"stdout" in internal_cat_insights_client_id'
```
abaiken: you can choose if you want this when (with no default) or if you want to just have the default, they will both do the same thing.

PR #1699, opened 2018-11-13
Collect etc/machine-id (#1683)

### LINKED ISSUE #1683: Collect RHEL machine id identifier
## Feature description:

Currently QPC uses the /etc/machine-id to compute a date, but we could also get the machine id.

myersCody:
Closes #1683


kholdaway:
Builds are failing also.

quipucords/api/deployments_report/model.py
```diff
@@ -137,6 +137,7 @@ class SystemFingerprint(models.Model):
 
     metadata = models.TextField(unique=False, null=False)
     sources = models.TextField(unique=False, null=False)
+    etc_machine_id = models.TextField(unique=False, null=True)
```
kholdaway: Add this to the serializer

roles/etc_release/tasks/main.yml
```diff
@@ -129,3 +129,14 @@
     - internal_uname_version is defined
     - '"stdout_lines" in internal_uname_version'
     - 'internal_uname_version["stdout_lines"]|length > 0'
+
+# gather RHEL machine id
```
kholdaway: Should this just be machine id instead of RHEL machine id?

kholdaway:
@myersCody Run a scan with root, sudo, and basic.  Post the output of the compare tool in the PR.  I uploaded a new version of the compare tool to google drive.

PR #1708, opened 2018-11-20
Add API to request qpc reports as JSON/CSV and return as tar.gz

### LINKED ISSUE #1705: Add API to request qpc reports as JSON/CSV and return as tar.gz
## Feature description:

Add the API endpoint `GET` and path `/api/v1/reports/{id}` with accept header `application/gzip`

Requests for accept of `application/json` and `text/csv` should return a 404.

This endpoint should create a directory called `reports` in the tar.gz.  The deployments report should be in the reports directory as a JSON file and CSV file.  The details report should be in the reports directory as a JSON file and CSV file.  The API should return a tar.gz containing all 4 files.

If the report ID is not found return a 404.

myersCody:
Closes #1705


quipucords/api/messages.py
```diff
@@ -195,3 +195,6 @@
     'the invalid filter key %s.'
 REPORT_INVALID_FILTER_QUERY_PARAM = 'One or more filter keys are not '\
     'attributes of the system fingerprint: %s.'
+REPORTS_MISSING_REQUIRED_VALUE = 'A required value for Reports was missing.'
```
abaiken: I feel like this error message is vague. This is printed if a `report_id` and `details.json` and `deployments.json` are all `None`. It seems like the report does not exist at that point but I'm not sure?

quipucords/api/messages.py
```diff
@@ -195,3 +195,6 @@
...
+REPORTS_NO_CSV = 'A CSV file could not be created.'
```
abaiken: Same comment as above - except this occurs if the `deployments.csv` and `details.csv` are both `None`. Does the report exist/ can we give any more info as to why it cannot be created?

quipucords/api/reports/reports_gzip_renderer.py
```diff
@@ -0,0 +1,80 @@
...
+        if not bool(reports_dict):
+            return None
+
+        files_data = dict()
+
+        report_id = reports_dict.get('report_id')
+        # Collect Json Data
+        details_json = reports_dict.get('details_json')
+        deployments_json = reports_dict.get('deployments_json')
+        if all(value is None for value in [report_id,
```
abaiken: i think this should be an `any` but you might want to double check

quipucords/api/reports/reports_gzip_renderer.py
```diff
@@ -0,0 +1,80 @@
...
+            logger.error(messages.REPORTS_MISSING_REQUIRED_VALUE)
+            return None
+        details_name = create_filename('details', 'json', report_id)
+        files_data[details_name] = details_json
+        deployments_name = create_filename('deployments', 'json', report_id)
+        files_data[deployments_name] = deployments_json
+        # Collect CSV Data
+        details_csv = create_details_csv(details_json)
+        deployments_csv = create_deployments_csv(deployments_json)
+        if all(value is None for value in [details_csv, deployments_json]):
```
abaiken: same comment as above

quipucords/api/common/report_json_gzip_renderer.py
```diff
@@ -37,8 +47,13 @@ def render(self, report_dict, media_type=None, renderer_context=None):
             return None
 
         report_id = report_dict.get('report_id')
+        report_type = report_dict.get('report_type')
         if report_id is None:
             return None
-
-        tar_buffer = create_tar_buffer([report_dict])
+        if all(value is None for value in [report_id, report_type]):
```
abaiken: you possibly want any instead of all

quipucords/api/common/report_csv_renderer.py
```diff
@@ -0,0 +1,37 @@
...
+#
+"""CSV renderer for reports."""
+
+from api.deployments_report.util import create_deployments_csv
+from api.details_report.util import create_details_csv
+
+from rest_framework import renderers
+
+
+class ReportCSVRenderer(renderers.BaseRenderer):
```
kholdaway: Let's split this back out into separate classes.  The utils are really the key for the new API.

PR #1730, opened 2018-12-20
WIP: Store product version json as a textfield and wrap saving fingerprints in a try/catch

abaiken:
1. Wrap saving fingerprints in try/catch
2. Change version from charfield to textfield 
3. Improve logging for failed fingerprints 
4. Fix roles to test if locate is working without giving stderror


quipucords/api/migrations/0004_auto_20181219_1354.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.1.3 on 2018-12-19 13:54
```
kholdaway: Let's name this 0004_product_version_fix.py

roles/check_dependencies/tasks/main.yml
```diff
@@ -104,10 +104,21 @@
   register: internal_have_locate_cmd
   ignore_errors: yes
 
+- name: test locate command
```
kholdaway: Let's add a `when: internal_have_locate_cmd`.  No need to test it when it is not there

quipucords/fingerprinter/task.py
```diff
@@ -235,15 +237,23 @@ def _process_details_report(self, manager_interrupt, details_report):
...
+                        if fingerprint_dict.get(field, None):
+                            fingerprint_dict[field] = \
+                                date_field.to_representation(
+                                    fingerprint_dict.get(field))
+                    final_fingerprint_list.append(fingerprint_dict)
+                    number_valid += 1
+                except DataError as error:
+                    number_invalid += 1
+                    self.scan_task.log_message(
+                        'The following fingerprint failed with error "%s": %s'
```
kholdaway: How about this?  `The fingerprint could not be saved.  Fingerprint: %s.  Error: %s`

PR #1745, opened 2019-01-10
Fixes cpu_socket_count containing stderr

### LINKED ISSUE #1732: cpu_socket_count contains stderr
## Bug Description:
'cpu_socket_count': '/sys/firmware/dmi/tables/smbios_entry_point: Permission denied\r\n/dev/mem: Permission denied\r\n0',

abaiken:
Closes #1732

### DIFF
--- a/quipucords/scanner/network/processing/cpu.py
+++ b/quipucords/scanner/network/processing/cpu.py
@@ -74,3 +74,80 @@ class ProcessCpuBogomips(process.Processor):
     def process(output, dependencies=None):
         """Pass the output back through."""
         return get_line(output['stdout_lines'])
+
+
+class ProcessCpuSocketCount(process.Processor):
+    """Process the cpu socket count."""
+
+    KEY = 'cpu_socket_count'
+    DEPS = ['internal_cpu_socket_count_dmi_cmd',
+            'internal_cpu_socket_count_cpuinfo_cmd',
+            'cpu_count']
+    REQUIRE_DEPS = False
+
+    @staticmethod
+    def process(output, dependencies):
+        """Return the correct value for cpu_socket_count output."""
+        # process the internal dmi cpu socket count result
+        dmi_cpu_socket_count = \
+            dependencies.get('internal_cpu_socket_count_dmi_cmd')
+        if dmi_cpu_socket_count and dmi_cpu_socket_count.get('rc') == 0:
+            dmi_count = dmi_cpu_socket_count.get('stdout').strip()
+            try:
+                if int(dmi_count) != 0:
+                    return dmi_count
+            except ValueError:
+                pass
+
+        # process the cpuinfo socket count as a fallback
+        cpuinfo_cpu_socket_count = \
+            dependencies.get('internal_cpu_socket_count_cpuinfo_cmd')
+        if cpuinfo_cpu_socket_count and \
+                cpuinfo_cpu_socket_count.get('rc') == 0 and \
+                cpuinfo_cpu_socket_count.get('stdout_lines'):
+            cpuinfo_count = cpuinfo_cpu_socket_count.get(
+                'stdout_lines', [0])[0]
+            try:
+                if int(cpuinfo_count) != 0:
+                    return cpuinfo_count
+            except ValueError:
+                pass
+
+        # assign the socket_count to the cpu_count as a last resort
+        cpu_count = dependencies.get('cpu_count')
+        return cpu_count
+
+
+class ProcessCpuCoreCount(process.Processor):
+    """Process the cpu core count."""
+
+    KEY = 'cpu_core_count'
+    DEPS = ['cpu_socket_count',
+            'cpu_core_per_socket',
+            'cpu_count',
+            'cpu_hyperthreading',
+            'virt_type']
+    REQUIRE_DEPS = False
+
+    @staticmethod
+    def process(output, dependencies):
+        """Return the correct value for cpu core count output."""
+        cpu_socket_count = dependencies.get('cpu_socket_count')
+        cpu_core_per_socket = dependencies.get('cpu_core_per_socket')
+        cpu_count = dependencies.get('cpu_count')
+        cpu_hyperthreading = dependencies.get('cpu_hyperthreading')
+        virt_type = dependencies.get('virt_type')
+        # if the virt_type is vmware and cpu_count exists
+        # then return cpu_count
+        if virt_type and virt_type == 'vmware' and cpu_count:
+            return int(cpu_count)
+        # if the cpu_core_per_socket & the cpu_socket_count are present
+        # return the product of the two
+        if cpu_core_per_socket and cpu_socket_count:
+            return int(cpu_core_per_socket) * int(cpu_socket_count)
+        if cpu_count:
+            if cpu_hyperthreading:
+                return int(cpu_count) / 2
+            # if there is no threading, return the cpu count
+            return int(cpu_count)
+        return ''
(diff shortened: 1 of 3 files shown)


quipucords/scanner/network/processing/cpu.py
```diff
@@ -74,3 +74,80 @@ class ProcessCpuBogomips(process.Processor):
...
+            try:
+                if int(cpuinfo_count) != 0:
+                    return cpuinfo_count
+            except ValueError:
+                pass
+
+        # assign the socket_count to the cpu_count as a last resort
+        cpu_count = dependencies.get('cpu_count')
+        return cpu_count
+
```
myersCody: You did an excellent job rewriting this. It is a lot easier to understand and is so clean. 🏅

quipucords/scanner/network/processing/cpu.py
```diff
@@ -74,3 +74,80 @@ class ProcessCpuBogomips(process.Processor):
     def process(output, dependencies=None):
         """Pass the output back through."""
         return get_line(output['stdout_lines'])
+
+
+class ProcessCpuSocketCount(process.Processor):
+    """Process the cpu socket count."""
+
```
myersCody: Quick question what are the KEY, DEPS, & REQUIRE_DEPS variables used for? I don't see them used in the static method, so I was just curious of how they are used outside of the class.

abaiken: For sure! So the processors defined in this file inherit from the processor class defined in  [process.py](https://github.com/quipucords/quipucords/blob/c212985e8ae57f9702e8b71a7414a8412ec5ce36/quipucords/scanner/network/processing/process.py#L170). You can read the comment within to learn more about them. Basically when you set `DEPS` the processor builds a dictionary to pass to the more specific processors (in this case the socket processor) that has the defined dependencies and their values inside. If the flag for `REQUIRE_DEPS` is set to true, it will not process the key you are trying to process if any of the dependencies are not found, so I had to set it to `false` so that it would still pass the values through but not skip processing if they were not available.

quipucords/scanner/network/processing/cpu.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2018 Red Hat, Inc.
+# Copyright (c) 2019 Red Hat, Inc.
```
kholdaway: should be 2018-2019 😬

quipucords/scanner/network/processing/cpu.py
```diff
@@ -74,3 +74,80 @@ class ProcessCpuBogomips(process.Processor):
...
+    @staticmethod
+    def process(output, dependencies):
+        """Return the correct value for cpu_socket_count output."""
+        # process the internal dmi cpu socket count result
+        dmi_cpu_socket_count = \
+            dependencies.get('internal_cpu_socket_count_dmi_cmd')
+        if dmi_cpu_socket_count and dmi_cpu_socket_count.get('rc') == 0:
+            dmi_count = dmi_cpu_socket_count.get('stdout').strip()
+            try:
+                if int(dmi_count) != 0:
```
kholdaway: Let's use the api/common/utils.py integer methods.

quipucords/scanner/network/processing/cpu.py
```diff
@@ -74,3 +74,80 @@ class ProcessCpuBogomips(process.Processor):
...
+        # process the cpuinfo socket count as a fallback
+        cpuinfo_cpu_socket_count = \
+            dependencies.get('internal_cpu_socket_count_cpuinfo_cmd')
+        if cpuinfo_cpu_socket_count and \
+                cpuinfo_cpu_socket_count.get('rc') == 0 and \
+                cpuinfo_cpu_socket_count.get('stdout_lines'):
+            cpuinfo_count = cpuinfo_cpu_socket_count.get(
+                'stdout_lines', [0])[0]
+            try:
+                if int(cpuinfo_count) != 0:
```
kholdaway: same

quipucords/scanner/network/processing/cpu.py
```diff
@@ -74,3 +74,80 @@ class ProcessCpuBogomips(process.Processor):
...
+                    return cpuinfo_count
+            except ValueError:
+                pass
+
+        # assign the socket_count to the cpu_count as a last resort
+        cpu_count = dependencies.get('cpu_count')
```
kholdaway: Return python numbers here (int/float)

quipucords/scanner/network/processing/cpu.py
```diff
@@ -74,3 +74,80 @@ class ProcessCpuBogomips(process.Processor):
...
+        # if the cpu_core_per_socket & the cpu_socket_count are present
+        # return the product of the two
+        if cpu_core_per_socket and cpu_socket_count:
+            return str(int(cpu_core_per_socket) * int(cpu_socket_count))
+        if cpu_count:
+            if cpu_hyperthreading:
+                return str(int(cpu_count) / 2)
+            # if there is no threading, return the cpu count
+            return str(cpu_count)
+        return ''
```
kholdaway: You can return None

quipucords/scanner/network/processing/cpu.py
```diff
@@ -74,3 +74,80 @@ class ProcessCpuBogomips(process.Processor):
...
+        """Return the correct value for cpu core count output."""
+        cpu_socket_count = dependencies.get('cpu_socket_count')
+        cpu_core_per_socket = dependencies.get('cpu_core_per_socket')
+        cpu_count = dependencies.get('cpu_count')
+        cpu_hyperthreading = dependencies.get('cpu_hyperthreading')
+        virt_type = dependencies.get('virt_type')
+        # if the virt_type is vmware and cpu_count exists
+        # then return cpu_count
+        if virt_type and virt_type == 'vmware' and cpu_count:
+            return str(cpu_count)
```
kholdaway: Return python numbers here (int/float)

quipucords/scanner/network/processing/cpu.py
```diff
@@ -74,3 +74,80 @@ class ProcessCpuBogomips(process.Processor):
...
+        # if the cpu_core_per_socket & the cpu_socket_count are present
+        # return the product of the two
+        if cpu_core_per_socket and cpu_socket_count:
+            return str(int(cpu_core_per_socket) * int(cpu_socket_count))
```
kholdaway: Return python numbers here (int/float)

quipucords/scanner/network/processing/cpu.py
```diff
@@ -74,3 +74,80 @@ class ProcessCpuBogomips(process.Processor):
...
+        if cpu_count:
+            if cpu_hyperthreading:
+                return str(int(cpu_count) / 2)
+            # if there is no threading, return the cpu count
+            return str(cpu_count)
```
kholdaway: Return python numbers here (int/float)

quipucords/scanner/network/processing/cpu.py
```diff
@@ -74,3 +74,80 @@ class ProcessCpuBogomips(process.Processor):
...
+        # then return cpu_count
+        if virt_type and virt_type == 'vmware' and cpu_count:
+            return str(cpu_count)
+        # if the cpu_core_per_socket & the cpu_socket_count are present
+        # return the product of the two
+        if cpu_core_per_socket and cpu_socket_count:
+            return str(int(cpu_core_per_socket) * int(cpu_socket_count))
+        if cpu_count:
+            if cpu_hyperthreading:
+                return str(int(cpu_count) / 2)
```
kholdaway: Return python numbers here (int/float)

quipucords/scanner/network/processing/test_cpu.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2018 Red Hat, Inc.
+# Copyright (c) 2019 Red Hat, Inc.
```
kholdaway: 2018-2019

PR #1749, opened 2019-01-11
Use vm.dns_name for vm_name if present

### LINKED ISSUE #1702: Use VM_dns_name for name if present
If a server is just found in vcenter could we use the vm_dns_name if present for "name" instead of vm_name because the vm_dns_name is more likely to match up then the vm_name

use the vm_dns_name if present for "name" instead of vm_name

___

## Acceptance Criteria:

- [ ] Verify that users can X Y Z
- [ ] Verify that an expected error is encountered when a user does A B C

___

abaiken:
Closes #1702

### DIFF
--- a/quipucords/scanner/vcenter/inspect.py
+++ b/quipucords/scanner/vcenter/inspect.py
@@ -208,10 +208,7 @@ class InspectTaskRunner(ScanTaskRunner):
                     facts['vm.cluster'] = host_facts.get('host.cluster')
                     facts['vm.datacenter'] = host_facts.get('host.datacenter')
 
-        if facts.get('vm.dns_name'):
-            vm_name = facts['vm.dns_name']
-        else:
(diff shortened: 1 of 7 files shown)


quipucords/scanner/vcenter/inspect.py
```diff
@@ -208,7 +208,10 @@ def parse_vm_props(self, props, host_dict):
                     facts['vm.cluster'] = host_facts.get('host.cluster')
                     facts['vm.datacenter'] = host_facts.get('host.datacenter')
 
-        vm_name = facts['vm.name']
+        if facts.get('vm.dns_name'):
+            vm_name = facts['vm.dns_name']
+        else:
+            vm_name = facts['vm.name']
```
myersCody: Should you do a `facts.get('vm.name')` here incase vm.name doesn't exist, or is it guaranteed to exist?

PR #1755, opened 2019-01-14
Change default port (#1651)

### LINKED ISSUE #1651: Change default port for Quipucords server to 9443
## Feature description:
Since QPC could potentially be installed on a system with other HTTPS services, we should choose a port other than the default. A good example is customers may have their Satellite server configured to be able to communicate with all RHEL servers in the network through SSH, so it's a good candidate to run QPC, but if QPC is installed, it will conflict with satellite on the default port. 

Potential to conflict with common applications

Change it to something less common through the automated installer and change the CLI to connect to that port by default. I think 8443 was suggested, but it might also be too common since JBoss EAP and others use it.

We will use 9443
___

## Acceptance Criteria:

- [ ] Verify that user who installs QPC with automated installer can log in without changing settings on CLI
- [ ] Verify that application configured for HTTPS already on the box are not affected by QPC's port

___

myersCody:
Closes #1651


client/package.json
```diff
@@ -45,7 +45,7 @@
     "start:ui-stage": "sh -ac '. .env.staging; react-app-rewired start'",
     "start": "run-p -l api:dev api:doc start:css start:ui",
     "start:stage": "open https://localhost:5001/login/; npm run build:clean; run-p -l api:stage start:css start:ui-stage",
-    "start:prod": "npm run api:prod; open https://localhost/login/",
+    "start:prod": "npm run api:prod; open https://localhost:9443/login/",
```
myersCody: I am uncertain about if this one needed to be changed or not.

docs/source/cli_configure.rst
```diff
@@ -10,12 +10,12 @@ The ``qpc server config`` command configures the connection between the qpc comm
 The ``qpc server config`` command takes the following options:
 
 - The ``--host`` option is required. If you are using the qpc command line interface on the same system where the server is running, you can supply the loopback address ``127.0.0.1`` as the value. Otherwise, supply the IP address for your Quipucords server.
-- The ``--port`` option is optional. The default value for this option is ``443``. If you decided to remap the Quipucords default exposed server port to another port, the port option is required. You must supply the port option and the remapped value in the command, for example, ``--port 8443``.
+- The ``--port`` option is optional. The default value for this option is ``9443``. If you decided to remap the Quipucords default exposed server port to another port, the port option is required. You must supply the port option and the remapped value in the command, for example, ``--port 9443``.
```
abaiken: The second part of this is talking about not using the default port (remapping the port), so give a different port number or leave it at ``--port 8443``.

docs/source/configure.rst
```diff
@@ -17,10 +17,10 @@ When you run the command to start the Quipucords server, you supply values for s
...
-   - If you want to remap the port on your system, you would supply a new value for the port when you run the Docker command to start the server. The syntax of this option is  ``-p <host_port>:<container_port>``. For example, to remap the port to ``8443``, you would enter the following option in the command: ``-p 8443:443``. Additionally, Docker supplies an option to select a free port for all exposed ports by using the ``-P`` option; the port mapping is then available from the ``sudo docker ps`` command.
+   - If you select to expose port 9443, you would use the following option when you run the Docker command to start the server: ``-p 9443:443``.
+   - If you want to remap the port on your system, you would supply a new value for the port when you run the Docker command to start the server. The syntax of this option is  ``-p <host_port>:<container_port>``. For example, to remap the port to ``9443``, you would enter the following option in the command: ``-p 9443:443``. Additionally, Docker supplies an option to select a free port for all exposed ports by using the ``-P`` option; the port mapping is then available from the ``sudo docker ps`` command.
```
abaiken: The second part of this is talking about remapping the port (not using the default) so we should use a value that is not the default, you could leave it at 8443.

docs/source/configure.rst
```diff
@@ -100,8 +100,8 @@ To change the default password for the Quipucords server, use the following step
...
 
-   - If the browser window is running on the same system as the server and you used the default port of ``443`` for the server, the URL is ``https://localhost/admin``.
-   - If the browser window is running on a remote system, or if it is on the same system but you changed the default HTTPS port, enter the URL in the following format: ``https://ip_address:port/admin``. For example, if the IP address for the server is 192.0.2.0 and the port is remapped to ``8443``, you would enter ``https://192.0.2.0:8443/admin`` in the browser window.
+   - If the browser window is running on the same system as the server and you used the default port of ``9443`` for the server, the URL is ``https://localhost:9433/admin``.
+   - If the browser window is running on a remote system, or if it is on the same system but you changed the default HTTPS port, enter the URL in the following format: ``https://ip_address:port/admin``. For example, if the IP address for the server is 192.0.2.0 and the port is remapped to ``9443``, you would enter ``https://192.0.2.0:9443/admin`` in the browser window.
```
abaiken: Like above, the second part of this is talking about remapping the port so we should use a value that isn't default.

install/install.sh
```diff
@@ -47,8 +47,8 @@ usage() {
          -e install_server=false
     * Skipping install for CLI:
          -e install_cli=false
-    * Specify server port (defaults to 443):
-         -e server_port=8443
+    * Specify server port (defaults to 9443):
+         -e server_port=9443
```
abaiken: This option is not as big of a deal as the above, but if the user is supplying the ```-e server_port=*``` option, they probably want to override the default, so 8443 or a different number could be used here as well.

client/package.json
kholdaway: I am pretty sure this one needs to be changed

PR #1760, opened 2019-01-17
Added hardware requirements. (#1737)

### LINKED ISSUE #1737: Add minimum system requirements for a host to run QPC to Docs
Would like to have it documented what are the minimum CPU, RAM and hard drive space needed to run QPC and what are the recommended amounts for each. E.g. will run with 1 GB RAM but recommended 4+ GB. 

___

## Acceptance Criteria:

- [ ] Verify that users can X Y Z
- [ ] Verify that an expected error is encountered when a user does A B C

___

myersCody:
Closes #1737


docs/source/requirements.rst
```diff
@@ -9,3 +9,9 @@ Before installing Quipucords in your environment, review the following guideline
 - The credential user account requires a ``bash`` shell. For example, the shell *cannot* be the ``/sbin/nologin`` or ``/bin/false`` shell.
 - Quipucords server requires a Postgres 9.6.10 database (this will be installed for scripted installations).
 - QPC CLI requires python 3.4+.
+
+Hardware Requirements
+~~~~~~~~~~~~~~~~~~~~~
+- *CPU:* 2 core minimum with a recommended 4 cores.
```
abaiken: can't they run this on a single core?

myersCody: CPU was actually the most difficult one for me to decide on. The reality is that different core counts & clock speeds make a big difference in the speed in which our scans work. It became kind of a balancing act between is it possible versus is it practical.  For example, while using only a single core & 1 gb of ram the `S62Scan` took about 30 mins to complete, I was kind of nervous to even run the LargeMixScan. If the customer is only scaning vcenter then yeah 1 core is perfect. I am open to thoughts and opinions on this? I can always run more test on the cpu count test at night to see how long a largemixscan takes on 1 cpu count.

PR #1770, opened 2019-01-24
Updated qpc copr repo locaton in Jenkinsfile

himmAllRight:
Updated qpc copr repo locaton in the Jenkinsfile and switched the deprecated 'archive' commands to 'archiveArtifacts'.


Jenkinsfile
```diff
@@ -52,7 +52,7 @@ def configureDocker = {
 
 def installQpcClient() {
     sh '''\
-    sudo wget -O /etc/yum.repos.d/chambridge-qpc-fedora-28.repo https://copr.fedorainfracloud.org/coprs/chambridge/qpc/repo/fedora-28/chambridge-qpc-fedora-28.repo
+    sudo wget -O /etc/yum.repos.d/group_quipucords-qpc-fedora-29.repo https://copr.fedorainfracloud.org/coprs/g/quipucords/qpc/repo/fedora-28/group_quipucords-qpc-fedora28.repo
```
himmAllRight: Could/should this be conditional?

elyezer: We are using Fedora 28 to build this, so the repo should be the one for that OS.

elyezer: Right now we are using a single OS to build and test, if we start testing on more OSes then we should add a conditional to properly select the repo.

himmAllRight: Got it. I was unsure about which to use.

himmAllRight: actually looking at this... it is using 28, but pulling it down to a file named 29. I'll fix that...

PR #1772, opened 2019-01-24
Disabled local cli rpm search for online installs.

myersCody:


install/roles/qpc_cli/tasks/main.yml
```diff
@@ -137,7 +139,7 @@
   when:
     - not is_fedora
     - install_cli
-    - find_qpc_rpm_local.stat.exists == false
+    - install_offline is not defined
```
abaiken: same as above

myersCody:
Let's pretend this didn't happen.

PR #1837, opened 2019-02-11
Modify roles to support Ansible Runner (#1835)

### LINKED ISSUE #1835: Modify Roles to Support Ansible Runner
## Feature description:

**playbook**
```
- name: example
  raw: echo "bar"
  register: foo
```

Ansible Runner currently only provides the `stdout` of raw commands. However, we process system information by associating the register name `foo` as a key and the `stdout` as the value e.g. `{'foo':'bar'}`.

In order for ansible runner to work with our current process we need to ensure that `raw` commands are captured through a `set_fact`. 

For example: 

```
- name: example
  raw: echo "bar"
  register: internal_foo

- name: set_foo_fact
  set_fact:
    foo: "{{internal_foo}}"
  ignore_errors: yes
```

The `set_fact` method will still provide us with the ability to obtain a key and value.

myersCody:
Closes #1835


roles/cpu/tasks/main.yml
```diff
@@ -4,29 +4,54 @@
...
+  register: internal_cpu_model_name
+  ignore_errors: yes
+
+- name: set cpu_model_name fact
+  set_fact:
+    cpu_model_name: "{{internal_cpu_model_name}}"
   ignore_errors: yes
 
-- name: gather cpu.bogomips fact
+- name: gather intenal_cpu_bogomips fact
```
abaiken: fix misspelling in internal

roles/jboss_brms/tasks/main.yml
```diff
@@ -85,46 +100,76 @@
...
-  register: jboss_brms_manifest_mf
+  register: internal_jboss_brms_manifest_mf
   ignore_errors: yes
   become: yes
   with_items: "{{ (business_central_candidates + decision_central_candidates + kie_server_candidates + internal_jboss_brms_kie_search_candidates.get('stdout_lines', []))|difference(['']) }}"
   when: 'user_has_sudo and jboss_brms'
 
+- name: set jboss_brms_manifest_mf fact
+  set_fact:
+    jboss_brms_manifest_mf: '{{ internal_jboss_brms_manifest_mf }}'
```
abaiken: should we use " instead of ' to be consistent? (for all of the set_facts i feel like we should pick one and be consistent)

roles/jboss_fuse_on_karaf/tasks/main.yml
```diff
@@ -40,131 +45,189 @@
...
 # Combine the outputs of the above into a single fact
 
 - name: combine KARAF_HOME candidates into single list
   set_fact:
       karaf_homes: "{{
-          (karaf_running_processes.get('stdout_lines', []) +
-            karaf_locate_karaf_jar.get('stdout_lines', []) +
-            karaf_find_karaf_jar.get('stdout_lines', []))|difference(['']) | unique }}"
+          (internal_karaf_running_processes.get('stdout_lines', []) +
+            internal_karaf_locate_karaf_jar.get('stdout_lines', []) +
```
abaiken: it could be the way the diff is showing up but should the second two variables indention be lined up underneath the first?

roles/jboss_brms/tasks/main.yml
myersCody: 🤕

abaiken: 😬

roles/jboss_fuse_on_karaf/tasks/main.yml
myersCody: 🤕

abaiken: 😬 sry

roles/jboss_brms/tasks/main.yml
kholdaway: I believe in your Cody.  You can make it consistent.

PR #1857, opened 2019-02-25
Migrate QPC server to Ansible 2.7

### LINKED ISSUE #1803: Migrate QPC server to Ansible 2.7
## Feature description:

After QPC is using Ansible runner, upgrade QPC ansible to 2.7.

## Acceptance Criteria:

- [ ] Ensure the details and deployment report are the same as when using ansible 2.3

myersCody:
Closes #1803


quipucords/scanner/network/connect.py
```diff
@@ -336,10 +336,22 @@ def _connect(manager_interrupt,
...
+            message = _construct_playbook_error_msg(final_status)
             if final_status not in ['unreachable', 'failed']:
-                scan_task.log_message(log_message)
-                raise AnsibleRunnerException(final_status)
+                if not scan_task.systems_scanned:
+                    log_message = 'Unexpected ansible status %s.  '\
+                        'No systems_scanned so failing.  Ansible error: %s' % (
+                            final_status, message)
+                    scan_task.log_message(log_message, log_level=logging.ERROR)
+                    raise AnsibleRunnerException(final_status)
```
kholdaway: Looking at this, we are inconsistent between how we handle error in connect and inspect.  Connect throws an exception while inspect returns a failed status and error.  I think inspect follows the prescribed model for the ScanTask run method.  The run() method returns a tuple with a status and message.  Exceptions should be for unexpected (not caught issues).

Let's change this to return the failed status and appropriate message.  We can log in both cases but return failed instead of the exception.

quipucords/scanner/network/connect.py
```diff
@@ -336,10 +336,22 @@ def _connect(manager_interrupt,
 
         final_status = runner_obj.status
         if final_status != 'successful':
-            log_message = _construct_playbook_error_msg(final_status)
+            message = _construct_playbook_error_msg(final_status)
             if final_status not in ['unreachable', 'failed']:
-                scan_task.log_message(log_message)
-                raise AnsibleRunnerException(final_status)
+                if not scan_task.systems_scanned:
+                    log_message = 'Unexpected ansible status %s.  '\
```
kholdaway: Also, I know I provided the text for these messages, but I hate them know.  Very relaxing drive home brought me inner peace.  Let's use this instead.

```
Unexpected ansible status %s.  Failing scan because there were no successful system connections.  Ansible error: %s
```

quipucords/scanner/network/connect.py
```diff
@@ -336,10 +336,22 @@ def _connect(manager_interrupt,
...
+                        'No systems_scanned so failing.  Ansible error: %s' % (
+                            final_status, message)
+                    scan_task.log_message(log_message, log_level=logging.ERROR)
+                    raise AnsibleRunnerException(final_status)
+                else:
+                    log_message = 'Unexpected ansible status %s.  '\
```
kholdaway:
```
Unexpected ansible status %s.  Continuing scan because there were %s successful system connections.  Ansible error: %s
```

quipucords/scanner/network/connect.py
myersCody: I believe that the reason the `inspect` and the `connect` scan have different error handling is because of how often they are used in the `run` method. For the connect scan the `_connect` method is actually called multiple times depending on the amount of credientials associated with the source. 

For example: 

```
credentials = [cred0, cred1]
for cred in credentials:
	_connect(parameters)

# Logic to determine if the scan task has failed or is completed. 
```

Therefore we cannot return a scan task failed inside of the `_connect` method because we don't know if the scan task will fail on later iterations of the credentials. For example, just because cred0 failed doesn't mean that we should report that the scan task is failed because cred1 could be successful. Which is why the logic to determined a failed scan task for the connection happens after the for loop. Whereas with the `inspect` scan we only run `_inspect` method once per `run` therefore we are able to determine a failure inside of the `_inspect` method.

requirements.txt
```diff
@@ -1,6 +1,4 @@
-# Include Locked Server Requirements
--r locked-requirements.txt
-
+ansible==2.7.0.0
```
kholdaway: Let's just go with 2.7.0

quipucords/messages.py
```diff
@@ -0,0 +1,23 @@
...
+#
+# This software is licensed to you under the GNU General Public License,
+# version 3 (GPLv3). There is NO WARRANTY for this software, express or
+# implied, including the implied warranties of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv3
+# along with this software; if not, see
+# https://www.gnu.org/licenses/gpl-3.0.txt.
+#
+#
+"""Log messages for Quipucords."""
```
kholdaway: let's put these messages in the log_messages module

PR #1877, opened 2019-03-19
Reduce facts sent to insights platform

abaiken:
- add cpu_core_per_socket to deployments report 
- make sure that we are only sending over facts that we use for insights


quipucords/api/insights_report/view.py
```diff
@@ -124,9 +124,16 @@ def get_hosts_from_fp(report, host_dicts):
     valid_hosts = verify_report_hosts(host_dicts)
     insights_hosts = {}
     for host in valid_hosts:
+        insights_host = {}
+        for fact in INSIGHTS_FACTS:
+            if host.get(fact):
+                insights_host[fact] = host.get(fact)
+        # for fact in facts_to_remove:
```
kholdaway: 🤔

abaiken: 😅

PR #1905, opened 2019-04-17
Install script will now open server port in firewall.

### LINKED ISSUE #1820: QPC server install should automatically open the QPC server port on machine firewall
## Feature description:

QPC install does not open any external ports on the machine running the QPC server.  QPC server would not be accessible from another machine.  This request is to have the QPC install open the QPC server port on the machine running the server.

myersCody:
closes #1820 

# Test
```
# Add internal yum repos (if needed)
iptables -xvn -L | grep 9443
#should see nothing
./install.sh
iptables -xvn -L | grep 9443
#       17     2056 ACCEPT     tcp  --  *      *       0.0.0.0/0            0.0.0.0/0           tcp dpt:9443
```


install/roles/qpc_server/tasks/main.yml
```diff
@@ -10,6 +10,13 @@
     server_port: "{{ ansible_env['SERVER_PORT'] | default('9443') }}"
   when: server_port is not defined
 
+- name: open port in firewall
+  iptables:
+    chain: INPUT
+    destination_port: "{{server_port}}"
+    protocol: tcp
+    jump: ACCEPT
+  ignore_errors: yes
```
abaiken: I think we should add `become: true` here.

abaiken: Also, I feel like we should have an `open_port` flag or something so that they can turn on/off this functionality

install/roles/qpc_server/tasks/main.yml
```diff
@@ -10,6 +10,26 @@
     server_port: "{{ ansible_env['SERVER_PORT'] | default('9443') }}"
   when: server_port is not defined
 
+- name: Set open port
```
abaiken: since we added the ability to set this flag we need to document it in the `install.sh` usage

install/install.sh
```diff
@@ -47,6 +47,8 @@ usage() {
          -e install_server=false
     * Skipping install for CLI:
          -e install_cli=false
+    * Skipping opening the server port in the firewall (if not specified the default value is 'true'):
```
abaiken: Skipping doesn't really make sense. We should say:
 ```Optionally open the server port in the firewall (defaults to true)```

myersCody:
https://drive.google.com/open?id=1vx7lpASBV3S8e7s0zjB8ii6oPKbAwvGv

kholdaway:
Looks good!

PR #1915, opened 2019-05-06
Swagger updates for insights reports

kholdaway:


docs/swagger.yml
```diff
@@ -1505,29 +1506,58 @@ definitions:
...
+            description: "The report version"
+            example: "1.0.0.afdsgev"
+          report_platform_id:
+            type: "string"
+            description: "Globally unique uuid for report"
+            example: "be3075ac-84d3-4b62-9f5c-a418a36f802d"
+          report_slices:
+            type: object
+            properties:
+              system_platform_id:
```
abaiken: instead of `system_platform_id` wouldn't this be `report_slice_id`?

kholdaway: Take a look again.

PR #1919, opened 2019-05-13
Allow the versions to be passed in as an argument to the installer

### LINKED ISSUE #1927: Install script should be generalized so it will install any version and does not reference a specific version
Install script should be generalized so it will install any version and does not reference a specific version

kholdaway:
Closes #1927

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -33,9 +33,9 @@ usage() {
 
 
     Extra Variables:
-    * Specify the version of the quipucords server to install.   (defaults to the latest release):
+    * Specify the version of the quipucords server to install (defaults to the latest release):
          -e server_version=0.0.46
-    * Specify the version of the qpc cli to install.   (defaults to the latest release):
+    * Specify the version of the qpc cli to install (defaults to the latest release):
          -e cli_version=0.0.46
     * Specify if quipucords docker container should run the server without supervisord(defaults to true):
          -e use_supervisord=false
--- a/install/install.sh
+++ b/install/install.sh
@@ -13,12 +13,11 @@
 
 export PATH=$PATH:$ANSIBLE_HOME/bin
 PLAYBOOKFILE="qpc_playbook.yml"
-RELEASE_TAG='-e RELEASE_TAG=1.0.0'
 POSTGRES_VERSION='-e POSTGRES_VERSION=9.6.10'
 
 declare -a args
 args=("$*")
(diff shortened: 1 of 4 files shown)


kholdaway:
This script was able to install 0.0.45 and then upgrade to 0.0.46.  In other words, the install script is not tied to a specified version.

install/install.sh
```diff
@@ -39,6 +33,10 @@ usage() {
 
 
     Extra Variables:
+    * Specify the version of the quipucords server to install.   (defaults to the latest release):
```
abaiken: Remove the awkward spacing before `(defaults...`

install/install.sh
```diff
@@ -39,6 +33,10 @@ usage() {
...
+         -e server_version=0.0.46
+    * Specify the version of the qpc cli to install.   (defaults to the latest release):
```
abaiken: Same comment as above

install/roles/qpc_server/tasks/main.yml
```diff
@@ -31,41 +31,89 @@
...
 
-- name: Set server image name
+- name: Set server docker image name
   set_fact:
-    image: "{{ ansible_env['SERVER_IMAGE'] | default('quipucords') }}"
-  when: image is not defined
+    server_docker_image_name: "quipucords"
 
-- name: Set server_image_and_tag
+- name: Set Quipucords docker image and tag
```
abaiken: not a blocker at all but there are a lot of inconsistencies within the name descriptions, for example, Quipucords vs quipucords, Postgres vs postgres

install/install.sh
kholdaway: Done

install/install.sh
kholdaway: Done

install/roles/qpc_server/tasks/main.yml
kholdaway: Fixed

install/roles/prep/tasks/main.yml
```diff
@@ -1,5 +1,10 @@
 ---
 
+- name: Checking if server_version is defined when install_offline is true
+  fail:
+  when:
+    - install_offline is defined and server_version is not defined
```
abaiken: What happens if `install_offline` is defined but is set to false?

kholdaway: Fixed.

PR #1943, opened 2019-06-04
Add canonical facts onto top level of hosts section in the report

### LINKED ISSUE #1941: Iterate on Insights report (missing canonical facts & potentially contains None)
## Bug Description:
Provide a description of the issue being raised.
The insights report is missing the top level canonical facts. See the [api spec](https://github.com/RedHatInsights/insights-host-inventory/blob/master/swagger/api.spec.yaml#L402) for details. 

### Expected behavior:
The insights report should have the top level canonical facts. For example, in the above spec under properties there is a `mac_addresses` field. This field is not present in the report generated by Quipucords. This should be the case with the rest of the canonical facts as well. 

### Actual behavior:
The insights report should have the top level canonical facts. 

### Steps to reproduce:
Run a scan and download an insights report - look at the hosts field inside the slice. 

___

## Acceptance Criteria:

- [ ] Verify that the hosts field in the insights report slice contains the top level canonical facts. 

___

Dostonbek1:
**Added canonical facts onto top level of hosts section in the report**

closes #1941

### DIFF
--- a/docs/swagger.yml
+++ b/docs/swagger.yml
@@ -1377,9 +1377,256 @@ definitions:
           $ref: "#/definitions/SystemFact"
         description: "The facts associated with this source"
         minItems: 1
+  NestedFacts:
+    title: Nested facts 
+    description: Nested facts of a host 
+    type: object
+    properties:
+      name:
+        type: "string"
+        description: "The system name"
+        example: "hostname.com"
+      os_release:
+        type: "string"
+        description: "The operating system release"
+        example: "Red Hat Enterprise Linux Server release 5.9 (Tikanga)"
+      os_version:
+        type: "string"
+        description: "The operating system version"
+        example: "5.9 (Tikanga)"
+      infrastructure_type:
+        type: "string"
+        enum:
+          - virtualized
+          - bare_metal
+          - hypervisor
+          - unknown
+        description: "The infrastructure type"
+      mac_addresses:
+        type: "array"
+        items:
+          type: "string"
+          description: "MAC addresses of system"
+          example: "0B:AA:CC:44:A8:25"
+        description: "The infrastructure type"
+      ip_addresses:
+        type: "array"
+        items:
+          type: "string"
+          description: "IP addresses of system"
+        example: ["192.168.0.1","192.168.0.2"]
+        description: "The infrastructure type"
+      cpu_count:
+        type: "integer"
+        format: "int64"
+        description: "The count of cpus"
+        example: 8
+      cpu_core_per_socket:
+        type: "integer"
+        format: "int64"
+        description: "The cpu core count per socket."
+        example: 1
+      architecture:
+        type: "string"
+        description: "The system architecture"
+        example: "x86_64"
+      bios_uuid:
+        type: "string"
+        description: "The system bios uuid"
+        example: "65541E42-1234-A2C4-2BB4-F82448BDCEC"
+      insights_client_id:
+        type: "string"
+        description: "The insights client id"
+        example: "b077190b74f14b3582e57ce8243ae356"
+      etc_machine_id:
+        type: "string"
+        description: "The system machine id"
+        example: "b077190b74f14b3582e57ce8243ae356"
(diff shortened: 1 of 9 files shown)


quipucords/fingerprinter/task.py
```diff
@@ -348,6 +348,10 @@ def _process_details_report(self, manager_interrupt, details_report):
                             fingerprint_dict.get('name')
                         insights_host['fqdn'] = \
                             fingerprint_dict.get('name')
+                        for fact in CANONICAL_FACTS:
```
abaiken: I think the canonical facts expected by the host inventory are slightly different than the canonical facts that are defined in quipucords. For example, the ones expected by host inventory are `bios_uuid`, `ip_addresses`, `mac_addresses`, `insights_client_id`, `rhel_machine_id`, `subscription_manager_id`. 

We will also need to update any affected test cases & update the swagger to include the new fields (canonical facts). 

One other thing - is we need to check and see if the system profile is empty before adding it to the insights host.

abaiken: Additionally, I think in the api spec, `insights_client_id` (from quipucords) will have to be renamed to `insights_id` when we add it to the host. 

We should talk to @kholdaway too because in one document, the `vm_uuid` is defined as a canonical fact - but there is no field for it in the api spec (which means we are only ever adding it to the nested facts field).

quipucords/fingerprinter/task.py
```diff
@@ -338,6 +338,15 @@ def _process_details_report(self, manager_interrupt, details_report):
                     for fact in CANONICAL_FACTS:
                         if fingerprint_dict.get(fact):
                             found_canonical_facts = True
+                            not_null_facts = {
```
abaiken: I would move this definition to line 355 so that it is clear where it is being used.

quipucords/fingerprinter/task.py
```diff
@@ -360,8 +373,9 @@ def _process_details_report(self, manager_interrupt, details_report):
                                      fingerprint_dict.get('products', []),
                                      fingerprint_dict.get('is_redhat'))}
                         insights_host['facts'] = [facts]
-                        insights_host['system_profile'] = \
-                            self.format_system_profile(nested_facts)
+                        if bool(nested_facts):
```
abaiken: using bool is fine here but unnecessary, you also can just check `if nested_facts`. If you like bool for readability then leave it 😄

Dostonbek1: I thought about that. I wasn't sure whether `if nested_facts` returns `False` when the dictionary is empty or when the variable does not exist.

Dostonbek1: I just checked. You are right. :-D

quipucords/fingerprinter/tests_fingerprint_task.py
```diff
@@ -1138,6 +1138,7 @@ def test_process_details_report_success(self):
                 'display_name':
```
abaiken: The only other thing I would do, is possibly add `etc_machine_id` & `insights_client_id` to the test `fact_collection` so that we can check that it shows up in the `expected_insights` dictionary as `rhel_machine_id` and `insights_id` at the top level, and their qpc names in the nested facts (since those two facts are the only ones we have to rename it would be a good regression test).

docs/swagger.yml
```diff
@@ -1377,9 +1377,256 @@ definitions:
...
+          type: string
   Host:
-    type: "object"
-    description: "Host data should conform to the host inventory swagger definition described here: https://github.com/RedHatInsights/insights-host-inventory/blob/system_profile_changes/swagger/api.spec.yaml"
+    title: Host data
+    description: >-
+      Data of a single host belonging to an account. Represents the hosts
+      without its Inventory metadata.
+    type: object
+    properties:
```
abaiken: Instead of updating the properties - we need to update the example and remove these. We want this to be an empty object with just an example. 

Also - remove the facts properties/nested facts properties

PR #1957, opened 2019-06-12
Update file pattern for Jenkins builds

kholdaway:


Jenkinsfile
```diff
@@ -27,7 +27,7 @@ node('f28-os') {
         sh "sudo setenforce 0"
     }
     stage('Copy UI into Server') {
-        copyArtifacts filter: 'quipucords-ui.*.tar.gz', fingerprintArtifacts: true, projectName: "qpc_${VERSION}_ui_distribution", selector: lastCompleted()
+        copyArtifacts filter: 'quipucords-ui-dist.tar', fingerprintArtifacts: true, projectName: "qpc_${VERSION}_ui_distribution", selector: lastCompleted()
```
myersCody: You dropped the `.gz` from the copyArtifacts. However, in the next line you try to untar a `tar -xvf quipucords-ui.*.tar.gz`. Additionally, does that line need to be changed too?

kholdaway: It is. Good eye.  Fixed.

PR #1962, opened 2019-06-18
Extract build version to Jenkins

### LINKED ISSUE #1961: Extract quipucords version out of source and have Jenkins inject the version
Extract quipucords version out of source and have Jenkins inject the version

kholdaway:
Closes #1961


Jenkinsfile
```diff
@@ -9,8 +8,6 @@ def postgres_dir = "postgres.${postgres_version}"
 def postgres_targzfile = "postgres.${postgres_version}.tar.gz"
 def postgres_license = "PostgreSQL_License.txt"
 def rename_license = "${postgres_dir}/license.txt"
-def release_info_template_file = "release_info_template.json"
-def release_info_file = "release_info.json"
```
myersCody: I think that removing this line means that we also should remove line [64](https://github.com/quipucords/quipucords/pull/1962/files#diff-58231b16fdee45a03a4ee3cf94a9f2c3L64)

Because renamed the template to `release_info.json` so it may still work.

quipucords/quipucords/tests_environment.py
```diff
@@ -72,3 +73,8 @@ def test_startup(self, mock_logger):  # pylint: disable=no-self-use
         """Test the startup method."""
         environment.startup()
         mock_logger.assert_called_with(ANY, ANY)
+
+    def test_server_version_default(self):
+        """Test the server version default."""
+        expected = '0.0.0.%s' % environment.commit()
```
myersCody: Should we use the `DEFAULT_VERSION` from `release.py` just in case the default ever changes?

Jenkinsfile
```diff
@@ -43,6 +40,11 @@ node('f28-os') {
         sh 'cat GIT_COMMIT'
         def commitHash = readFile('GIT_COMMIT').trim()
 
+        sh "sed s/BUILD_VERSION_PLACEHOLDER/${BUILD_VERSION}/ release_info.json > temp_release_info.json"
```
myersCody: There may be limitations because of Jenkins. 

However, you can use `sed -i` to find & replace within the file. 
`sh "sed -i s/BUILD_VERSION_PLACEHOLDER/${BUILD_VERSION}/g release_info.json"`

For example, I tried this locally: 
```
[root@rhel8 django-rest-framework]# sed -i s/BUILD_VERSION_PLACEHOLDER/0.0.9/g release_info.json
[root@rhel8 django-rest-framework]# cat release_info.json
{
	"version": "0.0.9"
}
[root@rhel8 django-rest-framework]#
```
That way you don't have to do a `sh "mv temp_release_info.json release_info.json"` right after.

quipucords/quipucords/tests_environment.py
```diff
@@ -16,6 +16,7 @@
 from django.test import TestCase
 
 from . import environment
+from . import release
```
myersCody: I don't think you need this import.

quipucords/quipucords/tests_environment.py
kholdaway: yep

quipucords/quipucords/tests_environment.py
kholdaway: Need it to implement your other suggestion

Jenkinsfile
kholdaway: removed that

Jenkinsfile
kholdaway: Fixed

PR #1984, opened 2019-07-23
Add satellite virtual_host_guest to deployment and insights reports

### LINKED ISSUE #1980: Investigate and size whether QPC can gather virtual guest's host from Satellite scans
## Feature description:
Determine difficulty and whether satellite API exposes this information to us.

### Details
Gather virtual guest's host from Satellite scans. Currently, QPC only gather's this information from vCenter scans. Satellite uses virt_who to determine this information. QPC should:

- [ ]   Extract this information from satellite
- [ ]   Include this information in insights reporting

___

kholdaway:
Closes #1980


quipucords/fingerprinter/task.py
```diff
@@ -1371,7 +1371,7 @@ def _process_vcenter_fact(self, source, fact):
         self._add_fact_to_fingerprint(source, 'vm.dns_name', fact,
                                       'vm_dns_name', fingerprint)
         self._add_fact_to_fingerprint(source, 'vm.host.name',
```
myersCody: Just out of curiosity why do we still have the `vm.host.name` in the fingerprinter here?

quipucords/fingerprinter/tests_fingerprint_task.py
```diff
@@ -410,7 +410,7 @@ def _validate_vcenter_result(self, fingerprint, fact):
         self.assertEqual(fact.get('vm.dns_name'),
                          fingerprint.get('vm_dns_name'))
         self.assertEqual(fact.get('vm.host.name'),
-                         fingerprint.get('vm_host'))
+                         fingerprint.get('virtual_host_name'))
 
         self.assertEqual(fact.get('vm.host.cpu_count'),
```
myersCody: Do we need to test for the `virtual_host_name` fact here?  Am I right in thinking that the `vm.host.name` & `virtual_host_name` are separate facts?

quipucords/fingerprinter/task.py
kholdaway: It's a mapping between details report names and deployment report names.

quipucords/fingerprinter/tests_fingerprint_task.py
kholdaway: The names in details don't always match the deployments.

PR #1995, opened 2019-08-07
Add ability to set the username/password/email for a user during installation

kholdaway:
This PR allows you to set the username/password/email for the quipucords server.  To test follow the instructions for building and running the container documented here --> https://github.com/quipucords/quipucords#container-image.  However, you must replace the run command for the quipucords server with:

```
docker run --name quipucords --link qpc-db:qpc-link -d -e QPC_SERVER_USERNAME=kevan -e QPC_SERVER_EMAIL=kholdawa@redhat.com -e QPC_SERVER_PASSWORD=nicekevan -e QPC_DBMS_HOST=qpc-db -p 9443:443 -v $QPC_VAR_DATA:/var/data -i quipucords:0.9.1
```

You can then docker exec and cat the quipucords logs to see that your user was created.

Now, if you `docker rm -f quipucords` and re-run just the run command and view logs you should see that the user already exists and will not be recreated.

### DIFF
--- a/deploy/setup_user.py
+++ b/deploy/setup_user.py
@@ -1,0 +1,33 @@
+#
+# Copyright (c) 2019 Red Hat, Inc.
+#
+# This software is licensed to you under the GNU General Public License,
+# version 3 (GPLv3). There is NO WARRANTY for this software, express or
+# implied, including the implied warranties of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv3
+# along with this software; if not, see
+# https://www.gnu.org/licenses/gpl-3.0.txt.
+#
+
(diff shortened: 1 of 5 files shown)


deploy/setup_user.py
```diff
@@ -0,0 +1,33 @@
...
+
+ADMIN_NOT_PRESENT = User.objects.filter(
+    username=QPC_SERVER_USERNAME).count() == 0
+
+if ADMIN_NOT_PRESENT:
+    User.objects.create_superuser(
+        QPC_SERVER_USERNAME,
+        QPC_SERVER_USER_EMAIL,
+        QPC_SERVER_PASSWORD)
+    print('Created username %s' % QPC_SERVER_USERNAME)
```
abaiken: should we change this to say `Created user %s` instead of username?

PR #2002, opened 2019-08-20
Update container image so that the image can work for podman on rhel7

myersCody:


Dockerfile
```diff
@@ -14,10 +14,13 @@ RUN mkdir -p /app
...
 
-# Create /etc/ssl
-RUN mkdir -p /etc/ssl/
-COPY deploy/ssl/* /etc/ssl/
-VOLUME /etc/ssl
+# Create /etc/ssl/qpc
+RUN mkdir -p /etc/ssl/qpc/
+VOLUME /etc/ssl/qpc/
+COPY deploy/ssl/server.crt /etc/ssl/qpc/
+COPY deploy/ssl/server.csr /etc/ssl/qpc/
```
myersCody: I removed the wildcard here to make it easier for me to tell if we added a new file for downstream. Cause it will show up in the git diff.

PR #2019, opened 2019-11-15
Add system_purpose json to the insights report

### LINKED ISSUE #2012: User Story: QPC should send system purpose up to platform for qpc/discovery systems
## Feature description:

QPC should send system purpose up to platform for qpc/discovery systems

kholdaway:
Closes #2012


quipucords/api/deployments_report/model.py
```diff
@@ -115,6 +115,7 @@ class SystemFingerprint(models.Model):
     system_creation_date = models.DateField(null=True)
     system_last_checkin_date = models.DateField(null=True)
 
+    system_purpose = models.TextField(unique=False, null=True)
```
abaiken: you should also add this to the `__str__` function

kholdaway: Good point!

docs/swagger.yml
```diff
@@ -2372,22 +2372,32 @@ definitions:
           type: "string"
           description: "The insights client id"
           example: "b077190b74f14b3582e57ce8243ae356"
-        system_intent_role:
+        system_purpose:
```
abaiken: https://github.com/quipucords/quipucords/blob/31c306cd82dbc067c7f1378ea4b6e8786ce02f74/docs/swagger.yml#L1380 

Should we also add it to the Host definition in the swagger?

kholdaway: Done

PR #2024, opened 2019-11-27
Add facts about how many users are on a system

### LINKED ISSUE #2009: Determine number of users on Linux OS
## Feature description:

Systems can sometimes be accessed by multiple users. This data is important to know when understanding if a system is being used as a single-user or multi-user. 

Currently when reviewing the use of a system for development purpose or production, we are not able to determine this from the current system facts returned within a network scan report.

We would like to add a new column to the network scan reporting which allows us to pull user data on a Linux system. Integrating the following commands into our current task list when running a network scan.

tail -5 /etc/passwd
tail -25 | last
history | grep userdel
___

## Acceptance Criteria:

Running the commands mentioned above returning the users on a system. 
Checking the frequency of the user or users logging into the Linux system
Verifying that users were not deleted recently

___

## Additional context
The simplest option would be to add 2 or 3 additional columns which populate this data when running a network scan and creating a detailed report.

abaiken:
Closes #2009 

To test: 
1. Run a network scan
2. See that the new facts exist for both details/deployments reports; `system_user_count` `user_login_history`. 

***_Note_***: 
-  In order to check that login is working - you should login to one of the systems that you are scanning as a different user than the one that you scan with and make sure that it is in the history. 
 - To check the number of users - you can also login to a system and add users/manually inspect the /etc/passwd file and make sure that the number of real users matches the number reported.


quipucords/scanner/network/runner/roles/user_data/tasks/main.yml
```diff
@@ -0,0 +1,57 @@
...
+    internal_user_login_history: "{{ internal_user_login_history_cmd }}"
+  ignore_errors: yes
+
+- name: initialize user_login_history
+  set_fact:
+    user_login_history: "QPC_FORCE_POST_PROCESS"
+  ignore_errors: yes
+
+- name: gather user_delete_history data fact
+  raw: cat /var/log/secure 2>/dev/null | grep userdel
```
myersCody: So, I think this command is missing some of the data. So, the issue with only looking at `/var/log/secure` is that the logs do get rotated over time. For example:
```
[root@dhcp181-3 ~]# ls /var/log/ | grep secure
secure
secure-20191110
secure-20191117
secure-20191124
secure-20191201
```

I think we can change the command to include the logs files that were rotated out of circuit. For example,
```
[root@dhcp181-3 ~]# grep userdel /var/log/secure*
/var/log/secure:Dec  2 10:28:39 dhcp181-3 userdel[26270]: delete user 'tmp_user'
/var/log/secure:Dec  2 10:28:39 dhcp181-3 userdel[26270]: removed group 'tmp_user' owned by 'tmp_user'
/var/log/secure:Dec  2 10:28:39 dhcp181-3 userdel[26270]: removed shadow group 'tmp_user' owned by 'tmp_user'
/var/log/secure-20191201:Nov 27 09:34:21 dhcp181-3 userdel[9776]: delete user 'newuser'
/var/log/secure-20191201:Nov 27 09:34:21 dhcp181-3 userdel[9776]: removed group 'newuser' owned by 'newuser'
/var/log/secure-20191201:Nov 27 09:34:21 dhcp181-3 userdel[9776]: removed shadow group 'newuser' owned by 'newuser'
```

abaiken: Great suggestion! I will make this change 👍

PR #2032, opened 2019-12-09
Add a query param to mask sensitive report data

abaiken:
the query param is `mask` and it expects a boolean value.

---

Test local development 
- [x] Vcenter Scan
- [x] Nework Scan
- [x] Satellite Scan

You can download the tar.gz by going to the endpoint with the query param:
http://127.0.0.1:8000/api/v1/reports/4/?mask=true

---

Test using a container in docker. 

- [ ] Run a scan, bring down the container & bring it back up, and run the scan again. We will need to make sure that the mask values can still be deduplicated on.

---

### *Note: Copy the .env.example to your .env file or add the `PYTHONHASHSEED=0` variable to your env file.


quipucords/api/deployments_report/csv_renderer.py
```diff
@@ -27,4 +27,4 @@ def render(self,
                accepted_media_type=None,
                renderer_context=None):
         """Render report as CSV."""
-        return create_deployments_csv(data)
+        return create_deployments_csv(data, renderer_context.get('request'))
```
kholdaway: Why did you need to add the render_context param?

kholdaway: nm

quipucords/api/deployments_report/view.py
```diff
@@ -64,27 +64,41 @@ def deployments(request, pk=None):
...
     :returns: json report data
     :raises: Raises validation error group_count on non-existent field.
     """
-    return {'report_id': report.id,
-            'status': report.status,
-            'report_type': report.report_type,
-            'report_version': report.report_version,
-            'report_platform_id': str(report.report_platform_id),
-            'system_fingerprints': json.loads(report.cached_fingerprints)}
+    system_fingerprints = json.loads(
```
kholdaway: Let's make an else clause for validate_query_param_bool and put this json.loads there.  We are loading this and not using it for the mask param.

quipucords/api/deployments_report/view.py
```diff
@@ -64,27 +64,41 @@ def deployments(request, pk=None):
...
                          'Deployment report %s could not be created.'
                          '  See server logs.' % report.details_report.id},
                         status=status.HTTP_424_FAILED_DEPENDENCY)
 
-    return Response(build_cached_json_report(report))
+    return Response(build_cached_json_report(report, mask_report))
 
 
-def build_cached_json_report(report):
+def build_cached_json_report(report, mask_report):
```
myersCody: We may want to add `mask_report` to the doc string just to clarify that the mask_report is a boolean and not the actual mask report.

quipucords/api/deployments_report/view.py
```diff
@@ -64,27 +64,41 @@ def deployments(request, pk=None):
...
-    return {'report_id': report.id,
-            'status': report.status,
-            'report_type': report.report_type,
-            'report_version': report.report_version,
-            'report_platform_id': str(report.report_platform_id),
-            'system_fingerprints': json.loads(report.cached_fingerprints)}
+    system_fingerprints = json.loads(
+        report.cached_fingerprints)
+    if validate_query_param_bool(mask_report):
+        if report.cached_masked_fingerprints:
```
myersCody: Would it be possible to use a `get_object_or_404(report.cached_masked_fingerprints)`? I am not sure, but just wanted to throw it out there.

myersCody: Actually, after thinking about it for a bit and looking at the django [doc](https://docs.djangoproject.com/en/3.0/topics/http/shortcuts/#get-object-or-404) I don't think this possible. You would probably need some type of pk or something. I am gonna go ahead and resolve this one.

quipucords/api/common/util.py
```diff
@@ -112,6 +114,22 @@ def convert_to_boolean(value):
...
+def validate_query_param_bool(param, param_name='mask'):
+    """Validate that the query param is a boolean return the bool.
+
+    :param: param: The query param to evaluate
+    :return The value as a boolean or a validation error
+    """
+    if is_boolean(param):
+        return convert_to_boolean(param)
+    error = {
+        'mask': [_(
```
myersCody: I think you may want to replace `'mask'` here with `param_name`, additionally I think param_name needs to be added to the docstring.

quipucords/api/deployments_report/view.py
abaiken: Done.

quipucords/api/deployments_report/view.py
abaiken: Done.

quipucords/api/common/util.py
abaiken: Done.

PR #2035, opened 2020-01-16
Update regex of hostname to allow _

kholdaway:


quipucords/api/source/tests_source.py
```diff
@@ -96,6 +96,9 @@ def create_expect_400(self, data, expected_response=None):
     def create_expect_201(self, data):
         """Create a source, return the response as a dict."""
         response = self.create(data)
+        if response.status_code != status.HTTP_201_CREATED:
+            print('#' * 1200)
```
abaiken: should we remove these debugging print statements?

PR #2124, opened 2022-01-20
Replace jenkins with Github Actions

bruno-fs:
Key differences with the original build pipeline:
- PostgreSQL image is not packaged
- UI is built from scratch (couldn't find a public tarball with it)
- Added a cache for both UI dependencies and docker layers to speed up future builds


bruno-fs:
Example of the build pipeline running on my fork: https://github.com/bruno-fs/quipucords/releases/tag/0.9.4-beta2

gtanzillo (approved):
Nice job @bruno-fs 👍

PR #2127, opened 2022-02-07
add docker-compose for local development

bruno-fs:


bruno-fs:
@nicolearagao and @gtanzillo I recommend reviewing in a commit per commit basis.
The reason I opened as a draft is for the last commit, adding docker-compose. I'd like some help testing this both on RHEL and macos.

Dockerfile
```diff
@@ -1,25 +1,21 @@
-FROM fedora:26
+FROM redhat/ubi8
 
-RUN dnf -y groupinstall "Development tools" \
-    && dnf -y install python-devel python-tools python3-devel python3-tools sshpass which supervisor procps\
-    && dnf clean all \
-    && rm -rf /var/cache/dnf
+RUN dnf -yq install python39 make openssh-clients glibc-langpack-en &&\
+    dnf update &&\
```
gtanzillo, line 4: @bruno-fs Testing for Mac, this line fails -
```
#6 7.533 Updating Subscription Management repositories.
#6 7.534 Unable to read consumer identity
#6 7.537
#6 7.537 This system is not registered with an entitlement server. You can use subscription-manager to register.
```
Because Macs can't be registered and there is no `subscription-manager`.

I commented it out locally and it built cleanly for me

bruno-fs: Mmmm maybe we can drop this line and just trust ubi8 packages are fresh enough

nicolearagao (approved):
@bruno-fs will you submit the changes to the README file as well ? You could leave that out. I'll open a PR for the complete one.

bruno-fs:
@nicolearagao 
> @bruno-fs will you submit the changes to the README file as well ? You could leave that out. I'll open a PR for the complete one.
I'll leave the changes to README out. Will only submit the changes to docker-compose (once we get it up and running on a mac)

@gtanzillo I changed the networking strategy here. Whenever you can please test this again

bruno-fs:
rebased against recent upstream changes

Dockerfile
bruno-fs: this was addressed on #2129

docker-compose.yml
```diff
@@ -0,0 +1,36 @@
...
+      QPC_SERVER_USERNAME: admin
+      QPC_SERVER_PORT: "9443"
+      USE_SUPERVISORD: "false"
+
+    volumes:
+      - .:/app
+      - ./deploy:/deploy
+    depends_on:
+      - qpc-db
+    network_mode: "host"
```
gtanzillo, line 22: It looks like "host" mode is not supported on Mac according to https://docs.docker.com/network/host/. Specifically, this statement -
>The host networking driver only works on Linux hosts, and is not supported on Docker Desktop for Mac, Docker Desktop for Windows, or Docker EE for Windows Server.

However, this is not an issue for me, since Mac is not a supported platform.

bruno-fs: Mmm we could go back to the slirp4netns, then. Maybe [installing it](https://formulae.brew.sh/formula/slirp4netns) on mac would be enough. If not, its better than host network for our purposes.

bruno-fs: @gtanzillo would you prefer I rebase with autosquash before or after your review?

gtanzillo (approved):
LGTM 👍

Issue #2128, opened 2022-02-07
Feature: object holding currently activated feature flags


bruno-fs:
## Feature description:
Add a helper class with attributes or methods returning booleans describing whether features are enabled (`True`) or not (`False`).

### Is your feature request related to a problem?
No

### Describe the solution you'd like
- As other configurations used by QPC, feature flags will be at least initially) controlled by env vars.
- These env vars SHOULD start with the prefix `QPC_FEATURE_<FEATURE_SHORTNAME>` and its value can be 0 or 1
- On the application side this will be available as either a method, attribute or property of a helper class named <feature_shortname> which returns either `True` or `False`
- These configs should be loaded once during application startup and accessed through django registry (`from django.conf import settings`)

___

## Acceptance Criteria:

- [ ] Add tests confirming helper class proper behaviour

___

PR #2131, opened 2022-02-14
Creates class FeatureFlag to handle when new features will be added

nicolearagao:
Creates class FeatureFlag to administrate environment variable and status(ENABLED-DISABLED).
Filters feature flags in function get_env_flags_from_environ(), making sure only variables that follow the pattern are added or updated.
Creates unit tests for the class and helper fuction.

### DIFF
--- a/quipucords/quipucords/featureflag.py
+++ b/quipucords/quipucords/featureflag.py
@@ -1,39 +1,49 @@
-import os
+"""Filters feature flags from systen env variables."""
+
 import logging
+import os
 
 DEFAULT_FEATURE_FLAGS_VALUES = {
-    'OVERALL_STATUS': False}  # default values for feature flags that are already implemented
-VALID_VALUES_FOR_ENV_VARIABLES = ["0", "1"]
+    'OVERALL_STATUS': False}  # default values for feature flags that
+# are already implemented
+
+VALID_VALUES_FOR_ENV_VARIABLES = ['0', '1']
 
 # Get an instance of a logger
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
 
 def get_feature_flags_from_env():
-    """Filters feature flags from environment variables"""
-
+    """Filter feature flags from environment variables."""
     global DEFAULT_FEATURE_FLAGS_VALUES
     global VALID_VALUES_FOR_ENV_VARIABLES
     for key, value in os.environ.items():
-        if key.upper().startswith("QPC_FEATURE_"):
-            feature = key.upper().replace("QPC_FEATURE_", "")
+        if key.upper().startswith('QPC_FEATURE_'):
+            feature = key.upper().replace('QPC_FEATURE_', '')
             try:
                 if value in VALID_VALUES_FOR_ENV_VARIABLES:
                     DEFAULT_FEATURE_FLAGS_VALUES[feature] = bool(int(value))
             except ValueError:
-                logger.info(f"{value} can`t be converted to int, verify your environment variables.")
+                logger.info(' %s can`t be converted to int, verify your '
+                            'environment variables.', value)
                 continue
     return DEFAULT_FEATURE_FLAGS_VALUES
 
 
 class FeatureFlag:
+    """Handle environment variables and its status."""
 
     def __init__(self, initial_data):
+        """Attributes and values are generated dynamically.
+
+        Attributes keys-values are received in initial_data.
+        """
         for key, value in initial_data.items():
             setattr(self, key, value)
 
     def is_feature_active(self, feature_name):
+        """Return attribute value."""
         try:
             return getattr(self, feature_name)
         except AttributeError:
-            raise ValueError(f"{feature_name=} is not a valid input.")
+            raise ValueError(f'{feature_name=} is not a valid input.')
--- a/quipucords/quipucords/tests_feature_flags.py
+++ b/quipucords/quipucords/tests_feature_flags.py
@@ -1,20 +1,23 @@
+"""Test the FeatureFlag class and helper function."""
+
 import os
 import unittest
 from unittest import mock
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


quipucords/quipucords/tests_feature_flags.py
```diff
@@ -0,0 +1,154 @@
...
+        'TEST') is True
+    assert setup_feature_flag_instance_for_tests.is_feature_active(
+        'OVERALL_STATUS') is False
+    with pytest.raises(ValueError):
+        assert setup_feature_flag_instance_for_tests.is_feature_active(
+            'FALSE_ATTRIBUTE') is True
+
+
+if __name__ == '__main__':
+    unittest.main()
```
bruno-fs, line 154: Remove these lines as well

quipucords/quipucords/tests_feature_flags.py
```diff
@@ -0,0 +1,154 @@
...
+
+def test_is_feature_active(setup_feature_flag_instance_for_tests):
+    """Tests method is_feature_active."""
+    assert setup_feature_flag_instance_for_tests.is_feature_active(
+        'TEST') is True
+    assert setup_feature_flag_instance_for_tests.is_feature_active(
+        'OVERALL_STATUS') is False
+    with pytest.raises(ValueError):
+        assert setup_feature_flag_instance_for_tests.is_feature_active(
+            'FALSE_ATTRIBUTE') is True
```
bruno-fs, line 150: This assertion will never run if `ValueError` is properly raised.
```suggestion
        setup_feature_flag_instance_for_tests.is_feature_active('FALSE_ATTRIBUTE')
```

quipucords/quipucords/featureflag.py
```diff
@@ -0,0 +1,49 @@
...
+    for key, value in os.environ.items():
+        if key.upper().startswith('QPC_FEATURE_'):
+            feature = key.upper().replace('QPC_FEATURE_', '')
+            try:
+                if value in VALID_VALUES_FOR_ENV_VARIABLES:
+                    DEFAULT_FEATURE_FLAGS_VALUES[feature] = bool(int(value))
+            except ValueError:
+                logger.info(' %s can`t be converted to int, verify your '
+                            'environment variables.', value)
+                continue
```
bruno-fs, line 29: Another thing that comes from this if: invalid values are silently ignored.

As an user I'd rather know I did something wrong (either with an warning or just throwing an error preventing the application to start). 

(I guess we forgot to discuss what should be the proper behaviour of this class when invalid values are passed).

Here's my suggestion:

```suggestion
            feature_flags = DEFAULT_FEATURE_FLAGS_VALUES.copy()
            try:
                feature_flags[feature] = bool(int(value))
            except ValueError:
                raise ValueError(f'{value} is invalid for feature '{feature}'. Verify your environment variables.')
```

quipucords/quipucords/featureflag.py
```diff
@@ -0,0 +1,49 @@
...
+    return DEFAULT_FEATURE_FLAGS_VALUES
+
+
+class FeatureFlag:
+    """Handle environment variables and its status."""
+
+    def __init__(self, initial_data):
```
bruno-fs, line 36: This is optional: I'd change how `__init__` handles initial data. Instead of expecting `initial_data` it could either

1. replace it with "parser function" that returns the dict with (pointing as default to `get_feature_flags_from_env`)
2. refactor `get_feature_flags_from_env` as a `classmethod`* that would be called inside `__init__` 

 On either case `initial_data` would be set inside `__init__` calling either the parser function or the proper method.

quipucords/quipucords/tests_feature_flags.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+@pytest.mark.parametrize(
+    'env_name,env_value,feature_name,feature_value',
+    (('QPC_FEATURE_OVERALL_STATUS', '1', 'OVERALL_STATUS', True),
+     ('QPC_FEATURE_TEST', '0', 'TEST', False)
+     )
+)
+def test_if_value_for_env_default_list_gets_updated(
+        env_name, env_value, feature_name, feature_value):
```
bruno-fs, line 44: [OPTIONAL] This is an overall comment on the style here. I'm not a huge fan on discussing code style but this still bothers me on how parenthesis are closed. Here's my suggestion:

```suggestion
@pytest.mark.parametrize(
    "env_name,env_value,feature_name,feature_value",
    (
        ("QPC_FEATURE_OVERALL_STATUS", "1", "OVERALL_STATUS", True),
        ("QPC_FEATURE_TEST", "0", "TEST", False),
    ),
)
def test_if_value_for_env_default_list_gets_updated(
    env_name, env_value, feature_name, feature_value
):
```

In a future PR I'll add black to the project so we won't need to discuss this :stuck_out_tongue:

quipucords/quipucords/tests_feature_flags.py
```diff
@@ -0,0 +1,154 @@
...
+        assert feature_name in dict_with_test_flags
+        assert dict_with_test_flags[feature_name] is feature_value
+
+
+@pytest.fixture
+def setup_feature_flag_instance_for_tests():
+    """Set up instance of FeatureFlag class for tests."""
+    with mock.patch.dict(os.environ, {'QPC_FEATURE_TEST': '1'}):
+        dict_with_test_flags = get_feature_flags_from_env()
+        feature_flag_instance = FeatureFlag(dict_with_test_flags)
```
bruno-fs, line 124: If you change `FeatureFlag.__init__` method you can replace this lines with

```suggestion
        feature_flag_instance = FeatureFlag()
```

bruno-fs (changes requested):
This is a good proposal overall, but still needs some changes. One of the requirements of the desing (#2128) wasn't implemented:

> These configs should be loaded once during application startup and accessed through django registry (`from django.conf import settings`)

Feel free to ignore comments marked as "optional".

PR #2133, opened 2022-02-23
Improve linting

bruno-fs:
- Update linters versions
- Add black/isort checks
- Enable most checks for pylint/flake/black/isort but only for newer code. Older code will be ignored until chaged.


gtanzillo (approved):
Looks good, thanks @bruno-fs 👍

nicolearagao (approved):
awesome @bruno-fs

PR #2136, opened 2022-02-27
Biz 400 / issue 2064- adds overall status fact to subman role

nicolearagao:
- Import FeatureFlags into scanjob as extra vars. Adds feature flag as condition for the gathering overall status fact.
- Addition of method as_dict for FeatureFlag class as Ansible extravars expects objects to be simpler than an instance of the class.
- Unit tests for new method.
- Adds subscription-management overall status fact to subman role.

Manually tested on:
-AWS machines(RHEL6, RHEL7, RHEL8)
-Vagrant(CentOS6, CentOS7, CentOS8)

### DIFF
--- a/quipucords/scanner/network/inspect.py
+++ b/quipucords/scanner/network/inspect.py
@@ -14,6 +14,7 @@ import os.path
 
 import ansible_runner
 from ansible_runner.exceptions import AnsibleRunnerException
+from django.conf import settings
 
 from api.credential.serializer import CredentialSerializer
 from api.models import (ScanJob,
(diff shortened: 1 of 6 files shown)


quipucords/scanner/network/inspect.py
```diff
@@ -78,16 +78,16 @@ def run(self, manager_interrupt):
 
         self.connect_scan_task = self.scan_task.prerequisites.first()
         if self.connect_scan_task.status != ScanTask.COMPLETED:
-            error_message = 'Prerequisites scan task %d failed.' %\
-                self.connect_scan_task.sequence_number
+            error_message = 'Prerequisites scan task %d failed.' % \
+                            self.connect_scan_task.sequence_number
```
bruno-fs: Now it makes sense why linters were looking to change some strings on this file to be reformatted.

I'm not against commits making changes to formatting, but I believe this kind of change should live in a separate commit. 
That said, don't need to change this particular commit.

bruno-fs (approved):
LGTM :+1:

gtanzillo (approved):
LGTM 👍

PR #2139, opened 2022-03-09
Molecule first steps 

nicolearagao:
-  Molecule installation 
-  Init Molecule scenario for the date role using podman driver 
-  Configuration of two test instances: rhel8 and centos8
-  adds Molecule packages to dev-requirements.txt

obs: Please take a look at the google doc shared for further explanations. 

Updates on March 14th:
- Converge step demanded a few modifications (changing raw to appropriate module, creation of role utils for shared facts)

### DIFF
--- a/quipucords/scanner/network/runner/roles/date/molecule/default/molecule.yml
+++ b/quipucords/scanner/network/runner/roles/date/molecule/default/molecule.yml
@@ -1,0 +1,13 @@
+---
+dependency:
+  name: galaxy
+driver:
+  name: podman
+platforms:
+  - name: instance
+    image: quay.io/centos/centos:stream8
+    pre_build_image: true
+provisioner:
+  name: ansible
+verifier:
+  name: ansible
--- a/quipucords/scanner/network/runner/roles/date/molecule/default/molecule.yml
+++ b/quipucords/scanner/network/runner/roles/date/molecule/default/molecule.yml
@@ -4,9 +4,25 @@ dependency:
 driver:
   name: podman
 platforms:
-  - name: instance
-    image: quay.io/centos/centos:stream8
+  - name: rhel8
+    image: registry.access.redhat.com/ubi8/ubi-init
+    tmpfs:
+      - /run
+      - /tmp
+    volumes:
+      - /sys/fs/cgroup:/sys/fs/cgroup:ro
+    capabilities:
+      - SYS_ADMIN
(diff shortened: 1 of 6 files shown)


quipucords/scanner/network/runner/roles/date/molecule/default/molecule.yml
```diff
@@ -11,3 +11,7 @@ provisioner:
   name: ansible
 verifier:
   name: ansible
+#lint: |
+#  set -e
+#  yamllint .
+#  ansible-lint .
```
bruno-fs, line 33: :thinking: why is this commented? :thinking:

bruno-fs (commented):
Looks promising :)

quipucords/scanner/network/runner/roles/date/molecule/default/molecule.yml
```diff
@@ -0,0 +1,13 @@
+---
+dependency:
+  name: galaxy
+driver:
+  name: podman
+platforms:
+  - name: instance
+    image: quay.io/centos/centos:stream8
```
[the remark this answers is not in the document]
nicolearagao: This is the boilerplate code created when you init a default scenario. I am not sure why they use this image,but they use it regardless the driver you choose. You can modify it later and that's what I did when I addeed ubi.

quipucords/scanner/network/runner/roles/date/molecule/default/molecule.yml
nicolearagao: Running ansible lint on this project raised a crazy amount of errors. Some really easy to fix, other's a bit more complex like: "replace ignore_errors:yes for failed_when:{{reason for failing}}".Do we have to map out errors for every task? How many different errors are there? Should we ignore this specific rule? All unanswered questions that I plan to bring up to the team when are a bit more advanced on the process of implementing this.

ruda: * Enable yamllint first and see what it catches, see if it worths running now.
* * Is there any configuration file for ansible-lint so we can make a config to avoid the linter being rather grumpy at this time?

nicolearagao: I ran yamllint and it pointed 0 errors. Yes, we can config what we want to enable/disable on ansible-lint.

PR #2140, opened 2022-03-16
Prepare for downstream

bruno-fs:
Most of these changes are required for quipucords to play nice with cachito
Also change the mechanism to detect version (move from build time - on github actions - to install time)


gtanzillo (approved):
LGTM, thanks for all the hard work on this, @bruno-fs

PR #2148, opened 2022-04-20
[DISCOVERY-94] Fix/scan manager unhandled

bruno-fs:


quipucords/conftest.py
```diff
@@ -0,0 +1,63 @@
...
+# pylint: disable=import-outside-toplevel
+from contextlib import suppress
+from unittest.mock import patch
+
+import pytest
+
+
+def _kill_scan_manager(manager_instance):
+    for job in manager_instance.scan_queue:
+        with suppress(AttributeError):
```
nicolearagao, line 25: Is there a difference between using contextlib.suppress and a try/catch besides readability?

quipucords/conftest.py
```diff
@@ -0,0 +1,63 @@
...
+    from scanner import manager
+
+    scan_manager_instance = manager.Manager()
+    with (
+        patch.object(manager, "sleep"),
+        patch.object(manager, "SCAN_MANAGER", scan_manager_instance),
+    ):
+        yield scan_manager_instance
+
+    _kill_scan_manager(scan_manager_instance)
```
nicolearagao, line 53: so after the yield we have the __exit__ and you called the func _kill_scan_manager to do it. Right?

quipucords/conftest.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+@pytest.fixture(autouse=True)
+def _patch_scan_manager(scan_manager):
+    with (
+        patch.object(scan_manager.__class__, "__call__", _return_itself),
+        patch("api.signal.scanjob_signal.manager.Manager", scan_manager),
+        patch("api.signal.scanjob_signal.manager.SCAN_MANAGER", scan_manager),
+    ):
+        yield
```
nicolearagao, line 63: What happens after the with block finishes on this one?

quipucords/conftest.py
bruno-fs: nope. this is exactly the same as 
```
try:
   job.kill()
except AttributeError:
    pass
```
just less verbose

quipucords/conftest.py
bruno-fs: in a pytest fixture, think about everything before yield as a setup (`TestCase.setUp` equivalent) and what comes after as a teardown (`TestCase.tearDown`)

quipucords/conftest.py
bruno-fs: you mean the context with the patches? if so, everything goes back to they original behaviour before the patching

nicolearagao: ok!

nicolearagao (approved):
LGTM =)

gtanzillo (approved):

PR #2150, opened 2022-05-05
Feature/collect all raw facts for network scan

bruno-fs:
Always collect raw fact names for network scan - even if they have a null value.


bruno-fs:
@nicolearagao @gtanzillo please ignore https://github.com/quipucords/quipucords/pull/2150/commits/f1fc2b45ee63a701374de089b20ca175014eae1e for review purposes. I'll open a separate PR with just it

quipucords/scanner/network/runner/roles/redhat_release/tasks/main.yml
```diff
@@ -4,25 +4,17 @@
   set_fact:
     internal_host_started_processing_role: "redhat_release"
 
-- name: initialize redhat_release facts
```
nicolearagao, line 7: Won't that cause any impacts on the way the report is printed in case there's no - "'stdout_lines' in internal_redhat_release"?

quipucords/scanner/network/utils.py
```diff
@@ -102,3 +103,22 @@ def expand_hostpattern(hostpattern):
...
+def _yaml_load(path_obj):
+    return yaml.load(path_obj.open(), Loader=yaml.SafeLoader)
+
+
+def collect_all_fact_names():
+    """Collect all fact names set on ansible playbooks."""
+    network_runner_path = settings.BASE_DIR / "scanner/network/runner"
+    for playbook in network_runner_path.rglob("roles/*/tasks/main.yml"):
+        commands = _yaml_load(playbook)
+        for command in commands:
```
nicolearagao, line 117: Maybe replace commands with tasks?

quipucords/scanner/network/runner/roles/redhat_release/tasks/main.yml
```diff
@@ -4,25 +4,17 @@
...
-  set_fact:
-    redhat_release_name: ''
-    redhat_release_version: ''
-    redhat_release_release: ''
-  ignore_errors: yes
-
 - name: gather internal_redhat_release fact
   raw: rpm -q --queryformat "%{NAME}\n%{VERSION}\n%{RELEASE}\n" --whatprovides redhat-release
   register: internal_redhat_release
```
nicolearagao, line 9: why did you change it to true instead of yes ?
It's a bit confussing but I guess Ansible standard is yes instead of true.
https://docs.ansible.com/ansible/latest/user_guide/playbooks_error_handling.html#

quipucords/fingerprinter/jboss_brms.py
```diff
@@ -76,6 +76,14 @@ def classify_version_string(version_string):
     return None
 
 
+def _get_val_or_empty_list(data_dict, key):
```
nicolearagao, line 79: Is this going to be called whenever we don't scan middleware ?

quipucords/scanner/network/test_sanity_check.py
```diff
@@ -0,0 +1,58 @@
...
+    inspect_data = _yaml_load(inspect_yml_path)
+    assert isinstance(inspect_data, list)
+    assert len(inspect_data) == 1
+    assert isinstance(inspect_data[0], dict)
+    assert (
+        inspect_data[0]["roles"][-1] == "host_done"
+    ), "host_done SHOULD be the last role"
+
+
+def test_check_all_roles_are_included(inspect_yml_path, path_to_playbooks):
```
nicolearagao, line 47: This was pretty smart.

quipucords/scanner/network/runner/roles/redhat_release/tasks/main.yml
bruno-fs: Mmm I think it will. Nice catch!

quipucords/scanner/network/runner/roles/redhat_release/tasks/main.yml
```diff
@@ -4,25 +4,17 @@
...
 
 - name: set redhat_release facts based on internal_redhat_release
   set_fact:
     redhat_release_name: "{{ internal_redhat_release['stdout_lines'][0] }}"
     redhat_release_version: "{{ internal_redhat_release['stdout_lines'][1] }}"
     redhat_release_release: "{{ internal_redhat_release['stdout_lines'][2] }}"
-  ignore_errors: yes
+  ignore_errors: true
   when:
     - "'stdout_lines' in internal_redhat_release"
```
bruno-fs, line 20: @nicolearagao do you know if ansible `when` short circuits? If so I think we can still remove the initial setup and add this

```suggestion
    - internal_redhat_release is defined
    - "'stdout_lines' in internal_redhat_release"
    - "internal_redhat_release['stdout_lines']|length == 3"
```

PS: I shouldn't have removed the length check. probably did this by mistake

nicolearagao: It's not short circuit evaluation. 
Why don't you do something like this? 
when: result.stdout | default('') == 'foo'

quipucords/fingerprinter/jboss_brms.py
bruno-fs: `detect_jboss_brms` will be called whether we scan middleware or not.
then it would use default dict getter for many keys with default set to `list()` or `set()`

This will be used later based on the premise that even if nothing is found an iterable will be there.
Changing the facts to always be filled with `None` breaks this behavior because the value `None` will take precedence over the default provided to the getter.

In other words...
```python
data_dict1 = {}
manifest_mfs1 = data_dict1.get(JBOSS_BRMS_MANIFEST_MF, set())
assert type(manifest_mfs1) == type(set())

data_dict2 = {JBOSS_BRMS_MANIFEST_MF: None}
manifest_mfs2 = data_dict2.get(JBOSS_BRMS_MANIFEST_MF, set())
assert type(manifest_mfs2) == type(None)
```

quipucords/scanner/network/runner/roles/redhat_release/tasks/main.yml
bruno-fs: Not sure if I got it. Would this work if `internal_redhat_release` is `undefined`?

```suggestion
    - internal_redhat_release.stdout | default("\n\n\n")
```

quipucords/scanner/network/runner/roles/redhat_release/tasks/main.yml
nicolearagao: awesome!!

quipucords/scanner/network/runner/roles/redhat_release/tasks/main.yml
nicolearagao: I apologize, my suggestion wasn't clear.  if internal_redhat_release is undefined, it means that either this is not a red hat machine or that there was a problem while collecting this information. 
For the further I think it's still worth it to have a default value if anything goes wrong. I don't think calling set_fact twice to do it is the most efficient way. 
You could do:
set_fact:
    redhat_release_name: "{{ internal_redhat_release['stdout_lines'][0] | default(" ") }}"
    redhat_release_version: "{{ internal_redhat_release['stdout_lines'][1] | default(" ")}}"
    redhat_release_release: "{{ internal_redhat_release['stdout_lines'][2] | default(" ")}}"
The reason why this won't blow up if internal_redhat_release is undefined is due to the flag ignore_errors. In any other scenario it would.

quipucords/scanner/network/runner/roles/redhat_release/tasks/main.yml
```diff
@@ -4,25 +4,17 @@
...
-    - "internal_redhat_release['stdout_lines']|length == 3"
```
nicolearagao, line 28: Why the length would be longer than 3? Probably if it raised an error, therefore o failed_when would be much more efficient than ignore_errors. 
I guess this calls for a refactor. hehe =P

bruno-fs: It's called "defensive programming". If the output don't have 3 lines, something is wrong.

quipucords/fingerprinter/jboss_fuse.py
```diff
@@ -71,6 +68,10 @@ def get_version(eap_version):
     return new_versions
 
 
+def _get_val_or_empty_list(data_dict, key):
```
gtanzillo, line 71: Maybe put this in a common file with the other helper methods to DRY it up

gtanzillo (commented):
@bruno-fs I did a first pass on this and it looks good. Since there are a lot of changes here, perhaps it would be good for you to do a walkthrough of the change.

quipucords/scanner/network/utils.py
bruno-fs: change to glob

bruno-fs:
we need a camayoc run before merging

bruno-fs:
we need a camayoc run before merging

quipucords/scanner/network/runner/roles/redhat_release/tasks/main.yml
bruno-fs: Setting a fact to " " can be problematic because the expected value for "empty" facts is "".
I could apply this suggestion with default to "" but I'm not confident this won't cause errors - it depends on 2 premises that might not be fulfilled (stdout_lines exists and has at least 3 lines).

Now I understood why this was implemented the way it was and I think I prefer this approach now. What do you think?

nicolearagao (approved):
LGTM. Great job!

quipucords/fingerprinter/jboss_fuse.py
bruno-fs: Oh I didn't comment on this but I added a DRY solution in a project level `utils` module.

gtanzillo (approved):
Looks great! Thanks @bruno-fs

quipucords/fingerprinter/jboss_fuse.py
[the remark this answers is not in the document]
bruno-fs: I replaced this one with the [default_getter](https://github.com/quipucords/quipucords/blob/96b821a58a7291c883d007c49e86f49418622bfb/quipucords/utils/default_getter.py)

bruno-fs: Replaced this function with the helper "default_getter" https://github.com/quipucords/quipucords/blob/96b821a58a7291c883d007c49e86f49418622bfb/quipucords/fingerprinter/jboss_fuse.py#L84-L101

bruno-fs:
# CAMAYOC REPORT 

## API tests
```
(venv) [root@discovery-qpc-new camayoc]# make test-qpc-api -e PYTEST_OPTIONS='-vvv -ra --disable-warnings'
========================================================================================================= short test summary info ==========================================================================================================
SKIPPED [1] camayoc/tests/qpc/api/v1/reports/test_reports.py:319: Skipping sat6 run until Quipucords Issue #2039 is resolved                                                                                                               
SKIPPED [3] camayoc/tests/qpc/api/v1/scanjobs/test_pause_cancel_restart.py:23: Test is flaky. Skipping until Quipucords Issue #2040 resoloved                                                                                              
SKIPPED [3] camayoc/tests/qpc/api/v1/scanjobs/test_pause_cancel_restart.py:53: Test is flaky. Skipping until Quipucords Issue #2040 resoloved                                                                                              
SKIPPED [2] camayoc/tests/qpc/api/v1/scanjobs/test_run_scanjobs.py:31: Test is flaky. Skipping until Quipucords Issue #2040 resoloved                                                                                                      
XFAIL camayoc/tests/qpc/api/v1/reports/test_reports.py::test_report_content_consistency
  reason: Config file does not have dependent scan "rhel-7".
XFAIL camayoc/tests/qpc/api/v1/reports/test_reports.py::test_merge_reports_from_scanjob
  reason: Config file does not have dependent scans "non-rhel" or "rhel-7".
==================================================================================== 238 passed, 9 skipped, 2 xfailed, 7 warnings in 364.07s (0:06:04) =====================================================================================

```

PR #2153, opened 2022-05-27
Feature/always collect fingerprints netscan 103

bruno-fs:
Always collect fingerprint values (even if null)


quipucords/tests/integration/test_network_scan.py
```diff
@@ -225,6 +228,48 @@ def fingerprint_fact_map():
...
+        "vm_state",
+        "vm_uuid",
+    }
+
+
+def test_sanity_check_raw_fact_matches(
+    expected_network_scan_facts, fingerprint_fact_map, unexpected_fingerprints
+):
+    """Ensure raw facts mapped to fingerprint facts match known facts."""
+    raw_facts_for_fingerprints = reduce(
```
nicolearagao, line 260: I stumbled upon this [article](https://www.artima.com/weblogs/viewpost.jsp?thread=98196) written by Guido saying that instead of using reduce(), the more "pythonic" way is to implement an accumulating function the old fashioned way. 
I honestly think reduce() is easier to read and "cleaner", but thought about sharing this with you.

quipucords/fingerprinter/task.py
```diff
@@ -87,12 +88,15 @@
...
 # keys are in reverse order of accuracy (last most accurate)
 # (date_key, date_pattern)
-RAW_DATE_KEYS = \
-    [('date_yum_history', ['%Y-%m-%d']),
-     ('date_filesystem_create', ['%Y-%m-%d']),
-     ('date_anaconda_log', ['%Y-%m-%d']),
-     ('registration_time', ['%Y-%m-%d %H:%M:%S', '%Y-%m-%d %H:%M:%S %z']),
-     ('date_machine_id', ['%Y-%m-%d'])]
+RAW_DATE_KEYS = OrderedDict(
+    [
```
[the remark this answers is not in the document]
bruno-fs: dict being currently ordered is more a happy accident than  a feature we should count on. That might change in the future. If you rely on order, use `OrderedDict`.

quipucords/tests/integration/test_network_scan.py
```diff
@@ -225,6 +228,48 @@ def fingerprint_fact_map():
...
+        or_operator,
+        (fact_expander(fact) for fact in fingerprint_fact_map.values()),
+    )
```
bruno-fs, line 263: Let's evaluate this together.
This might have a better readability considering the reader wouldn't need to know `reduce` and `or_` (aliased as `or_operator`

```suggestion
    raw_facts_for_fingerprints = set()
    for fact in fingerprint_fact_map.values():
        raw_facts_for_fingerprints |= fact_expander(fact)
```

quipucords/tests/integration/test_network_scan.py
bruno-fs: Oh, I remember this article. There's a reason for "hiding" reduce under functools (back in python 2 era reduce didn't need import).

I can rewrite this w/o reduce and we check which version has better readability

quipucords/tests/integration/test_network_scan.py
bruno-fs: In both cases it is required to know what the or operator does on a set but I'm now more inclined to think Guido was right XD

quipucords/fingerprinter/task.py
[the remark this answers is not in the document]
bruno-fs: Mmm what I said isn't true anymore. When this order came it wasn't something to be relied upon
https://docs.python.org/3/whatsnew/3.6.html#new-dict-implementation

I didn't knew that this got "oficial" status on 3.7.
> Changed in version 3.7: Dictionary order is guaranteed to be insertion order. This behavior was an implementation detail of CPython from 3.6.
https://docs.python.org/3/library/stdtypes.html#mapping-types-dict

Since we don't rely on any of the other features from `OrderedDict` we can use `dict`, then.

quipucords/tests/integration/test_network_scan.py
nicolearagao: I do prefer the one without reduce()!

quipucords/fingerprinter/task.py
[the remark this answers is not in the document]
nicolearagao: Let's go with dict then.

nicolearagao (approved):
Awesome work! =)

gtanzillo (approved):
LGTM 👍 Thanks @bruno-fs

PR #2161, opened 2022-06-01
Feature/always create null data vcenter

nicolearagao:
Creates template for Vcenter facts which contain all facts regardless if they are found or not. Their values are set to None and later updated when fact is found.

- Modifies results_template name and test_all_facts_with_null_value name
- Creates classes of RAW FACTS
- Creates raw_facts_template to replace empty facts list
- Use Alias name instead of raw_facts_template()
- Create Test to make sure fields are being filled with null if fact is not present

### DIFF
--- a/quipucords/fingerprinter/task.py
+++ b/quipucords/fingerprinter/task.py
@@ -57,6 +57,7 @@ from fingerprinter.jboss_fuse import detect_jboss_fuse
 from fingerprinter.jboss_web_server import detect_jboss_ws
 from fingerprinter.utils import strip_suffix
 from scanner.task import ScanTaskRunner
+from utils import default_getter
 
 # pylint: disable=too-many-lines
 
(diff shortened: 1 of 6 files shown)


quipucords/fingerprinter/task.py
```diff
@@ -1369,9 +1369,10 @@ def _process_vcenter_fact(self, source, fact):
         if vcenter_os_release != '':
             rhel_os_releases = ['red hat enterprise linux', 'rhel']
             for rhel_release in rhel_os_releases:
-                if rhel_release in vcenter_os_release.lower():
-                    is_redhat = True
-                    break
+                if vcenter_os_release is not None:
```
bruno-fs, line 1372: you can avoid changing this whole block by replacing the call on line 1367

```
vcenter_os_release = default_getter(fact, 'vm.os', '')
```

bruno-fs (commented):
Awesome work! :clap: 

I'm suggesting 2 small changes and would like to know what you think about refactoring proposal on the constants

gtanzillo (approved):
Looks great, awesome work!

bruno-fs (approved):
LGTM :+1: :heavy_check_mark:

gtanzillo (approved):

bruno-fs (approved):

PR #2163, opened 2022-06-06
Refactor/change list for set

nicolearagao:
Improvement on search.


nicolearagao:
@bruno-fs I tried to solve this using one of the infix operators, but couldn't really find a solution because what we want to know is if any of the set elements is a substring of the vcenter_os_release. Please point me in the right direction if this was not what you expected.

PR #2165, opened 2022-06-13
Feature/always create null data satellite

nicolearagao:
Creates template for Satellite facts which contain all facts regardless if they are found or not. Their values are set to None and later updated when fact is found.

- Creates raw_facts_templates to replace empty facts list
- adds raw_facts_template to satellite five/six config
- Creates test to function raw_facts_template()

### DIFF
--- a/quipucords/scanner/satellite/utils.py
+++ b/quipucords/scanner/satellite/utils.py
@@ -14,20 +14,17 @@ import logging
 import ssl
 import xmlrpc.client
 
-from api.vault import decrypt_data_as_unicode
-
-from quipucords import settings
-
 import requests
-
 from rest_framework import status as codes
 
-
-from scanner.satellite.api import (SATELLITE_VERSION_5,
-                                   SATELLITE_VERSION_6,
-                                   SatelliteAuthException,
-                                   SatelliteException)
-
+from api.vault import decrypt_data_as_unicode
+from quipucords import settings
+from scanner.satellite.api import (
+    SATELLITE_VERSION_5,
+    SATELLITE_VERSION_6,
+    SatelliteAuthException,
+    SatelliteException,
+)
 
 # Get an instance of a logger
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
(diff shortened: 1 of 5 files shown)


quipucords/scanner/satellite/utils.py
```diff
@@ -256,3 +256,29 @@ def validate_task_stats(task):
...
+        "os_name",
+        "os_release",
+        "os_version",
+        "packages_out_of_date",
+        "registration_time",
+        "uuid",
+        "virt_type",
+        "last_checkin_time",
+    )
+    return {fact_name: None for fact_name in sorted(satellite_raw_facts)}
```
bruno-fs, line 284: If we keep this design I would drop the `sorted` function and sort the raw facts from line 264 to 282

bruno-fs (approved):
Amazing!

I'd just make one small change

quipucords/scanner/satellite/utils.py
nicolearagao: why would you do that instead?

bruno-fs: I think ordered lists often have better readability

aaaand if its already ordered than you wouln't need the `sorted` function.

gtanzillo (approved):
Looks good! Nice work @nicolearagao!

quipucords/scanner/satellite/utils.py
```diff
@@ -155,15 +152,16 @@ def status(scan_task):
     """
     try:
         return _status6(scan_task)
-    except SatelliteException as sat_error:
-        message = 'Satellite 6 status check failed with error:' \
-            ' "%s". Attempting Satellite 5.' % sat_error
+    except SatelliteException as sat_error:  # pylint: disable=unused-variable # noqa F841
```
bruno-fs, line 155: instead of disabling this check consider removing the variable

(for the future I'd rather have the full traceback instead of only the message that follows)

quipucords/scanner/satellite/utils.py
```diff
@@ -155,15 +152,16 @@ def status(scan_task):
...
+            "Satellite 6 status check failed with error:"
+            ' f"{sat_error}". Attempting Satellite 5.'
+        )
         scan_task.log_message(message)
     try:
         return _status5(scan_task)
-    except SatelliteException as sat_error:
-        message = 'Satellite 5 status check failed with error:' \
-            ' "%s".' % sat_error
+    except SatelliteException as sat_error:  # pylint: disable=unused-variable  # noqa F841
```
bruno-fs, line 163: Same here. Consider removing `sat_error` since its unused.

bruno-fs (approved):
Added a comment on one small thing I missed on my first review but I'll keep this approved.

quipucords/scanner/satellite/utils.py
nicolearagao: done

quipucords/scanner/satellite/utils.py
nicolearagao: done

PR #2166, opened 2022-06-14
Fix/vcenter inconsistent number hosts discovery 90

nicolearagao:
Collects config.template attribute and set it as fact
Modifies vcenter tests to expect vm.is_template as a fact

*new fact was added to data dictionary

### DIFF
--- a/quipucords/scanner/vcenter/inspect.py
+++ b/quipucords/scanner/vcenter/inspect.py
@@ -188,6 +188,8 @@ class InspectTaskRunner(ScanTaskRunner):
                     facts[VcenterRawFacts.LAST_CHECK_IN] = now
             elif prop.name == 'summary.guest.hostName':
                 facts[VcenterRawFacts.DNS_NAME] = prop.val
+            elif prop.name == "config.template":
+                facts[VcenterRawFacts.TEMPLATE] = prop.val
             elif prop.name == 'summary.config.guestFullName':
                 facts[VcenterRawFacts.OS] = prop.val
             elif prop.name == 'summary.config.memorySizeMB':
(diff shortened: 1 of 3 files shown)


quipucords/scanner/vcenter/inspect.py
```diff
@@ -339,15 +341,16 @@ def _property_set(self):
...
-                'summary.guest.hostName',
-                'summary.runtime.powerState',
-                'summary.config.guestFullName',
-                'summary.config.memorySizeMB',
-                'summary.config.numCpu',
-                'summary.config.uuid'
+                "guest.net",
+                "name",
+                "runtime.host",
+                "config.template",
```
bruno-fs, line 347: I can see you just added "config.template" here. Don't need to change this commit but consider in the future separating one commit reformatting old code and another one adding something new.

quipucords/scanner/vcenter/tests_vc_inspect.py
```diff
@@ -184,15 +184,18 @@ def test_parse_vm_props(self, mock_dt):
...
+            "guest.net": "",  # mac/ip addr returned by get_nics
+            "summary.runtime.powerState": "poweredOn",
+            "summary.guest.hostName": "hostname",
+            "summary.config.guestFullName": "Red Hat 7",
+            "summary.config.memorySizeMB": 1024,
+            "summary.config.numCpu": 4,
+            "summary.config.uuid": "1111",
+            "runtime.host": "",
+            "config.template": True,
+        }  # runs through host_facts values
```
bruno-fs, line 198: for reviewing purposes is better to just add new code w/o reformatting old stuff.
don't need to change anything, though.

quipucords/scanner/vcenter/tests_vc_inspect.py
```diff
@@ -227,22 +230,25 @@ def test_parse_vm_props(self, mock_dt):
...
+                "vm.cluster": "cluster1",
+                "vm.cpu_count": 4,
+                "vm.datacenter": "dc1",
+                "vm.dns_name": "hostname",
+                "vm.host.cpu_cores": 12,
+                "vm.host.cpu_count": 2,
+                "vm.host.cpu_threads": 24,
+                "vm.host.name": "host1",
+                "vm.ip_addresses": ["1.2.3.4"],
+                "vm.is_template": True,
```
bruno-fs, line 243: I think its odd this "vm" is a template and also have ip/mac address.
I'd set the mock with "vm.is_template" = False just to avoid this.

bruno-fs (approved):
Looks good! But I'd change one of the mocks

bruno-fs (approved):

PR #2174, opened 2022-07-18
[DISCOVERY-129] - Serializer dedicated to format data for Insights

bruno-fs:
- Add YupanaPayloadSerializer and a bunch of "child" serializers to properly format data for insights
- Refactor insights report view to use YupanaPayloadSerializer

Contributes to DISCOVERY-130
Implements DISCOVERY-129

### DIFF
--- a/quipucords/api/insights_report/serializers.py
+++ b/quipucords/api/insights_report/serializers.py
@@ -12,13 +12,43 @@
 from rest_framework import fields
 from rest_framework.serializers import Serializer, ValidationError
 
+from api.common.serializer import ForcedListSerializer, NotEmptyMixin
 from api.insights_report.constants import CANONICAL_FACTS
+from api.status import get_server_id
+from quipucords.environment import server_version
 
 default_kwargs = dict(required=False)
 
 # pylint: disable=abstract-method,fixme
 # Serializer has update/create as pseudo-abstract methods we don't need to implement
 # disable complaints about 'fixme' until DISCOVERY-130 is done
+
+
+class FactsSerializer(Serializer):
+    """Serializer for HBI facts."""
+
+    source_types = fields.ListField()
+    last_discovered = fields.DateTimeField()
+    qpc_server_version = fields.CharField(default=server_version)
+    qpc_server_id = fields.CharField(default=get_server_id)
+
+
+class FactsetSerializer(Serializer):
+    """Serializer for Host facts - it should be formatted as a list."""
+
+    namespace = fields.CharField(default="qpc")
+    facts = FactsSerializer(source="*")
+
+    def __init__(self, instance=None, **kwargs):
+        """Ininialize factset serializer."""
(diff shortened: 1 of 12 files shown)


quipucords/api/insights_report/serializers.py
```diff
@@ -0,0 +1,92 @@
...
+    mac_addresses = fields.ListField(child=fields.CharField(), **default_kwargs)
+    provider_id = fields.CharField(
+        source="name", **default_kwargs
+    )  # TODO: not sure on this one
+    provider_type = fields.CharField(source="cloud_provider", **default_kwargs)
+    satellite_id = fields.CharField(
+        **default_kwargs
+    )  # TODO: not sure on this one https://github.com/RedHatInsights/insights-host-inventory/blob/813a290f3a1c702312d8e02d1e59ba328c6f8143/swagger/api.spec.yaml#L901-L907
+    subscription_manager_id = fields.CharField(**default_kwargs)
+    system_profile = SystemProfileSerializer(source="*", **default_kwargs)
```
bruno-fs, line 51: `source=*` means that the same data available to the "parent" serializer (`YupanaHostSerializer`) will be passed over to the "child" `system_profile`.

The entity both will use is essentially the `SystemFingerprints` model with some extra steps

bruno-fs: SystemFingerprint model: https://github.com/quipucords/quipucords/blob/c603b2f9d73906c1ff32b95fa4ec1faa3d8cf707/quipucords/api/deployments_report/model.py#L68-L185

quipucords/api/insights_report/tests_insights_report.py
```diff
@@ -109,63 +110,70 @@ def setUp(self):
...
-            expected_hosts)
-        self.deployments_report.save()
-        with patch('api.insights_report.view.get_object_or_404',
-                   return_value=self.deployments_report):
-            response = self.client.get(url)
-        self.assertEqual(response.status_code, 200)
+        deployment_report = DeploymentReportFactory(
+            number_of_fingerprints=10_001,
+            status=DeploymentsReport.STATUS_COMPLETE,
+        )
```
bruno-fs, line 122: Running with 10k +1 fingerprints is definitely an overkill. on GH actions this is taking 11min lol

because of the serializer we need to properly create objects in the database - instead of simple dicts = `{"unique_id": <some number>}`

I didn't mock the setting controlling the number of hosts to keep the spirit of the original implementation, but this approach is too slow.

I think I'll change that. What do you think @gtanzillo @nicolearagao ?

gtanzillo: @bruno-fs I think you should definitely mock the number of hosts to reduce the time. 11 min is way too long for a test running on GH.

gtanzillo (approved):
LGTM, nice work @bruno-fs 👍

nicolearagao (approved):
LGTM, great work.

PR #2175, opened 2022-07-19
Feature/report entity 129

bruno-fs:
Add ReportEntity and ReportSlice, entities aggregating hosts.

This will replace the slicing logic implemented on [_create_report_slices](https://github.com/quipucords/quipucords/blob/93705627cdd84e27abb68bc6b291ecdb8b0d1dda/quipucords/api/insights_report/view.py#L86-L154) function.


quipucords/tests/factories.py
```diff
@@ -0,0 +1,63 @@
...
+    ip_addresses = factory.List([factory.Faker("ipv4")])
+    architecture = factory.Iterator(["x86_64", "ARM"])
+
+    class Meta:
+        """Factory options."""
+
+        model = "api.SystemFingerprint"
+
+
+class DeploymentReportFactory(DjangoModelFactory):
```
gtanzillo, line 31: For consistency, maybe this factory should be named `DeploymentsReportFactory` since it's a factory for `DeploymentsReport`

gtanzillo: In that case, I agree that it's safer to leave it as is.

gtanzillo (approved):
Looks great, thanks @bruno-fs 👍

nicolearagao (approved):
LGTM  🥇

PR #2178, opened 2022-07-25
[chore] Update Django version and handle deprecation warnings

bruno-fs:
- Handle MOST of deprecation warnings
- Upgrade django to 3.2 version (which will receive security updates until Abril 2024)
- [cachito] Add poetry-core to build dependencies

Part of DISCOVERY-93


gtanzillo (approved):
LGTM 👍

PR #2182, opened 2022-08-17
[Discovery 173] fix container image build

ruda:
* Fix a typo in `.dockerignore` that may produce the side-effect of leaving the container image with UI broken (blank screen after logged in). `quipucords/quipucords/template` -> `quipucords/quipucords/templates`
* Create a Makefile rule to build the container image by passing the `UI_RELEASE` - Note: should we keep the release number inside the `ARG UI_RELEASE`?

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -26,7 +26,7 @@ package-lock.json
 quipucords.egg-info
 quipucords/client
 quipucords/db.sqlite3
-quipucords/quipucords/template
+quipucords/quipucords/templates
 quipucords/staticfiles
 quipucords/tests
 var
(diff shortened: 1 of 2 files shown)


bruno-fs (approved):
LGTM :+1:

>should we keep the release number inside the ARG UI_RELEASE?


With the current implementation, we need it. That's how quay and github actions are able to use the right version. You could adjust the makefile to deal with an empty UI_RELEASE value, setting it to the default.

bruno-fs:
Whoops. UI integration test is broken. Seems to be expected, the way the main js file is generated is a little bit different now :thinking:

bruno-fs:
I suggest removing the commit bumping the UI version since it breaks two tests. One of the tests needs to be adapted anyways, but the other one will fix itself on the next UI release.

ruda:
> I suggest removing the commit bumping the UI version since it breaks two tests. One of the tests needs to be adapted anyways, but the other one will fix itself on the next UI release.

I will remove the commit for bumping the UI and push it again.

gtanzillo (approved):

PR #2183, opened 2022-08-23
[DISCOVERY-177] - Fix older RHEL detection bug

bruno-fs:
Since discovery 1.0.0, ansible is **always** using `paramiko` to connect to hosts (regardless of the option we chose when configuring sources).
A regression introduced in paramiko 2.9 is the culprit for RHEL-6 not being properly detected.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,7 +6,7 @@ on:
     tags:
       - "*"
   pull_request:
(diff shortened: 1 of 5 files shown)


gtanzillo (approved):
LGTM, thanks @bruno-fs 👍

nicolearagao (approved):
Thanks Bruno

PR #2186, opened 2022-08-29
Backports from release/1.0

bruno-fs:
Also:
- Update postgres image to 12 (since this is the one we are recommending for use)
- QoL/Dx updates to docker-compose

PR #2187, opened 2022-09-01
[DISCOVERY-130] - collect system memory on network scans

bruno-fs:
Add a new fact `system_memory_bytes` to both details and deployment reports for network scans.

Part of DISCOVERY-130 implementation


bruno-fs:
waiting on #2186 to get merged before adding this one

Makefile
```diff
@@ -82,6 +83,10 @@ lint-black:
 
 lint: lint-black lint-flake8
 
+lint-ansible:
```
nicolearagao, line 86: WIll those be applied to all the playbooks ?

quipucords/scanner/network/runner/roles/memory/tasks/main.yml
```diff
@@ -0,0 +1,23 @@
...
+  # https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/deployment_guide/s2-proc-meminfo
+
+- name: Get memory in kb
+  raw: |
+    set -o pipefail
+    cat /proc/meminfo |
+    grep MemTotal |
+    awk '{print $2}'
+  register: memory_in_kb
+  changed_when: "'memory_in_kb' in vars"
```
nicolearagao, line 15: Do you know if the scan will be stopped if there's an error during the execution of the command? If yes, then I would suggest adding a failed_when. 
We definetely want to have exceptions blowing up to know what's wrong but we want to make sure the scan finishes and collects the rest of the facts as well.

quipucords/scanner/network/runner/inspect.yml
```diff
@@ -28,4 +28,5 @@
     - insights
     - system_purpose
     - redhat_release
+    - memory
```
nicolearagao, line 31: Just curiosity, any specific reason why you added memory in this exact place?

gtanzillo (approved):
LGTM

quipucords/scanner/network/runner/inspect.yml
bruno-fs: I'd add as the last, but the event processor we attach to ansible runner expects `host_done` to be the last one.

quipucords/scanner/network/runner/roles/memory/tasks/main.yml
bruno-fs: Yes, the scanner breaks if an error is not treated. That's proably the reason why all tasks (that I saw) have `ignore_errors: true` 

I honestly don't expect an error here, AFAIK `/proc/meminfo` should be available and readable for any user but I'will add one just in case

bruno-fs: Interesting. I artificially made this command blow up and a `failed_when` condition to match it and don't help the scanner: it will stop anyway.

The only difference `failed_when` adds is the command fails gracefully.

How exactly do you propose failed when usage? (I added `failed_when: memory_in_kb.rc != 0`)

Makefile
bruno-fs: Yes. It loads the playbooks and in turn a 	Ansible-specific (case-agnostic): ll related roles will also be loaded and checked as well.

I wanted to add an actual linter here (ansible-linter probably), but we'll need to properly configure it first.

quipucords/scanner/network/runner/roles/memory/tasks/main.yml
nicolearagao: hmm, it makes sense to behave like that. I was thinking more about failed_when: insert undesirable side effect here, such as changing the state of something, not the faillure of the command itself.
But I guess I am over complicating things, and we definetely don't want to use the ignore_errors flag. This looks good the way it is.

nicolearagao (approved):
LGTM 👍🏼

quipucords/scanner/network/runner/roles/memory/tasks/main.yml
bruno-fs: honestly, your question and experimenting with failed_when gave me second thoughts on `ignore_errors`. A failure will blow up the scanner, smth we don't want to happen.

I honestly prefer the approach of breaking things as fast as possible, but the scanners were designed to be resilient to errors no matter what :thinking: 

Would you care to elaborate why we wouldn't want `ignore_errors`?

nicolearagao: The reason why I think we shouldnt use ignore_errors is aligned to what you stated above, we don't want to hide errors, the faster we fail the faster we are back on track. However, as you also mentioned, the scanner needs to be resilient. That's why I suggested the failed_when flag. We want it to ignore errors except if they are critical. Not being able to collect a fact would be considered critical? No. What could be considered critical in this specific case? I know it's unlikely, but what if this command interfered in the OS/state of the machine? If something like this happens, we def want to know asap. 
Alternatively we could keep using ignore_erros and find way to log the errors. example: {system_memory: error log}, so instead of wondering why my system never returns fact x, I could just check at the logs. I don't have an immediate suggestion on how we would implement this, but I could do some research if you think it''s worthwile.

bruno-fs: Thanks for clarifying! I think it's worth investigating this but not atm.

bruno-fs: I've added extra lines to the memory role to protect the scanner from errors.

PR #2189, opened 2022-09-19
Update UI version

bruno-fs:
UI version updated to latest (but still configurable through QUIPUCORDS_UI_VERSION env var)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ ENV PRODUCTION=True
 ENV PYTHONPATH=/app/quipucords
 ENV QUIPUCORDS_LOG_LEVEL=INFO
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -27,7 +27,7 @@ RUN pip install -r requirements.txt
 
 # Fetch UI code
 COPY Makefile .
-ARG UI_RELEASE=0.9.3
+ARG UI_RELEASE=latest
```
bruno-fs, line 30: @gtanzillo the script is defaulting to latest, but we could be specific here if we prefer so.

gtanzillo: Sounds good, we may need to lock to s specific version in the future, but now latest is good

gtanzillo (approved):

PR #2190, opened 2022-09-21
Force quipucords-ui to version 0.11.1

bruno-fs:
The layer fetching UI on quay.io is already cached with an older version of the UI. In order to update it we will need to force it to update by changing UI_RELEASE value.


gtanzillo (approved):
LGTM 👍

PR #2192, opened 2022-09-23
Feature/add openshift credentials

nicolearagao:
Add openshift credentials.
Refactor serializer validation to make sure it occurs before serializer method is_valid() is called. 
This pr is the server side implementation of task [DISCOVERY-185].


quipucords/api/credential/serializer.py
```diff
@@ -198,9 +200,25 @@ def validate_satellite_cred(self, attrs):
         )
         return attrs
 
+    def validate_openshift_cred(self, attrs):
+        """Validate the attributes for openshift credentials."""
+        # Required field for OpenShift credential
+        if not self.partial:
+            auth_token = attrs.get("auth_token")
+
+        if not auth_token:
```
bruno-fs, line 209: if partial is `False`, the variable `auth_token` will not be attributed.

quipucords/api/credential/tests_credential.py
```diff
@@ -31,7 +31,6 @@ def setUp(self):
 
     def tearDown(self):
```
bruno-fs, line 32: You can remove this function

quipucords/api/credential/tests_credential.py
```diff
@@ -146,7 +145,7 @@ def test_hc_create_err_p_or_ssh(self):
         or an ssh_keyfile.
         """
         expected_error = {
-            "non_field_errors": [
+            "non_field_errors": [  # pylint disable:implicit-str-concat
```
bruno-fs, line 148: instead of disabling this check you might fix it. concatenating both strings will do the trick.. The same is valid for the next one.

quipucords/api/credential/test_serializer.py
```diff
@@ -0,0 +1,98 @@
...
+    credential_object = serializer.save()
+    assert credential_object.name == "cred1"
+    updated_data = {
+        "name": "cred2",
+        "auth_token": "test_auth_token",
+    }
+    serializer.update(credential_object, updated_data)
+    assert serializer.is_valid()
+    assert credential_object.name == "cred2"
+    assert credential_object.cred_type == Credential.OPENSHIFT_CRED_TYPE
```
bruno-fs, line 98: can you please add 2 tests to the serializer?

- update w/o auth_token
- update initializing the serializer with `partial=True`

bruno-fs (changes requested):
Nice work! I've made small suggestions on comments

quipucords/api/credential/serializer.py
nicolearagao: Do we want to allow the user to update the auth_token as well ?

quipucords/api/credential/test_serializer.py
nicolearagao: hm, there's not much to update in openshift except for the name and auth_token. if you allow auth_token to not be passed when self.partial then how will the serializer know what has to be updated? The update method itself uses the instance to get the objects id therefore I beleive we should remove the self.partial

nicolearagao: How will serializer know we want to update an object and not create a new one if I only update the name per example? How can I know that the serializer I initialize with partial=True will point to the same object than before ?

quipucords/api/credential/serializer.py
nicolearagao: Good reminder, thanks!

bruno-fs: if you create a variable inside a if/else and will use it outside the nested scope, you need to create it in other conditions as well (or outside the if/else scope).

gtanzillo (approved):
Looks good @nicolearagao 👍  Good to go one you make the requested changes

quipucords/api/credential/tests_credential.py
nicolearagao: @bruno-fs and @gtanzillo I realized that instead of checking the whole aswer, I could check only if the key "non_field_errors" can be found on response.data. That's why I moved this change from the lint commit to the refactor one.

PR #2193, opened 2022-10-05
Feature/openshift source

nicolearagao:
Add openshift source.
Refactor serializer validation to make sure it occurs before serializer method is_valid() is called.
This pr is the server side implementation of task [DISCOVERY-186].


gtanzillo (approved):
Looks great! 👍

PR #2194, opened 2022-10-14
[DISCOVERY-198] Allow turning off django DEBUG mode

bruno-fs:
This is a security risk for servers accessible outside of the user network.

Fix DISCOVERY-198.

NOTE: UI templates also require a small change - otherwise the application will always crash on login page if DEBUG mode is disabled.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,7 +27,7 @@ RUN pip install -r requirements.txt
 
 # Fetch UI code
 COPY Makefile .
-ARG UI_RELEASE="0.11.1"
+ARG UI_RELEASE="1.0.0"
 RUN make fetch-ui -e QUIPUCORDS_UI_RELEASE=${UI_RELEASE}
 
 # Create /etc/ssl/qpc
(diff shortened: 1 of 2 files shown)


bruno-fs:
Integration test is expected to fail without a small change on UI side. Instructions for testing are available at quipucords/quipucords-ui#182

gtanzillo (approved):
LGTM 👍

nicolearagao (approved):
LGTM 👍🏼

ruda:
LGTM!

Optional - one-liner modification:

```python
DEBUG = True if os.environ.get("DJANGO_DEBUG", "true").lower() == "true" else False
```

and remove the the two line above your change.

ruda (approved):

bruno-fs:
thanks for reviewing, @ruda ! did you had the opportunity to test these changes alongside the ones proposed on quipucords/quipucords-ui#182 :question:

ruda:
Not yet, next thing I will do. Do you think it is early to merge?

PR #2201, opened 2022-10-27
Enabled overall status and and uptime facts that were disabled, behind feature flags

gtanzillo:
Enabling 2 facts - Overall Status and Uptime that were previously made unavailable by feature flags.

This PR is for simply enabling the two facts by setting their features to `True` to satisfy the requirement of making these facts available.

A longer-term goal would be to support simply removing the features from the list altogether to enable them. Currently, that is not possible without also removing the code that explicitly checks that these features are enabled. Instead, the logic could be changed such that checking availability of an unlisted feature would always return true. That will likely be done in a follow-up PR

Fix DISCOVERY-176


bruno-fs (approved):
LGTM :+1:

nicolearagao (approved):
LGTM 👍🏼

PR #2203, opened 2022-11-03
Fix logic for SSL certificate verification in OpenShift scans

bruno-fs:
- OCP Scanner should now respect SSL options (disable and verify)
- Added the ability to control timeouts for OCP scans via env vars `QPC_CONNECT_TASK_TIMEOUT` and `QPC_INSPECT_TASK_TIMEOUT` (in seconds)


nicolearagao (approved):
LGTM 👍🏼

quipucords/api/scantask/model.py
```diff
@@ -213,7 +212,7 @@ def log_message(
         """Log a message for this task."""
         # pylint: disable=no-member
         if exception:
-            logger.error(traceback.format_tb(exception.__traceback__))
+            logger.exception("Got an exception.")
```
gtanzillo, line 215: Won't we lose what the exception is, in addition to removing the traceback? Or does `logger.exception` handle that?

bruno-fs: logger.exception handles that =D
turns out that formatting in the old way was losing a lot of context from the traceback.

This is how it looks like now
image

gtanzillo (approved):
LGTM 👍

PR #2205, opened 2022-11-16
[DISCOVERY-180] - Inventory tags/labels

bruno-fs:
With these changes, we will be able to send tags to HBI.

They should appear like this

image

## filtering with tags 
in this example: data-collector tag = dsc

(collector tag can be customized through the env var `QPC_INSIGHTS_DATA_COLLECTOR_LABEL` (it's default value is 'qpc')
image


gtanzillo (approved):

nicolearagao (approved):
LGTM 👍🏼

PR #2206, opened 2022-11-18
[DISCOVERY-192] - swatch integration

bruno-fs:
- Add fields to insights report required for swatch integration.
- Prevent insights reports from breaking due to missing canonical facts in a system


quipucords/api/common/entities.py
```diff
@@ -42,6 +42,34 @@ def __getattr__(self, attr):
...
+            # either a combined satellite+vcenter or network+vcenter scan (only vcenter
+            # populates host_socket_count). The priority for choosing which count to
+            # send to hbi/swatch is satellite -> vcenter -> network (in other words:
+            # cpu_socket_count -> host_socket_count -> cpu_socket_count)
+            if Source.SATELLITE_SOURCE_TYPE in self._fingerprints.source_types():
+                return cpu_socket_count
+            # assuming vcenter
+            return host_socket_count
+        # if we only have one, just send whatever we have available
+        return cpu_socket_count or host_socket_count
```
bruno-fs, line 85: @justinkreft here's how I translated our discussion regarding choosing cpu_socket_count / host_socket_count

quipucords/api/common/entities.py
```diff
@@ -82,6 +106,16 @@ def mac_addresses(self):
...
+    @property
+    def provider_id(self):
+        """Retrieve provider_id.
+
+        Relying on fingerprint name is wrong for at least some cases. On AWS EC2 vms,
+        this should be the content of /var/lib/cloud/data/instance-id (which would be a
+        a new fact).
+        """
+        return self._fingerprints.name
+
```
bruno-fs, line 118: Here I'm repeating the behaviour implemented on 1.0.1 release, BUT now we know this is probably wrong.
Do you think it is better removing this fact @justinkreft @gtanzillo ?

the only advantage I see on keeping it is that maintaining this field will send provider type information to hbi (for instance "aws"), but IDK if this is useful for HBI.

gtanzillo: Based on our slack conversation, and the fact that `ProviderId` is described this way in the [API spec](https://github.com/RedHatInsights/insights-host-inventory/blob/b7a0b529713d31ebfa5b668a3c8df4b7b624be01/swagger/api.spec.yaml#L922-L929), I would lean toward removing it. It seems that the current behavior will insert a hostname in most cases, but that is pretty much useless for the purpose of `provider_id`, as it's defined in the API spec.

quipucords/api/compat/db.py
```diff
@@ -0,0 +1,65 @@
...
+                # ok then... it's not ¯\_(ツ)_/¯
+                return default
+        return None
+
+
+class StringAgg:
+    """Proxy for the appropriate StringAgg based on current db connection."""
+
+    def __new__(cls, expressions, delimiter=",", **extra):
+        """Initialize the proper StringAgg."""
```
nicolearagao, line 56: I didn't quite understand where the connection.vendor information comes from. I understand that __new__ will control the creation of a new instance that doesn't exist up to this point, unlike  the __init__ method. But where exactly do you tell the class where exactly to look for the connection info ?

quipucords/api/common/test_entities.py
```diff
@@ -126,3 +161,53 @@ def test_slicing(
...
         assert hosts_from_slices == report_entity.hosts
+
+
+@pytest.mark.django_db
+class TestHostEntity:
+    """Test HostEntity."""
+
+    @classmethod
+    def host_init(cls, system_fingerprint) -> HostEntity:
+        """Initialize HostEntity with using ReportEntity annotated query."""
```
nicolearagao, line 194: with is unneccessary in this sentence.

quipucords/api/common/test_entities.py
```diff
@@ -121,6 +122,26 @@ def test_last_discovered_initialized(self, django_assert_num_queries):
         with django_assert_num_queries(2):
             assert isinstance(report.last_discovered, datetime)
 
+    def test_canonical_facts_skipping(self, caplog):
+        """Test the mecanism for skipping hosts w/o canonical facts."""
```
nicolearagao, line 127: typo mechanism

quipucords/api/compat/db.py
bruno-fs: `connection.vendor` info comes from current db connection. 
see line 4 and lines 57 to 60 in this module.

nicolearagao: Thanks for taking the time to explain. I get it now.

gtanzillo (approved):
Looks good!!

nicolearagao (approved):
LGTM 👍🏼

PR #2208, opened 2022-11-22
Chore: update dependencies and CI

bruno-fs:


bruno-fs:
@ruda could you check with camayoc with nothing blows up with these updates? :pleading_face: :pray:

ruda:
@bruno-fs  I ran camayoc against this change here and quipucords 1.1, both gave the exact same results. So, looks good, no impact with the updates.

ruda (approved):
LGTM!

PR #2211, opened 2022-11-28
[DISCOVERY-190] - remove rust dependencies

bruno-fs:
same as #2208 , but downgrading cryptography.

cryptography rust dependencies are causing compliance issues (see DISCOVERY-190).


bruno-fs:
we decided to keep using lookaside cache

PR #2215, opened 2022-12-14
feat(ocp): WIP credential using username+password

bruno-fs:


quipucords/api/credential/serializer.py
```diff
@@ -219,7 +228,14 @@ def validate_openshift_cred(self, attrs):
     def _check_for_disallowed_fields(self, credential_type, attrs, message):
         """Check if forbidden fields are being passed to credentials."""
         required_fields_map = {
-            Credential.OPENSHIFT_CRED_TYPE: {"id", "name", "cred_type", "auth_token"},
```
nicolearagao, line 222: 👀

quipucords/api/credential/serializer.py
```diff
@@ -235,3 +243,64 @@ def prepare_data_for_update(cls, data: dict):
...
+        extra_kwargs = {
+            "password": {**ENCRYPTED_FIELD_KWARGS},
+            "auth_token": {**ENCRYPTED_FIELD_KWARGS},
+        }
+
+    def validate(self, attrs):
+        """Run validations that require more than one field."""
+        has_user_or_pass, has_auth_token = self._get_user_pass_or_auth_token(attrs)
+
+        if not has_auth_token ^ has_user_or_pass:
```
infinitewarp, line 273: Minor quibble. This doesn't feel legible to me at a glance, probably because when I see `^`, I think of bitwise operations, and I have to pause to consider if these variables are more than just bools. I think I would prefer a more verbose, though arguably redundant, form of this check.

bruno-fs: mmm I think operator has an explicit xor function. would that help?

bruno-fs:
```python
if not xor(has_auth_token, has_user_pass)
```

bruno-fs: or would u prefer it trully explicit like
```python
        if has_auth_token and has_user_or_pass or (not has_auth_token and not has_user_or_pass):

```

abellotti: The logic would be simplified if we had 2 ValidationErrors:

```
    if not has_auth_token and not has_user_or_pass:
        raise ValidationError("Missing authentication")

    if has_auth_token and has_user_or_pass:
        raise ValidationError("Either auth_token or username+password can be specified"):
```

first message could also be "Either auth_token or username_password must be specified"

quipucords/api/credential/serializer.py
```diff
@@ -235,3 +243,64 @@ def prepare_data_for_update(cls, data: dict):
...
+        validator.is_valid(raise_exception=True)
+        data = validator.validated_data
+        # add extra data to ensure non-sense combinations (like auth_token + username)
+        # don't occur
+        data.update(**extra_data)
+        return data
+
+    def _get_user_pass_or_auth_token(self, attrs):
+        """
+        Detect if if input data have username/password and/or auth token.
```
abellotti, line 294: minor:  if if -> if.   and have -> has

quipucords/api/credential/serializer.py
```diff
@@ -235,3 +241,67 @@ def prepare_data_for_update(cls, data: dict):
...
+    def _get_user_pass_or_auth_token(self, attrs):
+        """
+        Detect if has input data have username/password and/or auth token.
+
+        Returns a tuple of booleans: has_user_or_pass and has_auth_token in this order.
+        """
+        has_user_or_pass = bool(attrs.get("password") or attrs.get("username"))
+        has_auth_token = bool(attrs.get("auth_token"))
+        if self.partial and not has_auth_token and not has_user_or_pass:
+            # self.partial means PATCH, so it's OK to infer from instance
```
abellotti, line 302: minor, but might be better for readability:

```
    # self.partial means PATCH, so it's OK to infer from instance

    if self.partial:
        if not has_auth_token and not has_user_or_pass:
            ...
```

quipucords/api/credential/serializer.py
```diff
@@ -244,3 +244,80 @@ def prepare_data_for_update(cls, data: dict):
...
+        if has_auth_token:
+            none_fields = {"username": None, "password": None}
+            validator_class = AuthTokenSerializer
+        else:
+            none_fields = {"auth_token": None}
+            validator_class = UsernamePasswordSerializer
+        self._setup_default_data(attrs)
+        # force values to some fields to None to ensure non-sense combinations (like
+        # auth_token + username) can't happen
+        attrs.update(**none_fields)
```
infinitewarp, line 290: Could this ever actually happen? Rather than silently squash potentially bad data, should we treat bad data as an error and raise a validation exception?

bruno-fs: >Could this "nonsense combination" ever actually happen? 

It will happen if you switch from "auth token" to "username+password". 2 of the new tests are actually covering that.

>Rather than silently squash potentially bad data, should we treat bad data as a 400 error and raise a validation exception?

I usually prefer throwing errors with non-sense data, but for this case I feel the UX would be bad (we would need to require that username + password is explicitly set to None when switching to auth_token.

bruno-fs: Overriding this data with null is equivalent to this:
https://github.com/quipucords/quipucords/blob/486a5a1245d1387516acd79e3745ca3536afc7b0/quipucords/api/credential/serializer.py#L236-L246

infinitewarp: ack. Thanks for the explanation!

infinitewarp (approved):

PR #2216, opened 2022-12-16
chore!: remove container fingerprints

bruno-fs:
As per DISCOVERY-227, we realized it'll be better to map ocp/k8s nodes to fingerprints instead of namespaces.

Co-authored-by: Nicole Aragao <nicolearagao@users.noreply.github.com>


nicolearagao (approved):
LGTM 👍🏼

gtanzillo (approved):
LGTM 👍

PR #2221, opened 2022-12-26
tests: add DummyScanManager

bruno-fs:
Replace actual ScanManager in tests with a dummy scan manager instance.

Drop tests_manager as this test would require the actual ScanManager to make sense.

Related to DISCOVERY-50


nicolearagao (approved):
👏🏼 🤙🏼

abellotti:
Code LGTM!! 👍  now just trying to run the tests on my mac, could just be an env issue.

PR #2223, opened 2022-12-28
Feature/node collection implementation

nicolearagao:
- Create node and nodeResources entities for the
proper modeling and collection of node facts.
- Create validators for cpu and memory_bytes fields,
aiming the convertion of the received values to the desired
data type.
- Add api functions in order to retrieve desired facts from openshift api. 
- Modify scan task inspection functions to see nodes as systems instead of projects. 

ps: Lint is complaining about a variable found in the cluster commit. I wont correct it since it's probably fixed in bruno's [pr ](https://github.com/quipucords/quipucords/pull/2224)

This feature is part of the implementation of [DISCOVERY-227](https://issues.redhat.com/browse/DISCOVERY-227)

### DIFF
--- a/quipucords/scanner/openshift/entities.py
+++ b/quipucords/scanner/openshift/entities.py
@@ -11,12 +11,16 @@
 
 from __future__ import annotations
 
+import datetime
 import json
+import re
 from typing import Dict, List
 
-from pydantic import Field  # pylint: disable=no-name-in-module
+from pydantic import Field, validator  # pylint: disable=no-name-in-module
 
 from compat.pydantic import BaseModel, raises
+
+MEMORY_CONVERSION_ERROR = "This value couldn't be converted."
 
 
 def load_entity(data: dict) -> OCPBaseEntity:
@@ -82,6 +86,24 @@ class OCPCluster(OCPBaseEntity):
         return f"cluster:{self.uuid}"
 
 
+class OCPNode(OCPBaseEntity):
+    """Entity representing OpenShift Node."""
+
+    name: str
+    creation_timestamp: datetime.datetime = None
+    labels: Dict[str, str] = None
+    addresses: List[dict] = None
+    allocatable: NodeResources = None
+    capacity: NodeResources = None
+    architecture: str = None
+    kernel_version: str = None
+    machine_id: str = None
+    operating_system: str = None
+    taints: List[dict] = None
+    errors: Dict[str, OCPError] = Field(default_factory=dict)
+    _kind = "node"
+
+
 class OCPProject(OCPBaseEntity):
     """Entity representing OpenShift Projects/Namespaces."""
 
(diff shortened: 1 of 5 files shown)


quipucords/scanner/openshift/entities.py
```diff
@@ -128,3 +155,27 @@ def __raise__(self):
 
 # update nested model references - this should always be the last thing to run
 _update_model_refs()
+
+
+@dataclass
+class Cpu:
```
bruno-fs, line 163: just a nitpick but I like to break CamelCase for class name conventions when the class represents an all uppercase acronym (like CPU). IoW, Cpu -> CPU.

quipucords/scanner/openshift/test_inspect.py
```diff
@@ -8,14 +8,13 @@
 # https://www.gnu.org/licenses/gpl-3.0.txt.
 
 """Test OpenShift InspectTaskRunner."""
-
```
bruno-fs, line 11: ?

quipucords/scanner/openshift/entities.py
```diff
@@ -155,29 +158,70 @@ def __raise__(self):
...
+            fixed_point_suffix = re.compile("^\d+([KMGTPE])$")
+            group_match_2 = fixed_point_suffix.match(value)
+            if group_match_1:
+                value = convert_memory_binary_unit(value, group_match_1)
+            elif group_match_2:
+                value = convert_memory_fixed_point(value, group_match_2)
+            else:
+                raise ValueError(MEMORY_CONVERSION_ERROR)
+        else:
+            raise ValueError(MEMORY_VALUE_ERROR)
```
bruno-fs, line 195: I don't think we need to raise an error in this case. Consider we are deserializing an already parsed result with this entity.

quipucords/scanner/openshift/entities.py
```diff
@@ -155,29 +158,70 @@ def __raise__(self):
...
+    _kind = "node-resources"
+
+    # https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu
+    @validator("cpu", pre=True)
+    def _convert_cpu_value(cls, value):
+        millicore_suffix = re.compile("^\d+m$")
+        if isinstance(value, str) and millicore_suffix.match(value):
+            value = float(value.replace("m", "")) / 1000
+        else:
+            raise ValueError(CPU_VALUE_ERROR)
```
bruno-fs, line 176: we don't need this error. consider if cpu is not in microcores (or even is an integer or float).

bruno-fs (changes requested):

quipucords/scanner/openshift/entities.py
nicolearagao: Your jugment is correct in both cases. I will modify.

quipucords/scanner/openshift/entities.py
nicolearagao: Your jugment is correct in both cases. I will modify.

quipucords/scanner/openshift/test_inspect.py
```diff
@@ -68,7 +80,57 @@ def error():
...
+    list_nodes = []
+    list_nodes.append(node_ok)
+    return mocker.patch.object(OpenShiftApi, "retrieve_nodes", return_value=list_nodes)
+
+
+@pytest.fixture
+def _patched_retrieve_nodes_partial(mocker, node_ok, node_err):
+    """Mock OpenShiftApi retrieve_nodes method to return our fixtures."""
+    list_nodes = []
+    list_nodes.extend([node_ok, node_err])
```
bruno-fs, line 121: you can do this in only one line, without extend.

quipucords/scanner/openshift/test_inspect.py
```diff
@@ -68,7 +80,57 @@ def error():
...
+    return mocker.patch.object(OpenShiftApi, "retrieve_nodes", return_value=list_nodes)
+
+
+@pytest.fixture
+def _patched_retrieve_nodes_failure(mocker, node_err):
+    """Mock OpenShiftApi retrieve_nodes method to return our fixtures."""
+    list_nodes = []
+    list_nodes.append(node_err)
```
bruno-fs, line 129: you can do this in only one line without append

quipucords/scanner/openshift/entities.py
```diff
@@ -36,6 +36,7 @@ def _update_model_refs():
         model.update_forward_refs()
 
 
+# pylint: disable=too-few-public-methods
```
bruno-fs, line 39: this linting check is already disabled: https://github.com/quipucords/linting/blob/main/base-flakeheaven.toml#L33

quipucords/scanner/openshift/test_entities.py
```diff
@@ -145,3 +148,76 @@ class TestClass(OCPBaseEntity):
...
         assert entity.kind is not None
+
+
+@pytest.mark.parametrize(
+    "value, expected_result",
+    [
+        ("200m", 0.2),
+        ("3500m", 3.5),
+        (2.0, 2.0),
+        (4, 4.0),
```
bruno-fs, line 159: please add another parameter here:

something like `("2", 2.0),`

nicolearagao: @bruno-fs  Adding this specific pair (value =" 2," expected_value =2.0) will fail the first assertion because unless value it's a string and contains a millicore suffix, it won't be treated here, it will be returned to pydantic to perform its magic. 

This fails because the function  _convert_cpu_value won't do anything with "2"
```
    converted_value = NodeResources._convert_cpu_value("2")
    assert converted_value == 2.0
```
This will be successful: 
```
    node_resource = NodeResources(cpu="2")
    assert node_resource.cpu == 2.0
```
If u want me to add this use case I can split the test_cpu_validator_with_convertible_values in two different tests, in one I test if values can be converted, in the other I test final values of attributes.

bruno-fs: Good point.
I think we have 2 options here:
- ensure "2" will bypass the validation `("2", "2")`
- change the test to instantiate NodeResources with cpu attribute (it'd cover what we need).

I prefer the first option, but do whatever you prefer.

bruno-fs (approved):

bruno-fs (approved):

PR #2228, opened 2023-01-09
[DISCOVERY-240] - Attach debugger to running container

bruno-fs:


docker-compose.yml
```diff
@@ -14,9 +14,11 @@ services:
       QPC_SERVER_USERNAME: admin
       NETWORK_CONNECT_JOB_TIMEOUT: 6
       NETWORK_INSPECT_JOB_TIMEOUT: 600
+      QPC_DEBUGPY: 1
```
gtanzillo, line 17: Should this default to `0`, assuming `0` means off?

bruno-fs: yeah, zero means off!

I set to `1` considering docker-compose is meant for developers, not end-users :thinking: 

But if you prefer I can change anyway. its easy to override docker-compose locally without actually editing the file :thinking:

gtanzillo: No, that makes sense, let's keep it `1`. Thanks for the clarification

gtanzillo (approved):
lgtm 👍

CONTRIBUTING.md
```diff
@@ -37,9 +39,46 @@ You can run the lint command on your branch to check compliance.
 make lint
 ```
 
+### Debugging
+
+You can easily attatch a debugger to the running qpc-server on docker compose, which
```
abellotti, line 44: attach

PR #2235, opened 2023-01-16
Removed the manpage Makefile directive as it is not currently implemented.

abellotti:
- Minor update to the Makefile to remove the unused manpage directive.


bruno-fs (commented):
LGTM!

would you mind rewording the commit message to something like

```
chore: remove unused makefile manpage directive

removed the manpage directive since it is not implemented
```

I'm asking this for 2 reasons:
- we try to keep the "title" of commits having at most 50 characters
- we unofficially adopted [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) style for commit messages (if you look to our commit history you will notice this is recent)

bruno-fs (approved):

PR #2244, opened 2023-01-26
refat: simplify entity returning as forcedlist

nicolearagao:
Insights require some fields to be arrays, so instead of using a class in Meta to force this conversion, create method that will return the entity instance as a list.
This is a recurrent problem, this will be used again to send network_interfaces information to SystemProfile.

### DIFF
--- a/quipucords/api/common/serializer.py
+++ b/quipucords/api/common/serializer.py
@@ -16,7 +16,6 @@ from collections import OrderedDict
 from rest_framework.serializers import (
     ChoiceField,
     Field,
-    ListSerializer,
     ModelSerializer,
     ValidationError,
 )
@@ -105,13 +104,3 @@ class CustomJSONField(Field):
         if value == "":
             return value
         return json.loads(value)
-
-
-class ForcedListSerializer(ListSerializer):  # pylint: disable=abstract-method
-    """Serializer that forces data output as list."""
-
-    def to_representation(self, data):
-        """Force instance as list."""
-        if not isinstance(data, list):
-            data = [data]
-        return super().to_representation(data)
--- a/quipucords/api/insights_report/test_serializers.py
+++ b/quipucords/api/insights_report/test_serializers.py
@@ -31,7 +31,7 @@ def report_entity():
 def test_factset_serializer(db, report_entity):
     """Test FactsetSerializer."""
     serializer = FactsetSerializer(
-        report_entity.hosts[0],
+        report_entity.hosts[0].as_list(),
         many=True,
     )
     assert isinstance(serializer.data, list)
--- a/quipucords/api/common/entities.py
+++ b/quipucords/api/common/entities.py
@@ -56,6 +56,13 @@ class HostEntity:
             return super().__getattribute__(attr)
         return getattr(self._fingerprints, attr)
 
+    def as_list(self):
+        """Return a list containing this instance."""
+        # HBI expects some fields to be formatted as an array of objects, which
+        # sometimes is not applicable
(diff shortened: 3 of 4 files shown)


quipucords/api/insights_report/test_serializers.py
```diff
@@ -27,11 +27,12 @@ def report_entity():
     return ReportEntity.from_report_id(deployment_report.id)
 
 
+# pylint: disable=redefined-outer-name
```
bruno-fs, line 30: huh. this is odd. pylint-pytest plugin should be preventing this check in tests.

are you sure this is required?

nicolearagao: Lint wont pass neither on the pipeline nor locally if this is removed.

bruno-fs: what if you clean pylint cache (`rm ~/.cache/pylint`) and then run `flakeheaven lint` instead of `make lint`?

nicolearagao: It still complains about it. I thought it could be a problem in my local config but it also broke on the lint step in the gh pipeline.

quipucords/api/common/serializer.py
```diff
@@ -105,13 +105,3 @@ def to_representation(self, value):
         if value == "":
             return value
         return json.loads(value)
-
-
-class ForcedListSerializer(ListSerializer):  # pylint: disable=abstract-method
```
abellotti, line 110: With this, shouldn't `ListSerializer` be removed from line 19 ?

```
from rest_framework.serializers import (
    ChoiceField,
    Field,
    ListSerializer,
    ModelSerializer,
    ValidationError,
)
```

quipucords/api/insights_report/test_serializers.py
abellotti: fwiw, seems to be ok without it on the mac.

```
abellotti $ git diff
diff --git a/quipucords/api/insights_report/test_serializers.py b/quipucords/api/insights_report/test_serializers.py
index 6b5e268a..66043f66 100644
--- a/quipucords/api/insights_report/test_serializers.py
+++ b/quipucords/api/insights_report/test_serializers.py
@@ -27,7 +27,6 @@ def report_entity():
     return ReportEntity.from_report_id(deployment_report.id)
 
 
-# pylint: disable=redefined-outer-name
 @pytest.mark.dbcompat
 def test_factset_serializer(db, report_entity):
     """Test FactsetSerializer."""
```

then followed with:

```
abellotti $ flakeheaven lint quipucords/api/insights_report/test_serializers.py
abellotti $
```

Not sure if this is package version related or not, I'm running with python 3.9.13
and the following pylint versions:

```
abellotti $ pip list | grep lint
pylint              2.15.6
pylint-django       2.5.3
pylint-plugin-utils 0.7
pylint-pytest       1.1.2
```

quipucords/api/common/serializer.py
bruno-fs: yep, you are right.

we definitelly need to update the way our linter is executed.

(it's only checking git diff to speed up things, probably that's why it's not complaining about it)

quipucords/api/insights_report/test_serializers.py
nicolearagao: I was able to perform an exorcism in my machine and got it working.

abellotti (approved):

abellotti:
Thanks @nicolearagao for the updates, LGTM!! 👍   just need a rebase/repush.

bruno-fs (approved):
lgtm! but don't forget to rebase

PR #2245, opened 2023-01-27
chore: fix Makefile clean to work on MacOS

abellotti:
The -P option to grep for handline perl-flavored regex is not included in the default grep on MacOS. While a different can be installed with brew to support perl, there is nothing perl specific in that regexp in the Makefile that can't be handled with ERE, so using grep -E (aka egrep) does the trick and works on both Linux/CentOS/RHEL and MacOS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ help:
 all: lint test-coverage
 
 clean:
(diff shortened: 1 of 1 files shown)


bruno-fs (approved):
:+1:

gtanzillo (approved):
LGTM, thanks @abellotti

PR #2247, opened 2023-01-31
fix(scans): Scans raise exceptions when running on MacOS.

abellotti:
Starting scans on MacOS cause an exception pertaining to the Django app not properly initialized and unable to use gettext for defining messages.

The issue is that for the scanner jobs, we create a Process for each. With Processes, MacOS (and Windows) spawn a new process instead of a fork as in Linux.  With the former, the Django is not passed through, so in which case, we need to import django and do a django.setup() similar to how a Celery app config is initialized.

This needs to be done early before any of the other imports are done like importing api models where gettext is used that requires django to be setup.

### DIFF
--- a/quipucords/manage.py
+++ b/quipucords/manage.py
@@ -10,10 +10,14 @@
 # https://www.gnu.org/licenses/gpl-3.0.txt.
 #
 """Django server management module."""
+import multiprocessing
 import os
 import sys
 
 if __name__ == "__main__":
+    if sys.platform == "darwin":
+        multiprocessing.set_start_method("fork")
+
     os.environ.setdefault("DJANGO_SETTINGS_MODULE", "quipucords.settings")
     try:
         from django.core.management import execute_from_command_line


abellotti:
Regarding the lint failure above, `darker` seems to behave differently (and incorrectly) than `black`.  `black` properly honors the pylint directives in the file where `darker` did not and expected a change.

```
abellotti $ darker --diff quipucords/scanner/job.py

--- job.py	2023-01-31 12:10:51.592117 +0000
+++ job.py	2023-01-31 12:20:57.486496 +0000
@@ -10,14 +10,10 @@
 #
 """ScanJobRunner runs a group of scan tasks."""
 import logging
 from multiprocessing import Process, Value
 
-# pylint: disable-next=wrong-import-order
-# pylint: disable-next=unused-import
-from scanner import job_process_init  # noqa: F401
-
 from django.db.models import Q
 
 from api.common.common_report import create_report_version
 from api.details_report.util import (
     build_sources_from_tasks,
@@ -26,10 +22,13 @@
 )
 from api.models import ScanJob, ScanTask, Source
 from fingerprinter.task import FingerprintTaskRunner
 
 # pylint: disable-next=ungrouped-imports
+# pylint: disable-next=wrong-import-order
+# pylint: disable-next=unused-import
+from scanner import job_process_init  # noqa: F401
 from scanner import network, openshift, satellite, vcenter
 
 # Get an instance of a logger
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
```

and `black`:

```
abellotti $ black --diff quipucords/scanner/job.py
All done! ✨ 🍰 ✨
1 file would be left unchanged.
```

abellotti:
@bruno-fs regarding https://github.com/quipucords/quipucords/pull/2247#issuecomment-1410254324 above, `darker` not properly honoring all pylint directives.  Do you think this is a good time to switch to `black` for make lint (forgoing the version baseline and doing a bulk update/reformat of all files) ?

abellotti:
The earlier approach taken was not handling all cases, and some scan tasks were not initialized properly, with python 3.8 having shifted to spawn, the better approach is to set the start method to fork for the whole app via manage.py. This simplifies the fix and eliminates the blocking by darker.

quipucords/manage.py
```diff
@@ -10,10 +10,14 @@
...
 #
 """Django server management module."""
+import multiprocessing
 import os
 import sys
 
 if __name__ == "__main__":
+    if sys.platform == "darwin":
+        multiprocessing.set_start_method("fork")
+
```
bruno-fs, line 20: :clap:

bruno-fs (approved):
simple and yet an elegant solution! LGTM :+1:

could you squash the two commits into just one?

bruno-fs:
> The earlier approach taken was not handling all cases, and some scan tasks were not initialized properly, with python 3.8 having shifted to spawn, the better approach is to set the multiprocessing start method to fork for the whole app via manage.py. This simplifies the fix and eliminates the blocking by darker.

We can definitely drop darker now. By doing so we also need to add a check for isort (darker was doing both black and isort).

PR #2248, opened 2023-01-31
Feature/send network interfaces field

nicolearagao:
Send system profile's network_interfaces field.

This feature implements [[DISCOVERY-235]](https://issues.redhat.com/browse/DISCOVERY-253)
system_profile
ipv4

### DIFF
--- a/quipucords/api/common/entities.py
+++ b/quipucords/api/common/entities.py
@@ -9,6 +9,7 @@
 
 """Common entities."""
 
+import ipaddress
 import json
 import uuid
 from contextlib import suppress
@@ -131,10 +132,40 @@ class HostEntity:
             return json.loads(self._fingerprints.ip_addresses)
 
     @property
+    def ipv4_addresses(self):
+        """Retrieve only ipv4 addresses."""
+        ip_addresses = self.ip_addresses
+        ipv4_addresses = []
+        if ip_addresses:
+            for ip in ip_addresses:  # pylint: disable=invalid-name
+                try:
+                    if ipaddress.ip_address(ip).version == 4:
+                        ipv4_addresses.append(ip)
+                except ValueError:
+                    continue
+        return ipv4_addresses
+
+    @property
+    def ipv6_addresses(self):
+        """Retrieve only ipv6 addresses."""
+        # "Currently, we do not collect this information, but since it is
+        # mandatory for Yuptoo, we have agreed to send a default value."
+        # https://github.com/RedHatInsights/yuptoo/blob/265e9033ab66dce26d2a7bb5a909d7ec0b38a7da/yuptoo/modifiers/transform_network_interfaces.py#L8  # noqa: E501
+        return []
+
+    @property
     def mac_addresses(self):
         """Retrieve mac_addresses."""
         with suppress(TypeError):
             return json.loads(self._fingerprints.mac_addresses)
+
+    @property
+    def name_of_interface(self):
+        """Return interface name."""
+        # "Currently, we do not collect this information, but since it is
+        # mandatory for Yuptoo, we have agreed to send a default value."
+        # https://github.com/RedHatInsights/yuptoo/blob/265e9033ab66dce26d2a7bb5a909d7ec0b38a7da/yuptoo/modifiers/transform_network_interfaces.py#L8  # noqa: E501
+        return "unknown"
 
     @property
     def provider_id(self):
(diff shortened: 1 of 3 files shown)


quipucords/api/insights_report/tests_insights_report.py
```diff
@@ -91,6 +91,25 @@ def setUp(self):
...
+                    "cloud_provider": "aws",
+                    "network_interfaces": [
+                        {
+                            "ipv4_addresses": ["10.10.181.116"],
+                            "ipv6_addresses": [],
+                            "name": "unknown",
+                            "mac_address": "00:50:56:9e:7b:19",
+                        }
+                    ],
+                },
```
bruno-fs, line 112: I don't get it. Which test is consuming this data? Is it testing the legacy and **soon to be removed** cached insights column?

quipucords/api/common/entities.py
```diff
@@ -125,12 +127,52 @@ def ip_addresses(self):
...
             return json.loads(self._fingerprints.mac_addresses)
 
+    @property
+    def mac_address(self):
+        """Retrieve and validate mac_address."""
+        mac_addresses = self.mac_addresses
+        mac_regex = r"^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})$"
+        if mac_addresses:
+            if re.match(mac_regex, mac_addresses[0]):
+                return mac_addresses[0]
```
bruno-fs, line 165: what happens if the first mac address is invalid but the second is fine?
how about we iterate `mac_addresses` until we find one that works?

```suggestion
        for mac_addr in self.mac_addresses:
            if re.match(mac_regex, mac_addr):
                return mac_addr
```

bruno-fs: BTW, I don't love the fact we are randomly picking one from (possibly) many mac addresses, but I understand this is what insights requires.

We could also decide only send if there's only ONE valid mac_address among many, send if there's only ONE mac address and it is valid, not send et al, etc... Did you notice the pattern here? I'm talking about business rules.

Could you please check with @justinkreft whether this behavior is acceptable?

bruno-fs (commented):
Nice work! Ip addresses support is fine

On the `mac_address` topic (see the comments), I believe we need to understand the business rules first. If we don't reach a conclusion maybe separate it from this PR :thinking:

quipucords/api/common/entities.py
nicolearagao: I will bring those questions to Justin in the slack channel and update the pr either by removing the support for mac_address or adjusting the way it is validated.

quipucords/api/insights_report/tests_insights_report.py
nicolearagao: It is. No wonder system-profile was not there previously. I will remove it.

quipucords/api/common/entities.py
nicolearagao: According to the conversation with Justin, I will remove mac_address from this pr.

bruno-fs (commented):
Thanks for making the changes! I'll mark the conversations as resolved.

Now one potential problem came to mind: 
when ipaddresses is empty the report will have a network interfaces, right?
something like
```
{
  "ipv6_addresses": [],
  "name": "unknown"
}
  
```
Do we truly want that?
I think it's fine, but worth checking with justin.
I'm under the impression the report will have interface name and ipv6 addresses

quipucords/api/common/entities.py
nicolearagao: > BTW, I don't love the fact we are randomly picking one from (possibly) many mac addresses, but I understand this is what insights requires.
> 
> We could also decide only send if there's only ONE valid mac_address among many, send if there's only ONE mac address and it is valid, not send et al, etc... Did you notice the pattern here? I'm talking about business rules.
> 
> Could you please check with @justinkreft whether this behavior is acceptable?

we got an ok from justin and kevin regarding your concern. We are good to go.

nicolearagao:
> Thanks for making the changes! I'll mark the conversations as resolved.
> 
> Now one potential problem came to mind: when ipaddresses is empty the report will have a network interfaces, right? something like
> 
> ```
> {
>   "ipv6_addresses": [],
>   "name": "unknown"
> }
>   
> ```
> 
> Do we truly want that? I think it's fine, but worth checking with justin. I'm under the impression the report will have interface name and ipv6 addresses

we got an ok from justin and kevin howell regarding your concern. We are good to go.

bruno-fs (commented):
LGTM. Could you please rebase it?

bruno-fs (approved):

PR #2249, opened 2023-02-06
fix(netscan): Avoid data duplication in raw facts

bruno-fs:
The template for raw facts was incorrectly cached, overriding the null values with the values of each system.
This was causing issues on scans where some data should be missing.

Fix: DISCOVERY-262


gtanzillo (approved):
LGTM 👍

PR #2253, opened 2023-02-07
perf(insights-report): speed up generation of tags

bruno-fs:
This might be seem as a bugfix too, because quipucords wasn't able to generate insights reports with a big number of hosts.

## steps to reproduce
create a report with 1000 fingerprints
```
poetry run quipucords/manage.py shell
>>>from tests.factories import DeploymentReportFactory
>>>DeploymentReportFactory.create(number_of_fingerprints=1000, status="completed")
```
take note of the report id generated, exit shell and then open a browser, login to quipucords and access this url
```
https://<quipucords-host>:<quipucords-port>/api/v1/reports/<REPORT-ID>/insights/?format=tar.gz
```
Before applying this PR quipucords won't be able to render a report so big.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -54,6 +54,18 @@ pyyaml = "*"
 six = "*"
 
 [[package]]
+name = "appnope"
+version = "0.1.3"
+description = "Disable App Nap on macOS >= 10.9"
+category = "dev"
+optional = false
+python-versions = "*"
+files = [
+    {file = "appnope-0.1.3-py2.py3-none-any.whl", hash = "sha256:265a455292d0bd8a72453494fa24df5a11eb18373a60c7c0430889f22548605e"},
+    {file = "appnope-0.1.3.tar.gz", hash = "sha256:02bd91c4de869fbb1e1c50aafc4098827a7a54ab2f39d9dcba6c9547ed920e24"},
+]
+
+[[package]]
 name = "asgiref"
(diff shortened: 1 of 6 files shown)


abellotti (approved):

abellotti:
@bruno-fs 👍 lgtm!!  I take it the default ttl cache timeout is much shorter than when these systems change ip address or hostname ?

bruno-fs:
> @bruno-fs +1 lgtm!! I take it the default ttl cache timeout is much shorter than when these systems change ip address or hostname ?

Mmmm I didn't give any thought about it besides "the system would probably be restarted". We can pass a shorter value to force it to invalidate sooner. Do you have a suggestion? (going for a really low value like 60 sec would be better than the current situation) I'd guess 10min is probably conservative enough

bruno-fs:
> > @bruno-fs +1 lgtm!! I take it the default ttl cache timeout is much shorter than when these systems change ip address or hostname ?
> 
> Mmmm I didn't give any thought about it besides "the system would probably be restarted". We can pass a shorter value to force it to invalidate sooner. Do you have a suggestion? (going for a really low value like 60 sec would be better than the current situation) I'd guess 10min is probably conservative enough

Instead of caching I decided to grab the values on serializer initialization. The performance improvement is maintained with this approach, with the advantage of instantly updating ip/hostname value should those change.

PR #2259, opened 2023-02-13
Feature/send system purpose field

nicolearagao:
Send system profile's system_purpose field.

### DIFF
--- a/quipucords/api/common/entities.py
+++ b/quipucords/api/common/entities.py
@@ -21,6 +21,11 @@ from typing import Dict, List
 from django.conf import settings
 from django.db.models import F, Q, Value
 
+from api.common.enumerators import (
+    SystemPurposeRole,
+    SystemPurposeSla,
+    SystemPurposeUsage,
+)
 from api.models import DeploymentsReport, Product, Source, SystemFingerprint
 from compat.db import StringAgg
 
@@ -153,6 +158,26 @@ class HostEntity:
         if value and value.lower() in valid_providers:
             return value
         return None
+
+    @property
+    def system_purpose(self):
+        """Return system_purpose dictionary."""
+        sla = self._fingerprints.system_service_level_agreement
+        system_purpose_map = (
+            ("role", self._fingerprints.system_role, SystemPurposeRole),
+            ("usage", self._fingerprints.system_usage_type, SystemPurposeUsage),
+            ("sla", sla, SystemPurposeSla),
+        )
+        _system_purpose = {}
+        for name, attr, enum in system_purpose_map:
+            # since inventory expects specific values for each key in system_purpose
+            # dict, we need to check their values
+            # ref: https://github.com/RedHatInsights/insights-host-inventory/blob/b2c067f/swagger/system_profile.spec.yaml#L526-L547  # noqa: E501
+            for enum_item in enum:
+                if attr == enum_item.value:
+                    _system_purpose[name] = attr
+                    break
+        return _system_purpose
 
     @property
     def products(self) -> set:
--- a/quipucords/api/common/enumerators.py
+++ b/quipucords/api/common/enumerators.py
@@ -1,0 +1,36 @@
+# Copyright (C) 2023  Red Hat, Inc.
+
+# This software is licensed to you under the GNU General Public License,
+# version 3 (GPLv3). There is NO WARRANTY for this software, express or
+# implied, including the implied warranties of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv3
+# along with this software; if not, see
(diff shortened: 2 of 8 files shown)


quipucords/api/common/entities.py
```diff
@@ -352,3 +380,27 @@ def hosts(self) -> List[HostEntity]:
...
+    self_suport = "Self-Support"
+    standard = "Standard"
+
+
+class SystemPurposeUsage(Enum):
+    """Representing system purpose usage values accepted in HBI system profile."""
+
+    development_test = "Development/Test"
+    production = "Production"
+    disaster_recovery = "Disaster Recovery"
```
bruno-fs, line 406: mmm does inventory restrict the value for these fields?

nicolearagao: I believe so. Besides mentioning an enum list in the [inventory-schema](https://github.com/RedHatInsights/inventory-schemas/blob/cc4995e7b9e311919e55eb00578e2c4051c1727e/schemas/system_profile/v1.yaml#L526),  "Dev-Professional" appeared as a suitable value when calling `subscription-manager syspurpose service-level --list`. When I set my vm's system purpose sla to this value, insights report got to HBI but not to console. Observing kibana I could see an exception was raised during that specific field validation. This value was not on the enum list provided in the inventoy-schema.

bruno-fs: Cool! FYI we should the [spec at inventory repo](https://github.com/RedHatInsights/insights-host-inventory/blob/master/swagger/system_profile.spec.yaml) and [it's models.py module](https://github.com/RedHatInsights/insights-host-inventory/blob/master/app/models.py) as reference. 

In this case the values seem to be equal, so we are good.

bruno-fs: Only thing that bothers me is this code duplication across multiple repos. ugh.

nicolearagao: What do you mean by code duplication? I didnt quite get it what u mean.

quipucords/api/common/entities.py
```diff
@@ -154,6 +155,33 @@ def provider_type(self):
...
         return None
 
+    @property
+    def role(self):
+        """Retrieve system purpose role."""
+        role = self._fingerprints.system_role
+        list_roles = set(item.value for item in SystemPurposeRole)
+        if role in list_roles:
+            return role
+        return None
```
bruno-fs, line 165: FYI, since this is a set, not a list, I'd call the variable accordingly.
Also, since this is a set comprehension you could drop the `set` and just use curly braces

`        set_roles = {item.value for item in SystemPurposeRole}`

but I'd rather propose something else. only loop through SystemPurposeRole until we find the value.
best case scenario we hit the result right away

```
        for system_purpose_role in SystemPurposeRole:
            if role == system_purpose_role.value:
                return role
        return None
```

buuuuuuuuuut, there's something else bothering me here: these new properties are starting to pollute HostEntity. It's convenient they are called role/sla/usage but these names only make sense under `system_purpose` namespace.

we could either prefix all these variables with `system_purpose_` to make it clear OR maybe just build the whole object in a single method

```python
    def system_purpose(self):
        """Return system_purpose dictionary."""
        system_purpose_map = (
            ("role", self._fingerprints.system_role, SystemPurposeRole),
            ("usage", self._fingerprints.system_usage_type, SystemPurposeUsage),
            ("sla", self._fingerprints.system_service_level_agreement, SystemPurposeSla),
        )
        _system_purpose = {}
        for name, attr, enum in system_purpose_map:
            # since inventory expects specific values for each key in system_purpose dict, we need to check their values
            # ref: https://github.com/RedHatInsights/insights-host-inventory/blob/b2c067f/swagger/system_profile.spec.yaml#L526-L547
            for enum_item in enum:
                if attr == enum_item.value:
                    _system_purpose[name] = attr
                    # since we found an appropriate value, skip looping
                    break
        return _system_purpose
```

if you follow this last suggestion, remember to remove 'source=*' from `SystemPurposeSerializer` invocation

bruno-fs (commented):
Nice job!
Just few comments to improve it and we are good to go!

quipucords/api/common/entities.py
nicolearagao: Not really, I meant the code I sent above. I tried it here and both commands have the same result. The vm is a RHEL8. 😶 
https://access.redhat.com/articles/system-purpose

quipucords/api/common/entities.py
nicolearagao: I went for your suggestion. Thanks!

quipucords/api/common/entities.py
bruno-fs: cool. I think I did something wrong on mine, then haha.

bruno-fs (approved):
Awesome work!

PR #2263, opened 2023-02-15
chore(refact): Moving functions out of __init__.py

abellotti:
Keeping PyCharm and the Python demi-gods happy, moving functions out of __init__.py and into separate files.

### DIFF
--- a/quipucords/api/status/misc.py
+++ b/quipucords/api/status/misc.py
@@ -7,9 +7,14 @@
 # along with this software; if not, see
 # https://www.gnu.org/licenses/gpl-3.0.txt.
 
-"""Status module."""
+"""Status misc module."""
 
 from functools import lru_cache
 
-from .misc import get_server_id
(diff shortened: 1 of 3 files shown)


quipucords/api/status/misc.py
```diff
@@ -0,0 +1,20 @@
...
+
+from functools import lru_cache
+
+from .model import ServerInformation
+
+
+@lru_cache
+def get_server_id():
+    """Get server_id and cache it."""
+    return ServerInformation.create_or_retreive_server_id()
```
bruno-fs, line 20: OMG i never noticed this typo lol

bruno-fs (approved):
LGTM

quipucords/api/status/misc.py
abellotti: oh wow, quite a few uses of this function too.  Probably best as separate PR.

PR #2265, opened 2023-02-16
chore(refact): Fix method name

abellotti:
- Replacing create_or_retreive_server_id() with create_or_retrieve_server_id().
- Also, updated a couple of comments for the same.


bruno-fs (approved):
:+1:

PR #2267, opened 2023-02-21
chore(deps): Upgrading to Django 4.1.7

abellotti:
- Updating to Django 4.1.x as we may want to use async gets for the reports.
- Fixed the Signal calls to be compatible with Django 4.


bruno-fs (approved):
LGTM - I'm hoping the changes on signals and BooleanFields will get rid of some pesky warnings :eyes:

PR #2269, opened 2023-03-01
feat: add option to disable scan manager thread

infinitewarp:


quipucords/quipucords/settings.py
```diff
@@ -30,6 +30,10 @@
 
 PRODUCTION = bool(os.environ.get("PRODUCTION", False))
 
+DISABLE_THREADED_SCAN_MANAGER = bool(
+    os.environ.get("DISABLE_THREADED_SCAN_MANAGER", False)
```
bruno-fs, line 34: We have a non-written tradition of prefixing our env vars with `QPC_` (most of them are following this pattern). I'd change the env var to follow this pattern.

On the other hand, considering this variable will probably be removed in the near future, I won't mind if the prefix is missing.

quipucords/quipucords/wsgi.py
```diff
@@ -16,9 +16,12 @@
...
 from . import environment  # noqa: E402 pylint: disable=C0413
 
 environment.startup()
-SCAN_MANAGER.start()
+
+from scanner import manager  # noqa: E402 pylint: disable=C0413
+
+if not manager.SCAN_MANAGER:
+    manager.reinitialize()
+manager.SCAN_MANAGER.start()
```
bruno-fs, line 27: mmm, I'm under the impression the changes added to `quipucords/api/apps.py` are not necessary when factoring in this snippet :thinking:

quipucords/quipucords/settings.py
infinitewarp: Good point. I'll add `QPC_` to this!

quipucords/api/apps.py
```diff
@@ -9,4 +11,8 @@ class ApiConfig(AppConfig):
     name = "api"
 
     def ready(self):
-        """Mark server ready."""
+        """Initialize the scan manager when the API server is ready."""
+        if "runserver" in sys.argv:
+            from scanner.manager import reinitialize  # pylint: disable=C0415
+
+            reinitialize()
```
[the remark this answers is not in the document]
infinitewarp: I think you're right about this being unnecessary. This was the _first_ place I started to reinitialize the variable when I started this branch, oh, about two weeks ago, but since then I made the _other_ changes that may have made this redundant. I'll remove this and test a few things locally to confirm.

bruno-fs (approved):

PR #2270, opened 2023-03-02
docs: clean up inconsistent copyright notices

infinitewarp:
Unless otherwise stated, all source code files in this project are copyright Red Hat, Inc. and are licensed according to the terms of the LICENSE file.

### DIFF
--- a/quipucords/manage.py
+++ b/quipucords/manage.py
@@ -1,14 +1,4 @@
 #!/usr/bin/env python
-#
-# Copyright (c) 2017 Red Hat, Inc.
-#
-# This software is licensed to you under the GNU General Public License,
-# version 3 (GPLv3). There is NO WARRANTY for this software, express or
-# implied, including the implied warranties of MERCHANTABILITY or FITNESS
-# FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv3
-# along with this software; if not, see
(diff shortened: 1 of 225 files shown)


quipucords/manage.py
```diff
@@ -1,14 +1,3 @@
-#!/usr/bin/env python
```
bruno-fs, line 1: @infinitewarp Could this be the culprit causing issues? We might need this shebang line :thinking:

infinitewarp: Deleting the shebang here was a mistake. Thanks for catching it!

bruno-fs (approved):
lgtm

abellotti:
Curious, because this code would live on a customer system (vs saas), do the files need to include something saying "for copyright notice see the LICENSE file ..." ?

infinitewarp:
> Curious, because this code would live on a customer system (vs saas), do the files need to include something saying "for copyright notice see the LICENSE file ..." ?

The customer would have to try pretty hard to find any of the copyright/license-related comments since it's all bundled inside the container image(s).

That being said, the popular libraries that we use don't do include header notices everywhere. I would look to [Django](https://github.com/django/django) and [Django Rest Framework](https://github.com/encode/django-rest-framework) as leading examples. They have a LICENSE file at the top, and the rest of the code files are clean.

abellotti (approved):

PR #2276, opened 2023-03-16
Migrating CustomJSONField uses to JSONField

abellotti:
- [x] DetailsReport -> sources
- [x] Entitlement -> metadata
- [x] ExtendedProductSearchOptions -> search_directories
- [x] Product -> metadata
- [x] Product -> version
- [x] Source -> hosts
- [x] Source -> exclude_hosts
- [x] SystemFingerprint -> ip_addresses
- [x] SystemFingerprint -> mac_addresses
- [x] SystemFingerprint -> metadata
- [x] SystemFingerprint -> sources
- [x] SystemFingerprint -> user_login_history

**_Note_**: This leaves RawFact value as the only database field using CustomJSONField, to be handled in a separate PR due to the increased scope of this PR.

Also, This PR is best reviewed by looking at the commits separately.


quipucords/api/source/model.py
```diff
@@ -69,8 +68,8 @@ class Source(models.Model):
     port = models.IntegerField(null=True)
     options = models.OneToOneField(SourceOptions, null=True, on_delete=models.CASCADE)
     credentials = models.ManyToManyField(Credential)
-    hosts = models.TextField(unique=False, null=False)
-    exclude_hosts = models.TextField(unique=False, null=True)
+    hosts = models.JSONField(unique=False, null=True)
```
infinitewarp, line 71: Why did this switch from `null=False` to `null=True`?

quipucords/scanner/network/tests_network_connect.py
```diff
@@ -1,605 +1,605 @@
...
-    def test_empty_hosts(self, mock_run):
-        """Test running a connect scan with mocked connection."""
-        mock_run.return_value.status = "successful"
-        scanner = ConnectTaskRunner(self.scan_job, self.scan_task)
-        result_store = MockResultStore([])
-        _, result = scanner.run_with_result_store(
-            Value("i", ScanJob.JOB_RUN), result_store
-        )
-        self.assertEqual(result, ScanTask.COMPLETED)
+"""Test the discovery scanner capabilities."""
```
infinitewarp, line 1: Why did this whole file change? Line endings?

abellotti: the only source file in the repo with DOS line endings, https://github.com/quipucords/quipucords/pull/2276/commits/14dbcfdea687540242d28df7f7f536cff7a0259d.    was showing diffs with ^M at end.  ye-old dos2unix to the rescue.

quipucords/api/source/model.py
abellotti: Whenever any of the JSONField were null=False, was getting tons of errors like:
```
django.db.utils.IntegrityError: null value in column "metadata" of relation "api_systemfingerprint" violates not-null constraint
```

I think we're ok here since the serializers are helping us with the object existence and such via _required_=True when data is written.

quipucords/scanner/network/tests_network_connect.py
infinitewarp: Ugh. DOS line endings. 🤢 

Good riddance! Thank you! 👏

quipucords/api/source/model.py
infinitewarp: Hmm, interesting. 🤔 I feel like this would be worth revisiting this later, but for now, yeah we can probably trust the serializer to do its enforcement job. Hopefully there's no other way to write data without going through the serializer.

One last question before resolving this conversation: When did you encounter that error? Was it during the DB migration, running tests, and/or manually running the real service itself? My thought, which I don't think is right but I'll say it anyway, is that _maybe_ we need a two-step migration for these JSON fields. One to change the column/field type, and then one to set the not-null constraint. Did you try anything like that during development?

abellotti: Migrations to/from had no issues at all. The errors flood out while running the tests.

Not for the null constraint violations, but I've been playing with a field conversion (i.e. rename field to temporary name, create new/real new JSONField field name, convert/copy data over, then delete temporary name field) during the migration, I presume the same is needed for reverse migration.   I `think` I'll need this for the RawFact values unless JSONField performs some magic for us.

If you have some examples for the 2-step migrations you mentioned (change field, then set the not-null construct), I'll gladly give it a try.

infinitewarp: Oooh. I like having an empty list as a default.

quipucords/api/deployments_report/model.py
```diff
@@ -241,7 +241,7 @@ class Entitlement(models.Model):
     name = models.CharField(max_length=256, unique=False, null=True)
     entitlement_id = models.CharField(max_length=256, unique=False, null=True)
 
-    metadata = models.TextField(unique=False, null=False)
+    metadata = models.JSONField(unique=False, null=True)
```
bruno-fs, line 244: `default=dict` instead of `null=True` should work for the use case of this field

bruno-fs (commented):
Awesome work! I made some suggestions for the fields that didn't accept null value and now accept.

quipucords/api/source/model.py
bruno-fs: using an empty list might be a problem (as well using `{}`): you are saying the default value is a very specific list/dictionary already initialized in memory. 

Unless django does something to handle that issue, we would be mutating the same list/dict when default value is used.

bruno-fs: tldr: use default=list or default=dict

abellotti: 👍

quipucords/api/common/entities.py
```diff
@@ -148,7 +147,7 @@ def ipv6_addresses(self):
     def mac_addresses(self):
         """Retrieve mac_addresses."""
         with suppress(TypeError):
-            return json.loads(self._fingerprints.mac_addresses)
+            return self._fingerprints.mac_addresses
```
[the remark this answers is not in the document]
abellotti: These properties normalize the access from the _fingerprints relation (SystemFingerprint), so I'd still keep it, but we can remove the suppress TypeError, same with the ip_addresses property.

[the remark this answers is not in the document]
abellotti: mac_addresses and ip_addresses properties updated.

quipucords/api/migrations/0023_migrate_systemfingerprint_sources_to_jsonfield.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ("api", "0022_migrate_metadata_to_jsonfield"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="systemfingerprint",
+            name="sources",
+            field=models.JSONField(null=True),
```
bruno-fs: ~I'm curious about migrations like this one.
with the change to default shouldn't they change as well?~

EDIT I did run the branch locally and there was no need to update migrations. This is interesting. So this means that "default" sets db to accept null and the default object is handled on python side. Would be nice if there's an option to do that at db level (at times like these I miss sqlalchemy)

quipucords/api/common/entities.py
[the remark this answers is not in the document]
bruno-fs: FYI `HostEntity` has some dark magic in its `__gettattr__` implementation. Every time an attribute is not found in `HostEntity` we look up `_fingerprints`

(no need to change, tho. explicit is better anyways)

https://github.com/quipucords/quipucords/blob/a25cae869dd71e0287d17bf94e30203e823268a4/quipucords/api/common/entities.py#L46-L55

bruno-fs (approved):
:+1:

quipucords/api/source/model.py
infinitewarp: Looks good! Thanks for the thorough feedback! 👍

quipucords/api/migrations/0023_migrate_systemfingerprint_sources_to_jsonfield.py
abellotti: Yeah, I updated the migrations here in commit [f543d61a](https://github.com/quipucords/quipucords/pull/2276/commits/f543d61a0fe8adfb524ac6678623a73de8f116a5)  as makemigrations wanted to see those changes. Otherwise, they will be reflected in any future run.

quipucords/api/common/entities.py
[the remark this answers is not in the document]
abellotti: Uh oh, dark magic.  yeah, I guess as long as the uses don't get too verbose (if ... get from here otherwise get from there) in too many places, we can do a follow-up PR

PR #2278, opened 2023-03-17
fix: add missing __init__.py to fix broken pylint execution on some systems

infinitewarp:
Without this file, `pylint` attempts to dynamically add the `tests` directory to `sys.path` at runtime on some systems. Altering the path in that way results in `pylint` failing catastrophically because a `constants` module exists at both `./quipucords/constants.py` and `./quipucords/tests/constants.py`. `pytest` crashes when some `quipucords` modules attempt to import from `constants` because the wrong module is on the altered path.

Long discussion thread here: https://redhat-internal.slack.com/archives/C02QSNF1UKE/p1679074293160889


bruno-fs (approved):
:rofl:

PR #2280, opened 2023-03-21
DISCOVERY-287 - Refactor Credential serializer

bruno-fs:
- Refactor Credential Serializer and removing lots of unnecessary code.
- Add new serializer tests to ensure this refactoring didn't break anything

A follow up PR will be created soon, but this one is already big enough as is.

### DIFF
--- a/quipucords/api/credential/serializer.py
+++ b/quipucords/api/credential/serializer.py
@@ -1,6 +1,7 @@
 """Module for serializing all model object for database storage."""
 
 import os
+from collections import defaultdict
 
 from django.utils.translation import gettext as _
 from rest_framework.serializers import CharField, ValidationError, empty
@@ -10,7 +11,6 @@ from api.common.serializer import NotEmptySerializer
 from api.common.util import check_for_existing_name
 from api.models import Credential
 from constants import DataSources
-from utils import get_from_object_or_dict
 
 
 def expand_filepath(filepath):
@@ -27,7 +27,7 @@ ENCRYPTED_FIELD_KWARGS = {"style": {"input_type": "password"}}
 
 
 class CredentialSerializer(NotEmptySerializer):
-    """Serializer for the Credential model."""
+    """Base Serializer for the Credential model."""
 
     name = CharField(required=True, max_length=64)
 
@@ -42,6 +42,46 @@ class CredentialSerializer(NotEmptySerializer):
             "ssh_passphrase": ENCRYPTED_FIELD_KWARGS,
             "become_password": ENCRYPTED_FIELD_KWARGS,
         }
+
+    # Instead of following DRF BaseSerializer.__new__ signature (cls, *args, **kwargs),
+    # our __new__ method uses BaseSerializer.__init__ signature. This is OK, because:
+    # A) class level __new__ just calls __init__ with the same *args, *kwargs [1];
+    # B) DRF BaseSerializer.__new__ only modifies kwargs [3];
+    #
+    # References
+    # [1]: https://docs.python.org/3/reference/datamodel.html#object.__new__
+    # [2]: https://github.com/encode/django-rest-framework/blob/3.14.0/rest_framework/serializers.py#L121  # noqa: E501
+
+    def __new__(
+        cls, instance=None, data=empty, **kwargs
+    ):  # pylint: disable=arguments-differ
+        """Overloaded __new__ to return the appropriate serializer."""
+        cred_type = cls._get_cred_type(instance, data)
+        klass = cls._get_serializer_class(cred_type)
+        return super().__new__(klass, instance=instance, data=data, **kwargs)
+
+    @classmethod
+    def _get_serializer_class(cls, cred_type):
+        """Return the appropriate serializer based on 'cred_type'."""
+        serializer_per_datasource = defaultdict(
+            lambda: CredentialSerializer,
+            {
+                DataSources.NETWORK: NetworkCredentialSerializer,
+                DataSources.OPENSHIFT: AuthTokenSerializer,
+                DataSources.VCENTER: UsernamePasswordSerializer,
+                DataSources.SATELLITE: UsernamePasswordSerializer,
+            },
+        )
+
+        return serializer_per_datasource[cred_type]
+
+    @classmethod
+    def _get_cred_type(cls, instance, data):
+        if isinstance(instance, Credential):
+            return instance.cred_type
+        if isinstance(data, dict):
+            return data.get("cred_type")
+        return None
 
     def __init__(self, instance=None, data=empty, **kwargs):
         """Customize class initialization."""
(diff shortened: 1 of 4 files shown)


quipucords/api/credential/test_serializer.py
```diff
@@ -151,3 +218,298 @@ def test_change_with_same_cred_type():
...
+        assert serializer.errors == {"ssh_passphrase": [messages.HC_NO_KEY_W_PASS]}
+
+    @pytest.mark.django_db
+    @pytest.mark.parametrize(
+        "extra_serializer_kwargs",
+        (
+            pytest.param({}, id="full-update"),
+            pytest.param(
+                {"partial": True},
+                marks=pytest.mark.xfail(reason="BUG DETECTED!"),
```
bruno-fs, line 488: this (and the other xfailed tests) will be fixed in a follow up PR

quipucords/api/credential/tests_credential.py
```diff
@@ -59,18 +59,18 @@ def create(self, data):
     def create_expect_400(self, data):
         """We will do a lot of create tests that expect HTTP 400s."""
         response = self.create(data)
-        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        assert response.status_code == status.HTTP_400_BAD_REQUEST
```
infinitewarp, line 62: Since these are in a Django `TestCase`, I would assume `self.assertEqual` (and similar methods) provides at least some useful syntactic sugar. Are we deliberately refactoring these Django-specific assert* calls to plain Python `assert` statements in general?

bruno-fs: there's a huge difference on how the error is formatted!
I have an example in [this commit](6c657c949e7c09b07bd998a868d0eaa669b6b663)

essentially, plain asserts are not treated in a special way by pytest, with custom formatters that will help spot differences in long diffs  - and they are affected by pytest -v flag.

unittest asserts, on the other hand, don't show the whole diff (at least not when invoked by pytest) and don't show nothing interesting.

quipucords/api/credential/test_serializer.py
bruno-fs: just by the sheer size the PR. also because these bugs were not introduced here.

If you don't mind I can push more commits fixing it.

quipucords/api/credential/tests_credential.py
bruno-fs: TBH I'd replace all unit test asserts with plain asserts. Everytime I need to debug I start replacing.

(there's nothing broken now, but while developing I had to debug it)

quipucords/api/credential/serializer.py
```diff
@@ -23,86 +23,134 @@ def expand_filepath(filepath):
...
+        if not getattr(cls.Meta, "qpc_data_sources", None):
+            raise NotImplementedError(
+                "subclasses of CredentialSerializer MUST have a 'Meta' class "
+                "with 'qpc_data_sources' set as a list of DataSources."
+            )
+
+        # pylint doesn't know cls in __init_subclass__ refer to it's subclasses...
+        # pylint: disable=no-member
+        for data_source in cls.Meta.qpc_data_sources:
+            # pylint: enable=no-member
```
infinitewarp, line 121: Minor style question: why is this defined as a block disable instead of an inline disable?

```suggestion
        for data_source in cls.Meta.qpc_data_sources:  # pylint: disable=no-member
```

quipucords/api/credential/serializer.py
```diff
@@ -65,6 +64,22 @@ class Meta:
...
+            # this is only required to avoid breaking current functionality
+            # that was treating 'cred_type' as optional for updates and required for
+            # credential creation
+            data.setdefault("cred_type", instance.cred_type)
+        super().__init__(instance=instance, data=data, **kwargs)
+
+    def validate_cred_type(self, cred_type):
+        """Validate cred_type field."""
+        if self.instance and cred_type != self.instance.cred_type:
+            raise ValidationError((messages.CRED_TYPE_NOT_ALLOWED_UPDATE))
```
bruno-fs, line 132: ugh, I forgot to add the gettext wrapper here

quipucords/api/credential/test_serializer.py
infinitewarp: > also because these bugs were not introduced here.
Ah! I did not realize these were preexisting issues. In that case, sure, I leave it at your discretion!

quipucords/api/credential/tests_credential.py
infinitewarp: Interesting. I hadn't experienced this kind of problem before because every other Django project I worked on used the Django test runner, not a third-party one like pytest. Thank you for providing context. I think I'm okay with this!

quipucords/api/credential/serializer.py
```diff
@@ -23,225 +22,180 @@ def expand_filepath(filepath):
...
+    # B) DRF BaseSerializer.__new__ only modifies kwargs [3];
+    #
+    # References
+    # [1]: https://docs.python.org/3/reference/datamodel.html#object.__new__
+    # [2]: https://github.com/encode/django-rest-framework/blob/3.14.0/rest_framework/serializers.py#L121  # noqa: E501
+
+    def __new__(
+        cls, instance=None, data=empty, **kwargs
+    ):  # pylint: disable=arguments-differ
+        """Overloaded __new__ to return the adequate serializer."""
```
infinitewarp, line 55: Very subtle English wording suggestion: `s/adequate/appropriate/` because "adequate" implies that the result isn't necessarily the best output, that it's only barely good enough, whereas "appropriate" implies that is is the correct output.

```suggestion
        """Overloaded __new__ to return the appropriate serializer."""
```

quipucords/api/credential/serializer.py
```diff
@@ -23,225 +22,180 @@ def expand_filepath(filepath):
...
+        cred_type = cls._get_cred_type(instance, data)
+        klass = cls._get_serializer_class(cred_type)
+        return super().__new__(klass, instance=instance, data=data, **kwargs)
+
+    @classmethod
+    def _get_serializer_class(cls, cred_type):
+        """Return the adequate serializer based on 'cred_type'."""
```
infinitewarp, line 62: Same as the earlier comment.
```suggestion
        """Return the appropriate serializer based on 'cred_type'."""
```

quipucords/api/credential/serializer.py
```diff
@@ -23,225 +22,180 @@ def expand_filepath(filepath):
...
+            return data.get("cred_type")
+        return None
+
+    def __init__(self, instance=None, data=empty, **kwargs):
+        """Customize class initialization."""
+        if instance and isinstance(data, dict):
+            # assume cred_type == instance.cred_type if not provided
+            # this is only required to avoid breaking current functionality
+            # that was treating 'cred_type' as optional for updates and required for
+            # credential creation
```
infinitewarp, line 93: Is this referring to a behavior of external clients like the web UI or qpc CLI that would break? Or is this something internal to quipucords that we could refactor independent of the clients, and if so, we could remove this condition?

bruno-fs: referring to UI and maybe CLI

quipucords/api/credential/serializer.py
bruno-fs: ty!

bruno-fs: mmm github automatically resolves comments with code suggestions once they are applied.

quipucords/api/credential/serializer.py
```diff
@@ -23,225 +22,180 @@ def expand_filepath(filepath):
...
+
+    @classmethod
+    def _get_serializer_class(cls, cred_type):
+        """Return the appropriate serializer based on 'cred_type'."""
+
+        def _default():
+            return CredentialSerializer
+
+        serializer_per_datasource = defaultdict(
+            _default,
```
infinitewarp, line 68: Minor question: Was there a reason other than preferred coding style for writing this as a function instead of an inline anonymous lambda expression like this?

```suggestion
        serializer_per_datasource = defaultdict(
            lambda: CredentialSerializer,
```

bruno-fs: just PTSD (I thought I was clever writing lots of lambdas for processing a HUGE amount of data - the machine I had access had 1Tb ram XD - until I realized lambdas were causing that implementation to be slow as hell and had to refactor)

for web development this makes zero difference and is even more readable. I'll addopt your suggestion :wink:

nicolearagao (approved):
LGTM 👏🏼

PR #2284, opened 2023-03-22
use django-environ and clean up some settings

infinitewarp:
There are still some settings definitions that I'm not _super_ happy about, but I'm trying to make only minimally invasive changes here so that we can use [django-environ](https://github.com/joke2k/django-environ). Perhaps we should return and scrutinize all of the settings more closely again in the not-too-distant future.


quipucords/quipucords/settings.py
```diff
@@ -149,6 +150,9 @@ def create_random_key():
 
 # Database Management System could be 'sqlite' or 'postgresql'
 QPC_DBMS = env.str("QPC_DBMS", "postgres").lower()
+allowed_db_engines = ["sqlite", "postgres"]
+if QPC_DBMS not in allowed_db_engines:
+    raise ImproperlyConfigured(f"QPC_DBMS must be one of {allowed_db_engines}")
```
bruno-fs, line 153: YES YES YES

FAIL FAST :clap:

quipucords/quipucords/settings.py
```diff
@@ -169,9 +116,7 @@ def create_random_key():
 TEMPLATES = [
     {
         "BACKEND": "django.template.backends.django.DjangoTemplates",
-        "DIRS": [
-            os.path.join(os.path.dirname(__file__), "templates").replace("\\", "/"),
-        ],
+        "DIRS": [os.path.join(os.path.dirname(__file__), "templates")],
```
bruno-fs, line 119: ahem, have you heard the gospel of the std lib `pathlib`, young man?
lemme tell ya

```suggestion
        "DIRS": [Path(__file__).parent /  "templates"],
```

(`pathlib` is probably my favorite lib in standard library XD)

bruno-fs (approved):
:clap: a much saner settings module! thank you <3

quipucords/quipucords/settings.py
infinitewarp: LOL yeah, I _definitely_ noticed how inconsistently strings and Path objects were being treated in the settings file, some using `/` some using `join`, and a whole mess of different variables that probably could be simplified and consolidated, but I mentally compartmentalized all that into the

> Perhaps we should return and scrutinize all of the settings more closely again in the not-too-distant future.

comment I put at the top of the PR. 😅

infinitewarp: I manually added your suggestion to a commit on this branch instead of accepting it in GitHub because an extra space in `/  "` makes black unhappy.

abellotti:
LGTM 👍    would love to see, maybe v2's support of production vs development vs test settings.py 😉

PR #2285, opened 2023-03-23
chore: migrate RawFact.value to JSONField

abellotti:
Migrate RawFact.value to JSONField. and updated relevant code so we don't JSON loads/dumps as that's now handled by Django.


quipucords/api/inspectresult/model.py
```diff
@@ -83,7 +83,7 @@ class RawFact(models.Model):
     """A model of a raw fact."""
 
     name = models.CharField(max_length=1024)
-    value = models.TextField()
+    value = models.JSONField(null=True)
```
bruno-fs, line 86: I understand the OCP part is a WIP, but I'd try a less aggressive approach: customize the [JSONField encoder][1]

```python
from json import JSONEncoder
from pydantic import BaseModel
from datetime import datetime

class RawFactEncoder(JSONEncoder):
    def default(self, object):
        if isinstance(object, BaseModel):
            return object.dict()
        if isinstance(object, datetime):
            return object.isoformat()
        return super().default(object)
 ```

[1]: https://docs.djangoproject.com/en/4.1/ref/models/fields/#django.db.models.JSONField.encoder

Comment (bruno-fs) [reply quipucords/api/inspectresult/model.py:86]: IoW, write the puppy above and then

```python
value = models.JSONField(null=True, encoder=RawFactEncoder)
```

abellotti: 👍 Thanks, yes, this is much nicer and simpler doing that type of encoding for all RawFact values at the model, essentially adding support for OCP* types by default.

abellotti:
I've created an OpenShift source using my CRC Local instance and created a details report, then updated the source to this PR, migrated the data, reran the details report, and got the exact report.

Though that may have been cached, so what I've also done is with the new migrated JSONField RawFacts, I've regenerated system fingerprints for that report via: 

```
serializer = SystemFingerprintSerializer(
  instance=SystemFingerprint.objects.prefetch_related(
    "products", "entitlements"
  ).filter(
    deployment_report_id=2), many=True,
  )
system_fingerprints(json.dumps(serializer.data))
```

Excluding null or "" values, and comparing that will the original system_fingerprints from the report using the TextField RawFact DB and both match.

abellotti:
For my sanity, I also ran a second type of test as follows

- migrate to previous 0028 migration
- checkout master
- create openshift credentials and source
- run a scan
- query and save both details and deployments reports

- checkout this PR
- migrate to latest 0029 migration (RawFact.value --> JSONField)
- create a second credentials and source (to make sure nothing is used from previous)
- run a scan
- query and save both details and deployments reports

🥁 .....

Diffing both details and deployments reports, the `ONLY` differences were related to different source_name, report id, report_version, report_platform_id  and deployment_report id. 🎉 

This PR is good to go.

quipucords/api/migrations/0028_migrate_rawfact_value_to_jsonfield.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.2.17 on 2023-03-21 08:19
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("api", "0027_migrate_search_directories_to_jsonfield"),
```
bruno-fs: could you edit the migration dependency in the first commit?
(currently 058be18a18e0a26946539d3c6db61502b3f18a75)

bruno-fs (approved):
awesome work! :clap:

infinitewarp (approved):

quipucords/api/migrations/0028_migrate_rawfact_value_to_jsonfield.py
bruno-fs: squashing works too =D

PR #2287, opened 2023-03-24
test: rewrite unittest-style asserts as pytest

bruno-fs:
plain asserts are way more powerful for debugging with pytest.

"BUT WHY?", you might be asking... I decided to write about it [here](https://bruno-fs.github.io/posts/test-assertion-styles/)

NOTE: pylint directive `use-implicit-booleaness-not-comparison` is failing. I think we should disable those globally in tests.
Thoughts?

### DIFF
--- a/quipucords/api/common/tests_common_util.py
+++ b/quipucords/api/common/tests_common_util.py
@@ -29,22 +29,22 @@ class CommonUtilTest(TestCase):
         """Test csv helper with empty values."""
         # Test Empty case
         value = self.csv_helper.serialize_value("header", {})
-        self.assertEqual("", value)
+        assert "" == value
         value = self.csv_helper.serialize_value("header", [])
-        self.assertEqual("", value)
+        assert "" == value
 
     def test_csv_serialize_dict_1_key(self):
         """Test csv helper with 1 key dict."""
         # Test flat 1 entry
         test_python = {"key": "value"}
         value = self.csv_helper.serialize_value("header", test_python)
-        self.assertEqual(value, "{key:value}")
+        assert value == "{key:value}"
 
     def test_csv_serialize_list_1_value(self):
         """Test csv helper with 1 item list."""
         test_python = ["value"]
         value = self.csv_helper.serialize_value("header", test_python)
-        self.assertEqual(value, "[value]")
+        assert value == "[value]"
 
     def test_csv_serialize_dict_2_keys(self):
         """Test csv helper with 2 key dict."""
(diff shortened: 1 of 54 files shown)


bruno-fs:
I'm glad this PR started this conversation! 

>I suspect this will lead to a rise in style inconsistencies

We already have that (and I'm the one to blame :sweat_smile:). Besides the style inconsistency brought to the project by forcing `pytest` adoption, every time I'm debugging a test failure with a nested `dict`, `list` or a longer string (things that are not uncommon in this codebase) I replace `unittest` style test with `pytest` just because of it's the ability to pinpoint where exactly is the difference. We recently had commits ([1] and [2]) doing just that (heck, I'm about to do it again for some csv related tests that need a small refactor with ansible controller introduction).

So, understanding this is already an issue, what's the solution?  We could go the other way around: I stop adding pytest-specific stuff and convert `pytest` tests to `unittest`. That's definitely an option (I'd argue against it, not only for personal preference but as for all benefits pytest brought to the project), but I feel it's too extreme.

With all cards on the table, I had the intention to slowly convert the test style to `pytest`. Old tests would be treated as "legacy" and would be converted to `pytest` when they required maintenance - especially when refactoring is required. This does not mean that the plan should move as "I wanted" - just giving context.

> ...I noticed you didn't replace any of the assertRaises uses in this PR. If you're going to go big, then go all in?

I might do that haha XD

> Mixing them together feels even messier.

Yes, definitely. IF (emphasis on if) we come out from this discussion with a decision to support pytest-style, there's a [flake8 plugin](https://github.com/vikingco/flake8-pytest) that enforces this.

> Also, functions >>>> statements, IMHO, since they're proper objects that you can override and extend, even though it may be uncommon to override the stock assert* methods.

I understand and hear you. I took a veeery long time to accept the poison pill offered by `pytest` with its dark magic. Reading a single line of an assert statement in `unittest` style you know which object or what exactly you are testing. In this regard, I still prefer `unittest`. I only advocate for pytest due to its powerful assertion. 


[1]: https://github.com/quipucords/quipucords/commit/ab33b2d72203b748c0b957335135197d519d95c3
[2]: https://github.com/quipucords/quipucords/commit/6c657c949e7c09b07bd998a868d0eaa669b6b663

bruno-fs:
Thanks for the feedback, @nicolearagao, but considering the ongoing discussion, I'll mark this one as DRAFT.

If we do decide to make the conversion, I'll "undraft" it and rebase.

infinitewarp:
After sitting on this for a long time, I retract my aforementioned concerns, and I recommend we proceed with this change!

bruno-fs:
Given all the changes in the repo, might be best to restart this one from scratch XD

bruno-fs:
We agreed on slowly migrating tests to pytest on a "u touch, u convert it" basis.

PR #2288, opened 2023-03-24
clean up startup messages

infinitewarp:
- Use the loaded Django settings instead of bypassing them and reaching back into the env vars. 
- Add `QPC_LOG_ALL_ENV_VARS_AT_STARTUP` optional env var to control whether or not we dump every visible env var. Default value is `True` to preserve the existing behavior for users who expect to see these.

I want to refactor **_more_** of this startup code, like being much more deliberate about logging specific settings and putting it all somewhere more appropriate like `ApiConfig.ready`, but I'm practicing restraint here and only changing the most bothersome parts. 🙂 I will likely make more changes here soon. Baby steps.


quipucords/quipucords/environment.py
```diff
@@ -113,31 +122,38 @@ def startup():
...
+            "QPC_DBMS_USER",
+        ],
+        "sqlite": [
+            "QPC_DBMS",
+            "DB_PATH",
+        ]
+    }
+
+    for setting_name in settings_to_log[settings.QPC_DBMS]:
+        logger.info("%s set to %s", setting_name, repr(getattr(settings, setting_name)))
```
bruno-fs, line 145: <3

bruno-fs (approved):

quipucords/quipucords/environment.py
infinitewarp: In a future refactoring step, this is also the pattern I would follow to log more/all of our settings. Start with lists or a dict categorized by general functional area, and then just iterate through them all.

PR #2297, opened 2023-03-29
refactor: make Celery an optional dev dependency, add QPC_CELERY_ENABLE env var

infinitewarp:
I do not like this, but _technically_ putting all Celery-related code behind conditional checks _could_ work.


quipucords/quipucords/__init__.py
```diff
@@ -1,5 +1,8 @@
 """quipucords Django project package with various configuration modules."""
 
-from quipucords.celery import app as celery_app
+from django.conf import settings
 
-__all__ = ["celery_app"]
+if settings.QPC_CELERY_ENABLE:
```
infinitewarp, line 5: Note that we can't actually do it this way. When the tests run, observe:

```
django.core.exceptions.ImproperlyConfigured: Requested setting QPC_CELERY_ENABLE, but settings are not configured.
```

This is what I alluded to in Slack when I mentioned that you can't access the Django settings early enough. I _could_ yank directly from `os.environ`, but then we'd be proliferating _lots_ of raw, direct environment variable checks throughout our code. We'd have to do something like this everywhere we define new Celery task functions and everywhere we conditionally fork logic between using the old Thread-based manager and the new Celery tasks.

infinitewarp: I've confirmed locally that swapping `settings.QPC_CELERY_ENABLE` with `bool(os.getenv("QPC_CELERY_ENABLE"))` and `environ.Env().bool("QPC_CELERY_ENABLE", False)` work as expected, not exploding on `ImproperlyConfigured`, but those sure feel ugly and brittle.

abellotti: I thought we did something like:

```
from django.conf import settings

app = Celery("...")
app.config_from_object("django.conf:settings", namespace="CELERY")
...
```

in a config/celery.py in that other project that must not be named 😉 

Can't that help here ?

infinitewarp: I guess I could do some gnarly `ModuleNotFoundError` exception handling around the actual import statements, but 🤮

bruno-fs: eeek

yeah, looks like this is too much trouble to avoid working around the build problem downstream.

(on that note, kombu is not interested in releasing a patch release with the fix for setuptools 67+...)

bruno-fs: thanks for testing this approach, tho 💖

infinitewarp:
Closing this draft PR since it has served its intended purpose of facilitating conversations!

PR #2301, opened 2023-03-30
Refactor/yadsr

bruno-fs:
yet another refactor related to datasources and ansible controller implementation.

(just pulled these commits from the branch I'm working on adding ansible controller support)


infinitewarp (approved):
Nice. I like that `supports_partial_results` is now a native property of the various classes.

PR #2302, opened 2023-03-30
add (optional) django extensions for devs

infinitewarp:
After a normal `poetry install`, set and export `QPC_ENABLE_DJANGO_EXTENSIONS=True` into your environment to enable Django Extensions and gain access to new management commands like `shell_plus`.

See the Django Extensions docs for usage details: https://django-extensions.readthedocs.io/

To start a Jupyter Notebook server, for example, you might run:
```
QPC_ENABLE_DJANGO_EXTENSIONS=True poetry run ./quipucords/manage.py shell_plus --notebook
```

I included `notebook` as a dev dependency in this PR, but please note that it pulls in _several_ more package dependencies. 😬 If this is too much, please let me know, and I can drop that commit and just install it manually in my local environment.

Also, I added a `.gitignore` line for a top-level `notebooks` directory. As a matter of habit, when writing notebooks during development, I put them in a top-level directory like this, and they generally shouldn't be included with the project. Again, I'm happy to remove this commit if you disagree; I think my system should fall back to my personal `~/.gitignore` file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,6 +74,7 @@ target/
 
 # Jupyter Notebook
 .ipynb_checkpoints
+notebooks/
 
 # pyenv
 .python-version
--- a/poetry.lock
+++ b/poetry.lock
@@ -69,6 +69,27 @@ pyyaml = "*"
 six = "*"
 
 [[package]]
+name = "anyio"
+version = "3.6.2"
+description = "High level compatibility layer for multiple asynchronous event loop implementations"
+category = "dev"
+optional = false
+python-versions = ">=3.6.2"
+files = [
+    {file = "anyio-3.6.2-py3-none-any.whl", hash = "sha256:fbbe32bd270d2a2ef3ed1c5d45041250284e31fc0a4df4a5a6071842051a51e3"},
+    {file = "anyio-3.6.2.tar.gz", hash = "sha256:25ea0d673ae30af41a0c442f81cf3b38c7e79fdc7b60335a4c14e05eb0947421"},
+]
+
+[package.dependencies]
+idna = ">=2.8"
+sniffio = ">=1.1"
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -74,6 +74,7 @@ target/
 
 # Jupyter Notebook
 .ipynb_checkpoints
+notebooks/
```
bruno-fs, line 77: You might want to add notebooks to the ignore list for flakeheaven as well (unless you want the notebooks linted)

infinitewarp:
> I don't love adding so many jupyter dependencies because I tend to think we should have one jupyter server and several ipykernels (at least one per project/repo). Maybe shell_plus has some wirings that force us to add the whole thing, tho.

I haven't dug into the internal workings of `shell_plus --notebook`, but their documentation simply says to `pip install notebook`, pulling _all the things_ into your Django project's virtualenv. I haven't tried to see if it can integrate a local ipykernel with a notebook server _outside_ the project. A quick search through the django-extensions docs doesn't reveal anything obvious.

I don't have any strong feelings about keeping `notebook` packages here since it's easy to just manually install it into my local virtualenv. Plus there are several _other_ optional packages you can install if you wanted _more_ of the django-extensions features. I just included `notebook` because I sometimes use it, and I was feeling selfish/lazy. 😛 

Should I drop the notebook commit(s) and only keep `django-extensions`?

bruno-fs:
> Do you think I should drop the notebook commit(s) and only keep django-extensions, @bruno-fs ?
Nah, that's too extreme. I'm fine with it. If we start to have conflicts between dependencies we give it a thought.

bruno-fs (approved):
👍🏽 django shell plus is a great addition

ty

PR #2304, opened 2023-03-31
One more refactor

bruno-fs:
DRY up a test and prepare our "specialized requests.Session" class for ansible controller integration.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -10,8 +10,6 @@ from urllib.parse import urljoin
 
 import pytest
 
-from tests.utils.http import BaseUrlClient
-
 
 @pytest.fixture(autouse=True)
 def scan_manager(mocker):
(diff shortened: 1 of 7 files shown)


conftest.py
```diff
@@ -31,7 +29,9 @@ def scan_manager(mocker):
 @pytest.fixture(scope="session")
 def django_client(live_server):
     """HTTP client which connects to django live server."""
-    client = BaseUrlClient(
+    from compat.requests import Session
```
infinitewarp, line 32: Why local import?

infinitewarp (approved):
D 👏 R 👏 Y 👏

conftest.py
bruno-fs: the toplevel conftest.py has a policy of not importing quipucords code (unless it is under `tests` namespace). That's to avoid loading pesking global variables or messing with any django configuration that should happen later.

This also makes mocking easier.

bruno-fs: `api`, `quipucords` and `scanner` are the dangerous modules, tho.
Importing this particular one wouldn't break anything. at least not today.

infinitewarp: Fair enough! Thanks for the explanation. 🙂

PR #2308, opened 2023-04-03
Feature/add installed products raw fact

nicolearagao:
Add new fact installed_products to network scans. 

This is the first part of the implementation of [[DISCOVERY-236]](https://issues.redhat.com/browse/DISCOVERY-236).

### DIFF
--- a/quipucords/scanner/network/processing/installed_products.py
+++ b/quipucords/scanner/network/processing/installed_products.py
@@ -1,0 +1,42 @@
+"""Initial processing of the shell output from the installed_products role."""
+
+import logging
+
+from scanner.network.processing import process
+
+logger = logging.getLogger(__name__)
+
+
+class ProcessInstalledProducts(process.Processor):
+    """Process the installed_products fact."""
+
+    KEY = "installed_products"
+
+    @staticmethod
+    def process(output, dependencies=None):
+        """Process installed_product fact output."""
+        products = []
+        installed_products_cmd_output = output.get("stdout_lines", "")
+        # since the command is using grep with surrounding context (-C), we expect each
+        # result to be delimited by --
+        grep_separator = "--"
+        for product in installed_products_cmd_output.split(grep_separator):
+            product_dict = {}
+            for line in product.splitlines():
+                if not line.strip():
+                    continue
+                key, value = line.strip().split(":", 1)
+                if key in ["Name", "ID"]:
+                    product_dict[key.lower()] = value.strip()
+            if not product_dict.get("id"):
+                # considering the command includes grep "ID:", if we don't parse product
+                # with at least ID, there's an error on the implementation.
+                logger.error(
+                    "Unable to parse relevant product information from the following "
+                    " input\n%s",
+                    product,
+                )
+                continue
+            products.append(product_dict)
+
+        return products
--- a/quipucords/scanner/network/processing/test_installed_products.py
+++ b/quipucords/scanner/network/processing/test_installed_products.py
@@ -1,0 +1,83 @@
+"""Unit tests for installed_products fact initial processing."""
+
+import pytest
+
+from scanner.network.processing import installed_products
+
+
+@pytest.mark.parametrize(
+    "stdout_lines",
+    [
+        """Product:
+        \tID: 69
+        \tName: Red Hat Enterprise Linux Server
+        \tVersion: 7.0
+    """,
+        """Product:
+        \tName: Red Hat Enterprise Linux Server
+        \tID: 69
+        \tVersion: 7.0
+    """,
+    ],
+)
+def test_success_found_id_and_name(stdout_lines):
+    """ID and Name match expected format."""
+    cmd_output = {"rc": 0, "stdout_lines": stdout_lines}
+    expected_fact = [{"id": "69", "name": "Red Hat Enterprise Linux Server"}]
+    assert (
+        installed_products.ProcessInstalledProducts.process(cmd_output) == expected_fact
+    )
+
+
+@pytest.mark.parametrize(
+    "stdout_lines",
+    [
+        """Product:
+        \tName: Red Hat Enterprise Linux Server
+        \tVersion: 7.0
+    """,
+        """Certificate:
+        \tPath: /etc/pki/product/69.pem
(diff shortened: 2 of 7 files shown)


quipucords/scanner/network/processing/installed_products.py
```diff
@@ -0,0 +1,34 @@
...
+        """Process installed_product fact output."""
+        products = []
+        installed_products_cmd_output = output.get("stdout_lines", [])
+        for index, line in enumerate(installed_products_cmd_output):
+            if line.strip().startswith("ID:"):
+                _, id_value = line.strip().split(":", 1)
+                id_value = id_value.strip()
+                if index + 1 < len(installed_products_cmd_output):
+                    name_line = installed_products_cmd_output[index + 1]
+                    if name_line.strip().startswith("Name:"):
```
abellotti, line 26: Why is the Name: check in an index conditional (line 24), isnt't just another possible line Syntax, so this check is a peer to line 21 (check for ID:) ?

nicolearagao: If I understood your questions correctly, the name by itself is not relevant for HBI, therefore I will only collect it if an ID is associated with it.

abellotti: 👍

bruno-fs (changes requested):
please add a test case covering at least 2 products being detected.

quipucords/scanner/network/processing/test_installed_products.py
```diff
@@ -51,3 +47,22 @@ def test_failure_no_relevant_info():
...
     )
+
+
+def test_success_multiple_products():
+    """ID and Name match expected format for multiple products."""
+    stdout_output = (
+        "\r\nProduct:\r\n\tID: 479\r\n\tName: Red Hat Enterprise Linux for x86_64"
+        "\r\n\tVersion: 8.6\r\n--\r\n\r\nProduct:\r\n\tID: 69\r\n\tName: Red Hat"
+        " Enterprise Linux Server\r\n\tVersion: 8.7\r\n"
+    )
```
bruno-fs, line 58: not a blocker, but I'd rather read this as a multiline string. will be easier to read/maintain this.

quipucords/scanner/network/processing/installed_products.py
```diff
@@ -0,0 +1,31 @@
...
+            if line.strip().startswith("ID:"):
+                name = ""
+                _, id_value = line.strip().split(":", 1)
+                id_value = id_value.strip()
+                if index + 1 < len(installed_products_cmd_output):
+                    name_line = installed_products_cmd_output[index + 1]
+                    if name_line.strip().startswith("Name:"):
+                        _, name = name_line.strip().split(":", 1)
+                        name = name.strip()
+                products.append({"id": id_value, "name": name})
```
bruno-fs, line 30:
```suggestion
        installed_products_cmd_output = output.get("stdout", "")
        # since the command is using grep with surrounding context (-C), we expect each
        # result to be delimited by --
        grep_separator = "--"
        for product in installed_products_cmd_output.split(grep_separator):
            product_dict = {}
            for line in product.splitlines():
                if not line.strip():
                    # skip empty lines
                    continue
                key, value = line.strip().split(":", 1)
                if key in ["Name", "ID"]:
                    product_dict[key.lower()] = value.strip()
            if not product_dict.get("id"):
                # considering the command includes grep "ID:", if we don't parse product
                # with at least ID, there's an error on the implementation.
                logger.error(
                    "Unable to parse relevant product information from the following "
                    f" input\n{product}"
                )
                continue
            products.append(product_dict)
```

I propose a dramatic change here for readability but also to protect us from future errors.

the protection from errors part is the logging message. The rest of it is pretty much optional.

(I personally would just throw an exception and explode everything, but that would change network scan ethos )

quipucords/scanner/network/processing/test_installed_products.py
```diff
@@ -0,0 +1,68 @@
...
+        )
+        == expected
+    )
+
+
+def test_success_found_id():
+    """Only ID matches expected format."""
+    stdout_output = (
+        "Product:\r\n\tName: Red Hat Enterprise Linux Server\r\n\tID: 69\r\n\t"
+        "Version: 7.0\r\n\t"
```
bruno-fs, line 26: Since we are using grep with `-C`, I think we should take advantage of its delimiter to separate things.

quipucords/scanner/network/processing/test_installed_products.py
abellotti: Same here 😍 would be easier to see the tested data.

quipucords/scanner/network/processing/installed_products.py
bruno-fs: another option, this one respecting "Name" order (is it on stone that will always be after ID? if so, I think it deserves a comment on code, regardless of any of these snippets being adopted or not.)

```suggestion
        installed_products_cmd_output = output.get("stdout", "")
        # since the command is using grep with surrounding context (-C), we expect each
        # result to be delimited by --
        grep_separator = "--"
        for product in installed_products_cmd_output.split(grep_separator):
            product_dict = {"name": ""}
            for line in product.splitlines():
                stripped_line = line.strip()
                if stripped_line.startswith("ID:"):
                    product_dict["id"] = stripped_line.split(":", 1)[1].strip()
                if product_dict.get("id") and stripped_line.startswith("Name:"):
                    product_dict["name"] = stripped_line.split(":", 1)[1].strip()
                    break
            if not product_dict.get("id"):
                # considering the command includes grep "ID:", if we don't parse product
                # with at least ID, there's an error on the implementation.
                logger.error(
                    "Unable to parse relevant product information from the following "
                    f" input\n{product}"
                )
                continue
            products.append(product_dict)
```

quipucords/scanner/network/processing/test_installed_products.py
nicolearagao: Your wish is my command, gents!

bruno-fs (approved):
much nicer! LGTM

bruno-fs:
please don't forget to squash the commits before merging

PR #2309, opened 2023-04-03
refactor(ScanJobRunner): unbind "runner" logic from multiprocessing

bruno-fs:
With this small change we can finally run tests synchronously.

The "timeout" related code was also removed from DummyScanManager because that same behavior in tests can be achieved through pytest-timeout (with --timeout=15).

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -9,6 +9,7 @@ from functools import partial
 from urllib.parse import urljoin
 
 import pytest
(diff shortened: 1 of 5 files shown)


abellotti:
😍 make test still successful on MacOS without the multiprocessing module and thus the related needed mp startup method override. LGTM!

infinitewarp (approved):

PR #2312, opened 2023-04-04
fix: fix broken build-container on Mac M1

abellotti:
fix: fix broken build-container on Mac M1

make build-container was failing on M1 Macs. Dockerfile could not install psycopg2_binary as it tries to build it for aarch64. Was missing gcc, postgresql-devel and the python39-devel packages.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,7 @@ ENV PYTHONPATH=/app/quipucords
 ENV QUIPUCORDS_LOG_LEVEL=INFO
 
 COPY scripts/dnf /usr/local/bin/dnf
+ARG BUILD_PACKAGES="gcc postgresql-devel python39-devel"
 RUN dnf install \
         git \
         glibc-langpack-en \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -16,12 +16,15 @@ ENV QUIPUCORDS_LOG_LEVEL=INFO
...
 RUN dnf install \
+        gcc \
         git \
         glibc-langpack-en \
         jq \
         make \
         openssh-clients \
+        postgresql-devel \
         python39 \
+        python39-devel \
```
bruno-fs, line 27: totally optional: since these new dependencies are required only for building, they could be uninstalled after we pip install and (presumably) compile everything we need.

we do that downstream:  
- https://pkgs.devel.redhat.com/cgit/containers/discovery-server/tree/Dockerfile?h=discovery-1.1-rhel-8#n10
- https://pkgs.devel.redhat.com/cgit/containers/discovery-server/tree/Dockerfile?h=discovery-1.1-rhel-8#n49

bruno-fs (approved):
LGTM

infinitewarp (approved):
Yep! I've seen these as needed for some systems before as well. Good catch. 👍

PR #2313, opened 2023-04-04
chore: Adding the man command to the server image

abellotti:
chore: Adding the man command to the server image

Adding the `man-db` and `man` packages to the quipucords server docker image.

This provides the man command, but no man pages are installed by default.
(default microdnf behavior is --nodocs)

Before PR:

```
$ podman run quipucords man ls
Error: preparing container 5cac... for attach: crun: executable file `man` not found in $PATH: No such file or directory: OCI runtime attempted to invoke a command that was not found
```

After PR:

```
$ podman run quipucords man ls
No manual entry for ls
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,6 +21,8 @@ RUN dnf install \
         glibc-langpack-en \
         jq \
         make \
+        man-db \
+        man \
         openssh-clients \
         python39 \
         sshpass \


abellotti:
**Note**: 

The default behavior of microdnf in ubi-minimal is to NOT install docs (i.e. --nodocs is implied). While there could be a way  to override this behavior for the packages we install (git, python, etc) via --set-opt='tsflags=' or some other config file obscurity, we could simply manually copy the man pages for the commands we care about (qpc, etc.) manually.

bruno-fs:
I have zero understanding of man, but given we only want manpages for qpc I think what you are proposing works.

This can propably help on the next steps: https://github.com/quipucords/qpc/blob/master/qpc.spec#L46-L47

infinitewarp:
Why would we want man in our built server container? Do we expect customers will be jumping into the container and manually executing man for something? If not, it you want to run man to look up something for local development, why not install man locally yourself?

bruno-fs:
@infinitewarp we will recommend en users to run something like `toolbox run --container-image discovery-toolbox man dsc`.

Docs will probably recommend that users save this in an alias.

And we could do that JUST downstream, but it is more convenient to replicate what's been proven upstream there instead of waiting hours building on brew just to check if everything worked (we also have a story for completely dropping support for qpc rpm and bundling it in quipucords container for usage with toolbox).

abellotti:
Closing this in preference of having a separate qpc container that would include its own man page.

PR #2314, opened 2023-04-07
refactor: dedupe several execute_request calls into a generator

infinitewarp:
This change is another cleanup on the long journey to Celery. I have introduced a generator to consolidate several calls to Satellite v6 sources that would require multiple requests through paginated results. This is an improvement, but I'm not done cleaning up this module. Expect to see other PRs touching the same code in the near future.

I recommend reviewing the changes with "hide whitespace" enabled because I de-indent some code blocks.

### DIFF
--- a/quipucords/scanner/satellite/six.py
+++ b/quipucords/scanner/satellite/six.py
@@ -1,12 +1,15 @@
 """Satellite 6 API handlers."""
+from __future__ import annotations
 
+import itertools
 import logging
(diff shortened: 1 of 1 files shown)


bruno-fs (approved):
DRY all the things!

abellotti (approved):

nicolearagao (approved):
LGTM 👍🏼

PR #2315, opened 2023-04-10
DISCOVERY-43 - ansible automation controller integration

bruno-fs:
reviewing commit by commit is highly advisable


quipucords/api/inspectresult/model.py
```diff
@@ -92,6 +92,8 @@ def default(self, o):
             return o.dict()
         if isinstance(o, datetime):
             return o.isoformat()
+        if isinstance(o, set):
+            return sorted(o)
```
infinitewarp, line 96: Just curious: Why are we sorting sets? Just for readability by human eyes and diff-ability?

bruno-fs: None.
Json encoder doesn't know how to serialize sets! 😰 😅

quipucords/api/common/entities.py
```diff
@@ -309,10 +309,11 @@ def _get_deployment_report(cls, report_id):
             .filter(pk=report_id)
             .get()
         )
-        # openshift sources won't make sense in insights reports.
+        # openshift/ansible sources won't make sense in insights reports.
         fingerprints = list(
             SystemFingerprint.objects.filter(deployment_report=deployment_report)
             .exclude(sources__icontains=DataSources.OPENSHIFT)
+            .exclude(sources__icontains=DataSources.ANSIBLE)
```
infinitewarp, line 316: Just a thought for the future, _not_ something to change now: Do you foresee this being a pattern as we expand source types in the not-too-distant future? At some point, would it be better to interrogate some kind of class metadata like some kind of "relevant for deployment reports" flag? Or is that _too_ forward thinking and possibly overkill?

quipucords/api/source/serializer.py
```diff
@@ -77,6 +77,13 @@ class SourceSerializer(NotEmptySerializer):
     options = SourceOptionsSerializer(required=False, many=False)
     credentials = CredentialsField(many=True, queryset=Credential.objects.all())
 
+    HTTP_SOURCE_TYPES = (
+        DataSources.ANSIBLE,
+        DataSources.OPENSHIFT,
+        DataSources.SATELLITE,
+        DataSources.VCENTER,
+    )
```
abellotti, line 85: how come DataSources.NETWORK is not here ?

quipucords/api/inspectresult/model.py
infinitewarp: IT DOESN'T? 😱 

Jeez. 🤦

quipucords/compat/requests.py
```diff
@@ -27,16 +28,32 @@ def __init__(self, *, base_url=None, auth=None, verify=True):
         self.verify = verify
         self.base_url = base_url
         self.auth = auth
-
-    def request(self, method, url, *args, **kwargs) -> requests.Response:
+        adapter = requests.adapters.HTTPAdapter(
+            max_retries=Retry(
+                3,
+                status_forcelist=[429, 500, 502, 503],
+                backoff_factor=0.1,
```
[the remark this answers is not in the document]
bruno-fs: no logic, it was just an educated guess XD

I don't see us exposing this to end users (unless we do it through env vars)

quipucords/api/source/serializer.py
bruno-fs: DataSources.NETWORK relies on the logic for "ssh_source", which is not part of the deduplication. Those other sources are kinda the same thing when we think about how their source model should behave

bruno-fs: we could have a `SSH_SOURCE_TYPES` constant, which would only contain `DataSources.NETWORK` 🤔

quipucords/compat/requests.py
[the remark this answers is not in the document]
bruno-fs: you know, what, I take it back. it's better to expose this as env vars. high latency could be a problem.

quipucords/tests/integration/test_smoker.py
```diff
@@ -0,0 +1,230 @@
...
+            metadata[fingerprint_fact] = {
+                "server_id": mock.ANY,
+                "source_name": self.SOURCE_NAME,
+                "source_type": self.SOURCE_TYPE,
+                "raw_fact_key": RawFactComparator(raw_fact),
+                "has_sudo": False,
+            }
+        return metadata
+
+    def test_deployments_report(  # pylint: disable=too-many-arguments
```
infinitewarp, line 158: I also ran into `too-many-arguments` recently, and I wonder if we disagree enough with pylint's "5 or fewer" opinion to add it to our global ignore list. At last check, we disable that check in 23 places (24 with this addition).

Since we do thorough code reviews, _hopefully_ we would notice and stop each other from needlessly tacking on lots of function arguments. Thoughts?

We don't need to act on this now. Seeing your comment just reminded me. 🙂

quipucords/compat/requests.py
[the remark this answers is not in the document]
infinitewarp: I see a magic number, and my first thought is "should this be env var?" 😀

quipucords/tests/integration/test_smoker.py
bruno-fs: I think it could have an exception for testing only  - kinda hard to avoid the fixtures XD - and/or configure pylint to support a higher value on this check. Outside of tests I personally think this number is a good advice, but that's just me.

infinitewarp: I don't _want_ to discourage GitHub issues because honestly they're so much better than Jira, but that issue belongs in Jira. 😕 Issues are going to slip through and be overlooked and forgotten if we have to remember to look in multiple places for them, and like it or not, Jira is the designated tracker for our project. I completely stopped looking at GitHub issues several weeks ago when I realized we were using Jira for everything (most things?).

infinitewarp (approved):
I would love to see big changes like this broken into more smaller PRs in the future. Changes to pagination and OpenShift tests, for example, deserve their own reviews. I found myself getting lost trying to keep this long diff in my head, and I had to go to the individual commits view to finish reviewing parts of it.

I feel like I'm not being as thorough with my review since I'm looking at so many changes at once and rushing to call it good enough on code freeze day; I know I've caught some things on a second reading that I missed on the first. Otherwise, 👍

quipucords/scanner/ansible/api.py
```diff
@@ -0,0 +1,68 @@
...
+        kwargs.setdefault("params", {})
+        kwargs["params"].update(**params)
+        return kwargs
+
+    def _get_pages_in_parallel(self, url, max_concurrency, page_kwargs):
+        with futures.ThreadPoolExecutor(max_workers=max_concurrency) as executor:
+            future_to_page = {
+                executor.submit(self.get, url, **kwargs) for kwargs in page_kwargs
+            }
+            for future_page in futures.as_completed(future_to_page):
```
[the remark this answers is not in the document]
bruno-fs: yep. order doesn't matter. we just need to collect the data

quipucords/api/messages.py
```diff
@@ -65,8 +65,7 @@
 NET_INVALID_PORT = "Source of type network must have ssh port in range [0, 65535]"
 NET_HC_DO_NOT_EXIST = "Host credential with id=%d could not be found in database."
 NET_SSL_OPTIONS_NOT_ALLOWED = "Invalid SSL options for network source: %s"
-VC_INVALID_OPTIONS = "Invalid options for vcenter source: %s"
-SAT_INVALID_OPTIONS = "Invalid options for satellite source: %s"
+INVALID_OPTIONS = "Invalid option for '%(source_type)s': %(options)s."
```
abellotti, line 68: should that be Invalid `options` since that is what is being dry'ed up ?

bruno-fs: whoopsie. Dunno why I removed the "s" here. Should be options.

abellotti (approved):

quipucords/scanner/ansible/inspect.py
```diff
@@ -68,10 +68,20 @@ def execute_task(self, manager_interrupt):
             return self.success_message, ScanTask.COMPLETED
         return self.failure_message, ScanTask.FAILED
 
+    @property
+    def max_concurrency(self):
+        """Return scan job max concurrency option."""
+        try:
+            return self.scan_task.job.options.max_concurrency
+        except AttributeError:
+            return ScanOptions.get_default_forks()
```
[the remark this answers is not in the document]
bruno-fs: created a jira ticket for this one https://url.corp.redhat.com/79ba4e4

PR #2316, opened 2023-04-10
refactor: move setup code outside of "with multiprocessing.Pool" blocks

infinitewarp:
I want to dedupe more old messy iteration logic. Originally, I copied recipes from the Python docs and updated them for our needs (one of them didn't work "out of the box" for us), but then I found `more-itertools` which comes recommended by the official Python docs here: https://docs.python.org/3/library/itertools.html#itertools-recipes

> Substantially all of these recipes and many, many others can be installed from the [more-itertools project](https://pypi.org/project/more-itertools/) found on the Python Package Index:
> 
> `python -m pip install more-itertools`
> 
> Many of the recipes offer the same high performance as the underlying toolset.

Great!

Otherwise, the changes in this PR are just to continue moving code away from the pool blocks.

Also note that the diff shrinks a bit if you enable "Hide whitespace" because some is just code dedenting.


abellotti:
❤️ unique_everseen and chunked, reducing lots of busy code.

quipucords/scanner/satellite/six.py
```diff
@@ -551,25 +550,25 @@ def hosts_facts(self, manager_interrupt):
                 if host not in deduplicated_hosts:
                     hosts_after_dedup.append(host)
                     deduplicated_hosts.append(host)
-            hosts = hosts_after_dedup
-            chunks = [
-                hosts[i : i + self.max_concurrency]
-                for i in range(0, len(hosts), self.max_concurrency)
-            ]
-            chunks_grouped_by_org.append(chunks)
```
bruno-fs: I don't feel 100% comfortable on removing the group by org part.
Did the satellite instance used for tests had more than one org?

quipucords/scanner/satellite/six.py
```diff
@@ -551,25 +550,25 @@ def hosts_facts(self, manager_interrupt):
...
+                    raise SatelliteCancelException()
 
-                    if manager_interrupt.value == ScanJob.JOB_TERMINATE_PAUSE:
-                        raise SatellitePauseException()
+                if manager_interrupt.value == ScanJob.JOB_TERMINATE_PAUSE:
+                    raise SatellitePauseException()
 
-                    host_params = self.prepare_host(chunk)
-                    results = pool.starmap(request_host_details, host_params)
-                    process_results(self, results, 1)
```
bruno-fs: see comment above about group by org

quipucords/scanner/satellite/six.py
infinitewarp: Unfortunately, the satellite instance I have been using (`satellite1.latest-el7.satellite...`) is v6 with API v2, and this org-specific grouping logic existed to support only v6 API v1. I would _love_ to have access to a v5 instance and a v6 API v1 instance as well. 🤷 Any idea where I could get some?

However, I believe the un-grouping here _should_ be safe based on how I read the existing implementation. In the old code, `request_results` was called once for each `org_id`, and after that function call, nothing else directly referenced `org_id` before feeding lists into the pooled `request_host_details` calls. In the new code, `request_results` is still being called once for each `org_id`. If the same host exists in two orgs, and if `request_results` returns a different object for them, then both objects are still kept. However, if `request_results` returns the same object for them, the new code removes the duplicate and only calls `request_host_details` once for that host.

If you look deeper into `request_host_details`'s existing code currently on master, it explicitly _ignores_ the `org_id` value, shoving `org_id=None` into all Satellite API calls: https://github.com/quipucords/quipucords/blob/e671635b3253b48e7ed3e14a2cc931f9de85e719/quipucords/scanner/satellite/six.py#L299-L387

Assuming those `org_id=None` args aren't a bug, then there appears to be no reason to differentiate the calls to `request_host_details` based on different `org_id` values.

If I've overlooked something, please let me know! That's also likely, considering the complexity of this code. 🙂  I'd also love to confirm my expectations against a real Sat v6 API v1 server; my entire understanding of this logic is based on reverse-engineering the existing code.

bruno-fs: >I would love to have access to a v5 instance and a v6 API v1 instance as well. shrug Any idea where I could get some?

I've no idea. This is the reason I tend to treat satellite code as broken until proven otherwise 😆 

>If I've overlooked something, please let me know! 

I think you are right. maybe ord_id had a use in the past and then was completely ignored - or it's a half-baked feature. I'll need some time digging through blames to understand this 🤔 💭

bruno-fs (approved):
Amazing refactor! 👏🏽

PR #2320, opened 2023-04-12
fix(CredentialSerializer): remove custom "strict" validation

bruno-fs:
The extra validation for non expected keys was both poorly ported from the original CredentialSerializer implementation and unnecessary. DRF Serializers default behavior is to completely ignore unexpected data, so the validation was only required when everything was just one huge serializer.

updates DISCOVERY-287

context here: https://url.corp.redhat.com/95032f2


bruno-fs:
changing the base to master. while we don't commit anything there that's not supposed to go to 1.2 we treat it as release branch 1.2

nicolearagao (approved):
LGTM 👍🏼

PR #2323, opened 2023-04-14
dedupe satellite code

infinitewarp:
This PR is an effort to reduce lots of old copy-paste duplicated code. I am trying to bring more sanity to the Satellite modules before I resume the Celery conversion so that there will be fewer things to update.


bruno-fs (approved):
This PR is so DRY I got thirsty

stay hidraded meme

PR #2324, opened 2023-04-16
chore: Add support for the QPC_SERVER_PORT environment variable.

abellotti:
chore: Add support for the QPC_SERVER_PORT environment variable.

As we need need to be more OCP/CRC/Kubernetes friendly, let's add support for the QPC_SERVER_PORT Env so Templates can drive what port we listen to. Openshift does not like services to run under port 1000, so we need to pick up a larger number. OCP/CRC template can now define the QPC_SERVER_PORT in the environment as needed (i.e.  8443).


deploy/gunicorn.conf.py
```diff
@@ -3,7 +3,8 @@
 import os
 
 # pylint: disable=invalid-name
-bind = "0.0.0.0:443"
+QPC_SERVER_PORT = int(os.getenv("QPC_SERVER_PORT", "443"))
```
infinitewarp, line 6: Since we are increasing the number of environment variables here, and I imagine more will be configurable in the future, I'd suggest adding `environ` and using that here and on line 12. It has more robust type handling and error reporting.

```
import environ
env = environ.Env()
```

```suggestion
QPC_SERVER_PORT = env.int("QPC_SERVER_PORT", 443)
```

```
timeout = env.int(QPC_SERVER_TIMEOUT", 120)
```

infinitewarp (approved):

PR #2327, opened 2023-04-17
DISCOVERY-287 - strict credential validation

bruno-fs:
Restore the very weird middle-ground validation we had for CredentialSerializer:
Don't allow extra fields UNLESS they are set to `None`.

I personally would rather have a FULL STRICT validation or keep DRF default, but CLI would need to change accordingly.


infinitewarp (approved):
Nice to see https://github.com/quipucords/quipucords/pull/2320 back from the dead so soon. 😅

PR #2336, opened 2023-04-20
perf(CredentialSerializer): avoid subclassing logic when not needed

bruno-fs:
NOTE: sorry for the ridiculous branch name. yet another credential serializer god knows how many at this point


infinitewarp (approved):
I've not been a big fan of the `new` overrides we have to use here, but I understand our options are limited due to the nature of DRF, and I approve of this optimization! 👍

PR #2339, opened 2023-04-22
chore: adding support for DJANGO_SECRET_KEY env variable.

abellotti:
chore: adding support for DJANGO_SECRET_KEY env variable.

This is necessary when the application keys are coming from OpenShift project secrets through the environment


abellotti:
Ran tests locally fine.

After some hair-pulling, realized why the re-scans (network) were failing.  When testing with a new secret, secrets get saved (secret.txt) and used to encrypt/decrypt data as sent to us via ansible (we pass in the path to the secret.txt).  If that puppy changes, decryption will fail. So re-tested with the new key and new database, and all the stars were aligned again.

bruno-fs (approved):
LGTM

this reminds me we need to discuss whether we will create "new tests" using pytest or unittest.

There's a PR with a open discussion on this topic #2287
I also wrote a quick [comparison between pytest/unittest asserts ](https://bruno-fs.github.io/posts/test-assertion-styles/)

PR #2350, opened 2023-05-06
fix: deduplicate installed products

nicolearagao:
Since product ids are unique identifiers, check for already saved product ids before saving new ones to avoid duplication.


quipucords/scanner/network/processing/test_installed_products.py
```diff
@@ -67,6 +67,16 @@ def test_success_multiple_products():
         \tName: Red Hat Enterprise Linux Server
         \tVersion: 8.7
     --
+    Product:
+        \tID: 69
+        \tName: Red Hat Enterprise Linux Server
+        \tVersion: 8.7
+    --
```
infinitewarp, line 74: Is there a test somewhere that asserts the presence of this value? I'm curious why it was added.

infinitewarp (approved):

PR #2351, opened 2023-05-08
update incorrect docstrings, remove unnecessary saves

infinitewarp:


quipucords/api/scanjob/model.py
```diff
@@ -576,9 +546,8 @@ def complete(self):
         self.log_current_status()
 
     def fail(self, message):
-        """Fail a job.
+        """Change status from RUNNING to COMPLETE.
```
abellotti, line 549: shouldn't this be from RUNNING to FAILED.?

infinitewarp: Good catch! Fixed and pushed.

PR #2353, opened 2023-05-09
split up job runner run method

infinitewarp:
This PR moves some code from `SyncScanJobRunner` to regular functions so that we can reuse them later with Celery tasks.

I'm also tidying up a little bit of related code along the way. It may help to look at the individual commit diffs.

### DIFF
--- a/quipucords/api/scan/tests_scan.py
+++ b/quipucords/api/scan/tests_scan.py
@@ -644,30 +644,27 @@ class TestScanList(TestCase):
             become_password=None,
             ssh_keyfile=None,
         )
-        self.cred.save()
         self.cred_for_upload = self.cred.id
 
-        self.source = Source(name="source1", source_type="network", port=22)
-        self.source.save()
+        self.source = Source.objects.create(
+            name="source1", source_type="network", port=22
+        )
         self.source.credentials.add(self.cred)
-        self.source.save()
 
-        self.test1 = Scan(name="test1", scan_type=ScanTask.SCAN_TYPE_INSPECT)
-        self.test1.save()
+        self.test1 = Scan.objects.create(
+            name="test1", scan_type=ScanTask.SCAN_TYPE_INSPECT
+        )
         self.test1.sources.add(self.source)
-        self.test1.save()
 
-        self.test2 = Scan(name="test2", scan_type=ScanTask.SCAN_TYPE_CONNECT)
-        self.test2.save()
+        self.test2 = Scan.objects.create(
+            name="test2", scan_type=ScanTask.SCAN_TYPE_CONNECT
+        )
         self.test2.sources.add(self.source)
-        self.test2.save()
 
         # self.test1 will not have a most_recent_scanjob, self.test2
         # will.
-        job = ScanJob(scan=self.test2)
-        job.save()
+        job = ScanJob.objects.create(scan=self.test2)
         job.sources.add(self.source)
-        job.save()
 
         self.test2.most_recent_scanjob = job
         self.test2.save()
--- a/quipucords/scanner/job.py
+++ b/quipucords/scanner/job.py
@@ -227,10 +227,7 @@ class SyncScanJobRunner:
                 # something went wrong or cancel/pause
                 return task_status
 
-        if self.scan_job.scan_type in [
-            ScanTask.SCAN_TYPE_INSPECT,
-            ScanTask.SCAN_TYPE_FINGERPRINT,
-        ]:
+        if self.scan_job.scan_type != ScanTask.SCAN_TYPE_CONNECT:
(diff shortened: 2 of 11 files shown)


quipucords/api/scan/tests_scan.py
```diff
@@ -644,30 +644,27 @@ def setUp(self):
...
         self.source.credentials.add(self.cred)
-        self.source.save()
 
-        self.test1 = Scan(name="test1", scan_type=ScanTask.SCAN_TYPE_INSPECT)
-        self.test1.save()
+        self.test1 = Scan.objects.create(
+            name="test1", scan_type=ScanTask.SCAN_TYPE_INSPECT
+        )
         self.test1.sources.add(self.source)
-        self.test1.save()
```
abellotti, line 658: don't we need the saves after adding the Many to Many relationships ? saw examples here https://docs.djangoproject.com/en/4.2/topics/db/examples/many_to_many/ doing the save.

quipucords/scanner/job.py
```diff
@@ -37,6 +37,112 @@ def get_task_runner_class(scan_task: ScanTask):
...
+        # truly unforeseen exception type should end up in here.
+        failed_task = runner.scan_task
+        context_message = (
+            "Unexpected failure occurred. See context below.\n"
+            f"SCAN JOB: {runner.scan_job}\nTASK: {failed_task}\n"
+        )
+        if failed_task.scan_type != ScanTask.SCAN_TYPE_FINGERPRINT:
+            creds = [str(cred) for cred in failed_task.source.credentials.all()]
+            context_message += f"SOURCE: {failed_task.source}\nCREDENTIALS: [{creds}]"
+        failed_task.fail(context_message)
```
abellotti, line 117: are the creds from the line above being logged with this failure ?

quipucords/api/scan/tests_scan.py
infinitewarp: https://docs.djangoproject.com/en/4.2/ref/models/relations/

> Note that add(), aadd(), create(), acreate(), remove(), aremove(), clear(), aclear(), set(), and aset() all apply database changes immediately for all types of related fields. In other words, there is no need to call save()/asave() on either end of the relationship.

🪄 ✨

quipucords/scanner/job.py
```diff
@@ -37,6 +37,29 @@ def get_task_runner_class(scan_task: ScanTask):
     raise NotImplementedError
 
 
+def get_task_runners_for_job(
+    scan_job: ScanJob,
+) -> tuple[list[ScanTaskRunner], FingerprintTaskRunner | None]:
+    """Get ScanTaskRunners for all the tasks that need to run for this job."""
+    incomplete_scan_tasks = scan_job.tasks.filter(
```
nicolearagao, line 44: dummy question. why using Q here? couldnt it be:
```
    incomplete_scan_tasks = scan_job.tasks.filter(
        status__in=[ScanTask.RUNNING, ScanTask.PENDING]
    ).order_by("sequence_number")
```

quipucords/api/scan/tests_scan.py
infinitewarp: FWIW, that `Note` was relatively new information for me too. I only somewhat recently found that comment, and I've been taking advantage of that knowledge for cleanups like this. I wonder if that automatic save was not always true for older versions of Django (I've been Django-ing since at least 1.x…), but it does seem to work consistently now.

infinitewarp: LOL… well, that note has been there since the oldest version of the docs for Django 1.8. I guess I just never noticed it until recently! https://docs.djangoproject.com/en/1.8/ref/models/relations/

abellotti: Ahh, thank you for that Note.  Could have been a new addition to Django, and the examples I saw were not updated accordingly.

nicolearagao: At least for create() this hasnt changed. 
https://docs.djangoproject.com/en/4.2/ref/models/querysets/#django.db.models.query.QuerySet.create
https://docs.djangoproject.com/en/1.8/ref/models/querysets/#django.db.models.query.QuerySet.create

quipucords/scanner/job.py
abellotti: Ahh good, thanks for checking.  Glad to see it's just the id. 👍

quipucords/scanner/job.py
infinitewarp: Updated and pushed! Let me know if you spotted other filters like this, and I'm happy to update.

nicolearagao (approved):
LGTM 👍🏼

PR #2355, opened 2023-05-10
chore: upgrade to ubi9.2 and python3.11

infinitewarp:


Dockerfile
```diff
@@ -15,28 +15,29 @@ ENV PYTHONPATH=/app/quipucords
...
         tar \
         which \
         ${BUILD_PACKAGES} \
         -y &&\
     dnf clean all &&\
-    python3 -m venv /opt/venv
+    python3.11 -m venv /opt/venv
 
-RUN pip install --upgrade pip wheel
+RUN pip3.11 install --upgrade pip wheel
```
bruno-fs, line 36: because we are using a virtualenv, this change should not be required

Dockerfile
```diff
@@ -62,7 +63,7 @@ VOLUME /var/data
 COPY . .
 
 # Install quipucords as package
-RUN pip install -v -e .
+RUN pip3.11 install -v -e .
```
bruno-fs, line 66: because we are using a virtualenv, this change should not be required

PR #2356, opened 2023-05-15
rename model status-setting methods

infinitewarp:
This is a legibility improvement, IMO. Readability counts, and explicit is better than implicit.

Previously, `ScanJob` and `ScanTask` had methods like `start` and `fail`, but those methods don't _actually_ start or fail their respective jobs and tasks. They simply update (with conditional sanity-checking) their _status values_ in the database. Other code is responsible for actually starting, failing, etc. the functions that perform work for these jobs and tasks.

These new names still aren't perfect, but I think they are much clearer than the old names and should cause me (and hopefully anyone else new to our code) less confusion.

### DIFF
--- a/quipucords/api/scan/view.py
+++ b/quipucords/api/scan/view.py
@@ -253,7 +253,7 @@ class ScanViewSet(ModelViewSet):
                 )
                 if jobs_to_cancel:
                     for job in jobs_to_cancel:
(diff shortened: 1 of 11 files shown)


nicolearagao (approved):

bruno-fs (approved):
lgtm

PR #2357, opened 2023-05-17
chore: bumping github action to python 3.11

abellotti:
chore: bumping github action to python 3.11

Bumping python version in github actions for make test to python 3.11


bruno-fs (commented):
lgtm

PR #2358, opened 2023-05-17
refactor: gracefuly handle when interrupt signal is not set

infinitewarp:
As we migrate functionality into Celery tasks, we will no longer be passing multiprocessing `Value` interrupt-handling objects throughout the system. We will need a completely different mechanism to stop processing activity in asynchronous Celery workers.

I introduced this change in a local branch where I am working on those Celery tasks, but I pulled it out to this separate branch/PR for early review in isolation. For now, I am simply wrapping all uses of the interrupt object with conditional checks. If it's not set, we simply proceed as though we're not being interrupted.


bruno-fs (approved):
lgtm 👍🏽

PR #2359, opened 2023-05-17
Poc/map raw facts to insights

nicolearagao:
Captura de tela de 2023-05-17 18-06-51

### DIFF
--- a/quipucords/api/common/entities.py
+++ b/quipucords/api/common/entities.py
@@ -40,6 +40,7 @@ class HostEntity:
     """A proxy over SystemFingerprint with some extra steps."""
 
     _fingerprints: SystemFingerprint
+    _raw_facts: dict
     last_discovered: datetime
 
     def __getattr__(self, attr):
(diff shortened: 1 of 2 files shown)


quipucords/api/common/entities.py
```diff
@@ -287,8 +293,13 @@ def last_discovered(self) -> datetime:
     @classmethod
     def from_report_id(cls, report_id, skip_non_canonical=True):
         """Create a Report from a report_id."""
-        deployment_report, fingerprints = cls._get_deployment_report(report_id)
-        hosts = [HostEntity(f, deployment_report.last_discovered) for f in fingerprints]
+        deployment_report, fingerprints, raw_facts = cls._get_deployment_report(report_id)
+        hosts = []
+        for fingerprint in fingerprints:
+            matching_facts = next((f for f in raw_facts if f.get('etc_machine_id') == fingerprint.etc_machine_id), None)
```
bruno-fs, line 299: I found it too arbitrary just select one set of raw facts when the fingerprint might have been generated by merging multiple sets of raw facts.

I'd rather see:

1.  a function that will search for a raw fact iterating over a list of dicts, searching in each one of them the unmapped attribute
2. instead querying for raw facts on details report and then of matching for `etc_machine_id`, I'd rather see a change in fingerprint generation: it should be the raw facts that contributed in its creation and (preferably ordered in the same way they were prioritized during fact selection).

PR #2360, opened 2023-05-18
Deprecate QPC_DISABLE_AUTHENTICATION

bruno-fs:
Remove the hidden feature used for tests that disables authentication.

Since views without authentication are less numerous than the ones requiring authentication (from the top of my head I can only think of the login and status views), in the near future we should probably set auth+permission classes globally and just disable them in the views that don't need it.


infinitewarp (approved):
Excellent! I love the solution of the client fixture plus the logged in mixin class.

PR #2361, opened 2023-05-18
more fixes

infinitewarp:
Here's another catch-all bundle of fixes and cleanups of things I've encountered while preparing Celery tasks in another branch. You may want to review by individual commits.

### DIFF
--- a/quipucords/api/reports/view.py
+++ b/quipucords/api/reports/view.py
@@ -25,9 +25,7 @@ from api.reports.reports_gzip_renderer import ReportsGzipRenderer
 from api.serializers import DetailsReportSerializer
 from api.user.authentication import QuipucordsExpiringTokenAuthentication
 
-# Get an instance of a logger
-# pylint: disable=invalid-name
-logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+logger = logging.getLogger(__name__)
 
 auth_classes = (QuipucordsExpiringTokenAuthentication, SessionAuthentication)
 perm_classes = (IsAuthenticated,)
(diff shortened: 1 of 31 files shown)


quipucords/scanner/job.py
```diff
@@ -204,7 +204,25 @@ def check_manager_interrupt(self) -> str | None:
...
+        - run the inspection tasks (typically 0 or 1 per source)
+        - check status for each task
+            - remember any that failed to be logged later
+            - early return if any are not failed or complete
+        - attempt to get/create the details report
+            - save a reference of it to both the job and the fingerprint task
+            - early return if this fails
+        - run the fingerprint task (only one per job)
+        - log individual IDs of any failed tasks
+
```
bruno-fs, line 223: <3 ty for adding this <3

quipucords/api/reports/view.py
```diff
@@ -41,25 +41,27 @@
...
+def reports(request, report_id=None):
     """Lookup and return reports."""
     reports_dict = {}
     mask_report = request.query_params.get("mask", False)
-    if pk is not None:
-        if not is_int(pk):
+    if report_id is not None:
+        if not is_int(report_id):
             error = {"report_id": [_(messages.COMMON_ID_INV)]}
             raise ValidationError(error)
```
bruno-fs, line 51: Every time I see this pattern I want to remove it because:

1. there's no URL mapped to this view without pk/report_id
2. weird HTTP semantics: how NOT FOUND is a 400? LOL

Don't need to remove it, tho. Just wanted to point how weird this is

bruno-fs (approved):
amazing! 👍🏽

quipucords/api/reports/view.py
infinitewarp: Oh yes, I wanted to refactor this when I saw it too, but I had to make myself stop or I'd be rewriting everything all the time. 😂 Glad we agree, though!

abellotti (approved):

PR #2363, opened 2023-05-19
Mock "mask value" utility

bruno-fs:
Mock "mask value" utility and get rid of now useless env vars for testing


infinitewarp (approved):
Clever solution.

PR #2364, opened 2023-05-19
celery scan manager

infinitewarp:
Say hello to the latest prototype of Celery-powered scan job runners! 👋

This is _**not 100% feature complete**_ compared to existing `Manager`/`ScanJobRunner` implementations. Specifically:

- It does not yet support stopping, pausing, or resuming running tasks.
- I have not exhaustively tried error/failure/edge cases.

However, it does have a few good things in its favor:

- It offloads the scan work from the main web server process/container to a separate Celery worker.
- It works (in my manual happy path testing) for scans with network, OpenShift, Ansible, Satellite v6, and vCenter type sources.
- It runs the connect and inspect tasks in parallel if multiple sources are defined for a scan.

To use this new functionality with local quipucords:

- Set and export `QPC_ENABLE_CELERY_SCAN_MANAGER=True` in your environment (or add to your `.env` file if you use it).
- Install and start `redis`. If not on the default `localhost:6379`, set `REDIS_HOST`, `REDIS_PORT`, `REDIS_USERNAME`, and `REDIS_REDIS_PASSWORD` as appropriate.
- Run quipucords in one shell: `poetry run make serve`
- Run the Celery workers in another shell: `poetry run make celery-worker`
- Use the UI or CLI as usual!

Most of the new Celery tasks currently return a tuple that you can see in the Celery worker logs like this:

```
INFO Task scanner.tasks.fingerprint[d3a94d18-38a8-49d9-bbd4-7203d00fbe5a] succeeded in 1.926941808022093s: (True, 481, 'completed')
```

Nothing currently _uses_ the return value from these tasks, but I suspect these could be useful in the future as we iterate and make the task chains smarter. Right now, though, it's just logging fluff.


quipucords/scanner/job.py
```diff
@@ -147,11 +149,90 @@ class ScanJobRunner:
...
+            job__id=self.scan_job.id,
+            scan_type=ScanTask.SCAN_TYPE_FINGERPRINT,
+            status__in=[ScanTask.RUNNING, ScanTask.PENDING],
+        ).first():
+            task_chain |= fingerprint.si(scan_task.id)
+
+        # Add a final post-processing task to the chain.
+        task_chain |= finalize_scan.si(self.scan_job.id)
+        task_chain.apply_async()
+
```
bruno-fs, line 234: <3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3<3

quipucords/scanner/tasks.py
```diff
@@ -0,0 +1,117 @@
...
+
+    :returns: tuple containing success bool, scan task id, and scan task status
+    """
+    scan_task = ScanTask.objects.get(id=scan_task_id)
+    runner_class = get_task_runner_class(source_type, scan_type)
+    runner: ScanTaskRunner = runner_class(scan_task.job, scan_task)
+    success = (task_status := run_task_runner(runner)) == ScanTask.COMPLETED
+    if not success:
+        # If this task failed, we do not want subsequent tasks in its chain to run.
+        self.request.chain = None
```
bruno-fs, line 50: didn't knew this trick. Neat!

quipucords/scanner/tasks.py
```diff
@@ -0,0 +1,117 @@
...
+
+def get_task_runner_class(source_type, scan_type) -> type[ScanTaskRunner]:
+    """Get the ScanTaskRunner class type for the given source and scan types.
+
+    :returns: runner class type (not an instance of the class)
+    """
+    if scan_type == ScanTask.SCAN_TYPE_CONNECT:
+        runner_class_name = "ConnectTaskRunner"
+    elif scan_type == ScanTask.SCAN_TYPE_INSPECT:
+        runner_class_name = "InspectTaskRunner"
```
bruno-fs, line 24: Not a blocker for this PR, but we will might need to support SCAN_TYPE_FINGERPRINT

we have the following usecases:
- POST /api/v1/reports/ https://github.com/quipucords/quipucords/blob/590cd36ea970fb59a9c37549ce1559f85e7f2a11/quipucords/api/details_report/view.py#L96-L98
- PUT /api/v1/reports/merge/ merge https://github.com/quipucords/quipucords/blob/590cd36ea970fb59a9c37549ce1559f85e7f2a11/quipucords/api/merge_report/view.py#L54-L59
- PUT/POST /api/v1/reports/merge/jobs/ merge https://github.com/quipucords/quipucords/blob/590cd36ea970fb59a9c37549ce1559f85e7f2a11/quipucords/api/merge_report/view.py#L152-L161

From the quoted code, only the last one actually would require proper celery support as it was designed to run thread based manager. (I wish we could deprecate the other ones - first one could just upload data and push to celery for processing, w/o actually returning results)

bruno-fs (approved):
LGTM

quipucords/scanner/tasks.py
infinitewarp: Yes, thank you for providing that context! You may recall I asked about this last week, but I promptly forgot to research and confirm. 😅 I believe you are right; my code likely needs some minor changes to support this use case.

bruno-fs: You probably already thought that, but I'll say anyway: this particular topic is probably its own jira story "celery: support 'fingerprint' type scans" or smth like that.

infinitewarp: I've added this to the backlog to handle later: https://issues.redhat.com/browse/DISCOVERY-340

quipucords/scanner/satellite/api.py
```diff
@@ -126,7 +130,53 @@ def _prepare_host_request_options(self):
...
+
+        :param hosts: iterable of host dicts
+        :param request_host_details: API version-specific function to get host details,
+            must also be a registered Celery task
+        :param process_results: API version-specific function to process results
+        """
+        all_prepared_hosts = self.prepare_host(hosts, ids_only=True)
+        # Note: the second argument for `chunks`
+        # pylint: disable=fixme
+        # TODO should we use something other than self.max_concurrency to drive this?
```
bruno-fs, line 162: do we even need to limit? the client could have handle rate limit errors (assuming satellite can emit those).

IMO max concurrency is only relevant to network scan

infinitewarp: 🪓'd

quipucords/scanner/job.py
```diff
@@ -147,11 +150,90 @@ class ScanJobRunner:
...
+            scan_type = scan_task.scan_type
+            signature = celery_run_task_runner.si(scan_task_id, source_type, scan_type)
+            celery_signatures_by_source[source_id].append(signature)
+
+        # Start a chain with a group that contains n chains, one chain for each Source.
+        # Each chain contains the tasks required for that Source. Typically, this will
+        # contain one "connect" task and one "inspect" task.
+        task_chain = celery.group(
+            celery.chain(*sigs) for sigs in celery_signatures_by_source.values()
+        )
```
bruno-fs, line 220: This is great for mimicking current thread-based manager behavior.

In a multi-source scenario we will only need to chain connect to inspect per source and have them running in parallel for each source (but we will still need to group them together to need to chain fingerprint in the end)

quipucords/scanner/satellite/api.py
```diff
@@ -126,7 +128,62 @@ def _prepare_host_request_options(self):
...
+        """
+        all_prepared_hosts = self.prepare_host(hosts, ids_only=True)
+        # At the time of implementation, we don't know if it's okay to create many
+        # tasks (one per hosts) in a group like this. If this proves problematic, we
+        # should consider replacing `group` with `chunks` and calculate a reasonable
+        # size to limit parallel load. For example:
+        #     chunk_size = ceil(len(all_prepared_hosts) / self.max_concurrency)
+        #     request_host_details.chunks(all_prepared_hosts, chunk_size)().get()
+        #     results = list(itertools.chain.from_iterable(results))
+        # FIXME We SHOULD NOT use disable_sync_subtasks=False.  # pylint: disable=fixme
```
bruno-fs, line 174: please drop the pylint directive

quipucords/scanner/satellite/api.py
```diff
@@ -126,7 +128,62 @@ def _prepare_host_request_options(self):
...
+        # Although unlikely in this case, this option in general may lead to deadlocks
+        # if there aren't enough available Celery workers.
+        results = (
+            celery.group(
+                request_host_details.s(*host_params)
+                for host_params in all_prepared_hosts
+            )
+            .apply_async()
+            .get(disable_sync_subtasks=False)
```
bruno-fs, line 183: Even reading the comments I didn't get why we need this parameter. Could you walk through the rationale here?

infinitewarp: And to be clear, `get` is always a blocking call with Celery, and that's the "problem" here. `get` blocks the current process until the broker says the related async task is done.

quipucords/scanner/satellite/api.py
infinitewarp: Dropped in an amended commit! This pylint directive never existed. 🫥

quipucords/tests/scanner/tasks/test_tasks.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+@override_settings(CELERY_TASK_ALWAYS_EAGER=True)
+@pytest.mark.django_db
+def test_finalize_scan_completed_no_failures():
+    """Test finalize_scan sets expected COMPLETED status if no tasks failed."""
+    scan_job = ScanJobFactory(status=ScanTask.RUNNING)
+    tasks.finalize_scan.delay(scan_job.id)
+    scan_job.refresh_from_db()
+    assert scan_job.status == ScanTask.COMPLETED
```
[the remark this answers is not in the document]
infinitewarp: I amended the commit that introduced this test to add another test and to explain with more detail what this does. Please take a look and tell me what you think.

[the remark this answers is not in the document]
bruno-fs: LGTM! =D

bruno-fs (approved):

quipucords/quipucords/celery.py
```diff
@@ -16,4 +18,20 @@
 app.config_from_object("django.conf:settings", namespace="CELERY")
 
 # Load task modules from all registered Django apps.
-app.autodiscover_tasks()
+task_packages = ["scanner", "scanner.satellite.five", "scanner.satellite.six"]
```
abellotti, line 21: why are we limiting to those only?

infinitewarp: It's the other way around. Because of the way quipucords is currently structured, the default `autodiscover_tasks` _isn't actually discovering_ all the tasks. So, I have to explicitly define these module paths for now. Eventually I hope to make this better and bring us back to using defaults.

PR #2369, opened 2023-05-24
test: changing credential types should fail

abellotti:
- Adding tests that verifies that changing credentials types fail with a 400.
  - Testing Network credential -> OpenShift type
  - Testing vCenter credential -> Network type


nicolearagao (approved):
LGTM

PR #2370, opened 2023-05-24
refactor: deduplicate multiprocessing code in satellite scanner modules

infinitewarp:
Moving, deduplicating, and encapsulating the `multiprocessor.Pool` code in the Satellite scanners will make it easier to override and replace with Celery tasks in the near future.


bruno-fs (approved):
LGTM

PR #2373, opened 2023-05-26
test: Test failures for all cred type updates

abellotti:
test: Test failures for all cred type updates

Added tests so we test that changing cred_type for all of the DataSource types will return in a failure.


bruno-fs (approved):
pytest >>>>>> unittest

🕺🏽 💃🏽

PR #2374, opened 2023-05-26
test: using unittest_parametrize for credentials tests

abellotti:
test: using unittest_parametrize for credentials tests

- this allows us to run the parametrized Django tests within PyCharm

Wasn't happy with https://github.com/quipucords/quipucords/pull/2373 as I could not run the tests within PyCharm. Kicking the tires here with unittest-parametrize.


abellotti:
Well, I can do the PyCharm RunTest on the class CredentialTest and the new tests will also run:

image

abellotti:
On-Hold and closing for now, we first need to decide which Path to go and stick with (pytest or unittest). We can revisit later.

PR #2375, opened 2023-05-26
refactor: reorganize test files

infinitewarp:
Feeling spicy today. 🔥 How about cleaning up our project structure and putting all the tests in one place?

This was largely a three-step process.

### Step 1: rename `tests_*` files to `test_*`.

I did this with a simple `find` command, specifically this:

```
find quipucords -name 'tests_*.py' -exec bash -c 'NAME={}; git mv $NAME $(echo $NAME | sed "s/tests_/test_/g")' \;
```

### Step 2: move `test_*` files out of the main code packages

...and into the `tests` package path. I wrote a crude one-off Python script to do this since I would have to `mkdir` and `touch __init__.py` a lot. Here is the content of that script; I haven't committed it anywhere since it's one-time-use throwaway code:

```python
from pathlib import Path, PurePath

repo_root_path = Path.home() / "projects/quipucords"
code_root_path = repo_root_path / "quipucords"
tests_root_path = code_root_path / "tests"

code_root_purepath = PurePath(code_root_path)
tests_root_purepath = PurePath(tests_root_path)

test_paths = list(code_root_path.glob("**/test_*.py"))
for test_path in test_paths:
    test_pure_path = PurePath(test_path)
    if test_pure_path.is_relative_to(tests_root_purepath):
        continue
    test_rel_pure_path = test_pure_path.relative_to(code_root_purepath)
    new_test_path = tests_root_path / test_rel_pure_path.parent / test_pure_path.name
    new_test_path.parent.mkdir(parents=True, exist_ok=True)
    (new_test_path.parent / "__init__.py").touch()
    test_path.rename(new_test_path)
```

### Step 3: Manually fix a few issues.

Some test files had relative imports like `from . import vault` that were no longer valid and needed fixing. Some files weren't caught by the above script, and I had to manually move them (like `test_data/api_resource_list.json`).

That's basically it! 

I locally ran `poetry run make lint test test-integration` to confirm that nothing broke. Looks good so far. 🤞


nicolearagao (approved):
Right on the money 👏🏼

pyproject.toml
```diff
@@ -127,9 +127,3 @@ addopts = [
     "--block-network",
     "--allowed-hosts=127.0.0.1,::1",
 ]
-python_files = [
-    "tests.py",
-    "test_*.py",
```
abellotti, line 132: how come we're losing the test_*.py above signatures ?  don't we need this for the new location/files ?

bruno-fs: not anymore =D

bruno-fs: pytest defaults covers what we have now. ty @infinitewarp <3

abellotti (approved):

bruno-fs (approved):
Explendid job!

PR #2380, opened 2023-05-30
lint: replace flakeheaven with ruff

bruno-fs:
ruff is waaaaay faster and has all flake8&friends checks. It doesn't have all pylint checks, but it won't execute our code during linting so that's a + to me.


quipucords/api/common/common_report.py
```diff
@@ -55,7 +55,7 @@ def extract_tar_gz(file_like_obj):
         tar_name = f"/tmp/api_tmp_{time.strftime('%Y%m%d_%H%M%S')}.tar.gz"
         with open(tar_name, "wb") as out_file:
             out_file.write(file_like_obj)
-        tar = tarfile.open(tar_name)  # pylint: disable=consider-using-with
+        tar = tarfile.open(tar_name)
```
infinitewarp, line 58: I figure we don't have time to fix everything that this PR touches, but _yikes_… 😬

I will 🙈 and pretend I didn't see this (and probably other disabled violations that should never have been disabled).

quipucords/fingerprinter/runner.py
```diff
@@ -106,7 +101,7 @@ def format_certs(redhat_certs):
         """
         try:
             return [int(cert.strip(".pem")) for cert in redhat_certs if cert]
-        except Exception:  # pylint: disable=broad-except
+        except Exception:
             return []
```
infinitewarp, line 105: 😢 

Again, not saying we need to fix things like this right now, but this one made me _oof_.

infinitewarp (approved):
SHIP IT.

quipucords/quipucords/environment.py
```diff
@@ -109,7 +107,7 @@ def log_all_environment_variables():
     env_list = []
     for key, value in os.environ.items():
         if "password" in key.lower():
-            value = "*" * 8
+            value = "*" * 8  # noqa: PLW2901
```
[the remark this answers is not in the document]
bruno-fs: rofl

PR #2384, opened 2023-06-02
fix: skip test_openshift_source_update since it does not make sense

infinitewarp:
This change should resolve the following warning emitted during test runs:

    PytestUnknownMarkWarning: Unknown pytest.mark.xpass - is this a typo?
    You can register custom marks to avoid this warning - for details,
    see https://docs.pytest.org/en/stable/how-to/mark.html

2023-06-05 update: Instead of `xfail`, we decided simply to use `skip` on this test because we agree there is no value in attempting to run it. Another issue exists here to reconsider and fix the underlying implementation: https://issues.redhat.com/browse/DISCOVERY-352

### DIFF
--- a/quipucords/tests/api/source/test_serializer.py
+++ b/quipucords/tests/api/source/test_serializer.py
@@ -108,7 +108,7 @@ def test_openshift_source_default_port(openshift_cred_id):
     assert serializer.validated_data["port"] == 6443
 
 
-@pytest.mark.xpass("hosts ('5.4.3.2.1') is clearly invalid in this case")
+@pytest.mark.xfail(reason="hosts ('5.4.3.2.1') is clearly invalid in this case")
 @pytest.mark.django_db
 def test_openshift_source_update(openshift_source, openshift_cred_id):
     """Test if serializer updates fields correctly."""
--- a/quipucords/tests/api/source/test_serializer.py
+++ b/quipucords/tests/api/source/test_serializer.py
@@ -108,7 +108,10 @@ def test_openshift_source_default_port(openshift_cred_id):
     assert serializer.validated_data["port"] == 6443
 
 
-@pytest.mark.xfail(reason="hosts ('5.4.3.2.1') is clearly invalid in this case")
+@pytest.mark.skip(
+    "The code being tested is broken. '5.4.3.2.1' should not be a valid host address. "
+    "See also: https://issues.redhat.com/browse/DISCOVERY-352"
+)
 @pytest.mark.django_db
 def test_openshift_source_update(openshift_source, openshift_cred_id):
     """Test if serializer updates fields correctly."""


quipucords/tests/api/source/test_serializer.py
```diff
@@ -108,7 +108,7 @@ def test_openshift_source_default_port(openshift_cred_id):
     assert serializer.validated_data["port"] == 6443
 
 
-@pytest.mark.xpass("hosts ('5.4.3.2.1') is clearly invalid in this case")
+@pytest.mark.xfail(reason="hosts ('5.4.3.2.1') is clearly invalid in this case")
```
bruno-fs, line 111: I think the motivation for an "xpass" here (even though the marker does not work) is supposed to be broken, but it's not.

I'm not 100% sure if it should fail, tho. the assumption that it should fail comes from the premise that 5.4.3.2.1 was supposed to be an IP address BUT we do also accept URLs. Is this a valid URL? (I really don't know) If so, we can drop this mark. Otherwise, xfail is as wrong as the test is (since it should be testing for something invalid - OTH, if we really start a discussion on the merit of this negative test, it should actually be a for the IP address parsing function, not the serializer IMO.

TLDR: I don't even know this test should exist lol

mirekdlugosz, line 111: I don't understand. Shouldn't this test use [`pytest.raises`](https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions) to check that correct exception is raised when incorrect data is provided?

xfail is mostly intended for TDD (where you have a test for something that is not yet implemented) and heavily-parameterized tests where certain combinations are invalid.

bruno-fs: no exception is raised in this test :/

mirekdlugosz: I think such test is fine overall (putting aside details like whether it should be focused on IP parsing function or serializer) - I strongly prefer most negative / unhappy path tests to be at unit level, which allows us to focus on actual functionality at higher levels (e.g. in Camayoc).

This does not seem to be a valid URL. URL may contain "host", which must be IPv6 (put inside square brackets), IPv4 (a sequence of four numbers separated by dot) or "registered name", as defined by [RFC 1034](https://datatracker.ietf.org/doc/html/rfc1034#section-3.5). According to that, last DNS component must start with a letter.

So something like "5.4.3.2.a1" could be a valid URL, but "5.4.3.2.1" is definitely not. It's not a valid IPv4 and it's not a valid registered name.

(I _think_ it's common for apps and libraries to keep a list of valid and recognized TLDs, so they would reject "5.4.3.2.a1" anyway; but it is a valid syntax.)

bruno-fs: but I agree with you, if the test actually failed (as it should, given your point in the other comment), we should have no markers on this one and use pytest.raises on the statement that would raise the failure.

infinitewarp: The problem is `xpass` isn't even a valid marker. It's just apparently a random word we slapped on the test.

`xfail` means "this test _should_ fail but it's okay to expect it to pass" which I am pretty sure is the case for this bogus IP address.

infinitewarp: Correct. This test _should have always_ failed, but it _doesn't_ fail.

The "right" thing to do here would be to fix the underlying application logic so that invalid IP addresses are rejected. If we do that, then this test would _actually_ fail, prompting us to rewrite the test.

This is all just a result of bad old code. I haven't touched any of this code. It's just there, and this `xpass` "marker" has been wrong ever since it was implemented. I'm just fixing the marker to do the right thing in these circumstances.

The net result is that amid the `......` of the test output, this test will now say `X` instead of `.`. It will not change the result code of the test. It still "passes" but it a) no longer emits this warning message about an invalid marker, and b) clearly indicates that the test doesn't behave as it _should_.

quipucords/tests/api/source/test_serializer.py
```diff
@@ -108,7 +108,10 @@ def test_openshift_source_default_port(openshift_cred_id):
     assert serializer.validated_data["port"] == 6443
 
 
-@pytest.mark.xfail(reason="hosts ('5.4.3.2.1') is clearly invalid in this case")
+@pytest.mark.skip(
+    "The code being tested is broken. '5.4.3.2.1' should not be a valid host address. "
+    "See also: https://issues.redhat.com/browse/DISCOVERY-352"
+)
```
bruno-fs, line 114: nice

bruno-fs (approved):

PR #2389, opened 2023-06-06
chore: replacing isort with the isort rule enablement in ruff

abellotti:
chore: replacing isort with the isort rule enablement in ruff

- Removing the lint-isort directive
- Adding the "I" isort rule in ruff
- Fixing one inconsistency between the two.
- Removing isort package dependency


pyproject.toml
```diff
@@ -66,20 +65,6 @@ ruff = "^0.0.270"
...
-    "api",
-    "compat",
-    "constants",
-    "fingerprinter",
-    "log_messages",
-    "quipucords",
-    "scanner",
-    "tests",
-    "utils",
-]
```
bruno-fs, line 81: similar to qpc side, I believe we need the known first party  here
https://beta.ruff.rs/docs/settings/#known-first-party

PR #2390, opened 2023-06-08
Adding support for ssh keys

abellotti:
Adding support for SSH keys in the database. SSH Key values can now be added to a Credential via the API and not needing to reference an SSH Key file.
    
- Adding the 16KB encrypted ssh_keyv field to the Credential model.
- Updated the serializer to show the ssh_key, just like password, it is obfuscated.
- Updated Credential validator to check for ssh_key existence vs ssh_keyfile and password.
- ssh_keyfile or ssh_key need to be specified, not both.

Todo's
- [x] Updating code using the ssh_keyfile to also support ssh_keys for  communication
- [x] Manually test connectivity to scanned devices using ssh_key, both with and without a passphrase.
- [x]  While the private keys for ansible runner are properly set as user-only access, let's store the private keys in a separate ssh_keyfiles directory which we can simply zap the generated ssh keyfiles and related parent directories once emptied at the end of scans.
- [x] Prototype qpc CLI to leverage the new ssh_key parameter for adding an SSH key to a credential and verifying connectivity to scanned devices still works.  This is to make sure we're not blocking the CLI's enhancement for this. -> https://github.com/quipucords/qpc/pull/247
- [x] Adding tests

_Note_: while 16KB for the ssh_key would have been sufficient, Ansible vault seems to quadruple the space used for encryption.

_Note_: Couldn't use ye-old with tempfile... in our code for the private key file generation/cleanup since Ansible runs the scripts after the logic where we create them. We simply need to zap'em at the end of the scans (even though they're user-only accessible, they shouldn't stick around).


quipucords/api/credential/model.py
```diff
@@ -39,6 +39,7 @@ class Credential(models.Model):
     password = models.CharField(max_length=1024, null=True)
     auth_token = models.CharField(max_length=1024, null=True)
     ssh_keyfile = models.CharField(max_length=1024, null=True)
+    ssh_keyvalue = models.CharField(max_length=65536, null=True)
```
infinitewarp, line 42: I'm thinking about this `max_length`. Is 64K large enough for 99% of possible keys? I recall you mentioned earlier that you tested using a large key, but I don't recall exactly how big it was.

Internally, I believe this is a `VARCHAR` at the DB level, and Postgres is smart enough to store only what is needed. So, setting an absurdly large maximum length _should_ be safe as long as users aren't deliberately abusing the feature.

bruno-fs: VARCHAR maps to Django's TextField, right? I was going to ask if we couldn't simply use that 🤔

quipucords/api/messages.py
```diff
@@ -11,10 +11,18 @@
 
 # host credential messages
 PLURAL_HOST_CREDENTIALS_MSG = "Credentials"
-HC_PWD_OR_KEYFILE = "A host credential must have either a password or an ssh_keyfile."
-HC_NOT_BOTH = (
+HC_KEYFILE_OR_KEYVALUE = (
+    "A host credential must have either an ssh_keyfile or an ssh_keyvalue, not both."
+)
+HC_PWD_OR_KEYFILE_OR_KEYVALUE = (
+    "A host credential must have either a password, ssh_keyfile or an ssh_keyvalue."
```
infinitewarp, line 18:
```suggestion
    "A host credential must have a password, ssh_keyfile, or ssh_keyvalue exclusively."
```
Swapping "either" since that should be used only for a list of two items, and adding the oxford comma since our docs friends agreed to its use elsewhere.

quipucords/scanner/network/connect.py
```diff
@@ -301,8 +302,12 @@ def _connect(  # noqa: PLR0913, PLR0912, PLR0915
                 verbosity=verbosity_lvl,
             )
         except Exception as err_msg:  # noqa: BLE001
+            delete_ssh_keyfiles(inventory)
             raise AnsibleRunnerException(err_msg) from err_msg
 
+        # Let's delete any private ssh key files that we generated
+        delete_ssh_keyfiles(inventory)
```
infinitewarp, line 309: I'm not thrilled about the `delete_ssh_keyfiles` being explicitly called in multiple places. 😕  I wish we had a context manager that could do file cleanup automatically, but it looks like the loading and accessing of this data is so complexly intertwined (in legacy code, not just this new code), after walking through code for a few minutes, I couldn't find anywhere reasonably "clean" that could be wrapped like that.

I'll accept this since any alternative that I currently see would be _a lot_ more work, but I'm putting this here to see if anyone else has similar feelings or alternate suggestions

quipucords/scanner/network/utils.py
```diff
@@ -173,3 +181,54 @@ def get_fact_names():
...
+    """Remove a generated ssh_keyfile and delete its empty parent directory."""
+    keyfile_pattern = (
+        f"^/.*/{GEN_KEYFILES_DIRECTORY_PREFIX}.*/{GEN_KEYFILE_PREFIX}[0-9]+$"
+    )
+    if re.match(keyfile_pattern, private_keyfile_path):
+        if os.path.isfile(private_keyfile_path):
+            os.remove(private_keyfile_path)
+            parent_dir = os.path.abspath(os.path.join(private_keyfile_path, os.pardir))
+            if not os.listdir(parent_dir):
+                os.rmdir(parent_dir)
```
infinitewarp, line 234: I'm pretty sure this will break things unexpectedly in Celery workers when performing more than one scan, maybe also breaking the legacy threaded/multiprocessing manager implementation. Imagine two threads or processes on the same machine accessing keyfiles, and one deletes before the other has completed its work. 💥 

I believe we need to use something like [tempfile.TemporaryDirectory](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) and keep an in-memory reference to the directory so we can clean it up (assuming again that we can't convert this whole file handling process into a context manager, but I don't think that's happening any time soon).

quipucords/scanner/network/utils.py
```diff
@@ -173,3 +181,54 @@ def get_fact_names():
...
+    """Generate an ssh private keyfile for the ssh key content specified."""
+    private_keyfile_path = os.path.join(
+        tempfile.mkdtemp(prefix=GEN_KEYFILES_DIRECTORY_PREFIX),
+        f"{GEN_KEYFILE_PREFIX}{credential.get('id')}",
+    )
+    private_key_fp = open(private_keyfile_path, "w+")
+    private_key_fp.write(ssh_keyvalue)
+    private_key_fp.write("\n")
+    private_key_fp.close()
+    os.chmod(private_keyfile_path, 0o600)
```
infinitewarp, line 200: Mentioned in another comment, but adding here for local context. We can probably simplify and harden this code by using [tempfile.TemporaryDirectory](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) and [tempfile.NamedTemporaryFile](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile).

quipucords/scanner/network/utils.py
```diff
@@ -173,3 +181,54 @@ def get_fact_names():
...
+def generate_ssh_keyfile(credential, ssh_keyvalue):
+    """Generate an ssh private keyfile for the ssh key content specified."""
+    private_keyfile_path = os.path.join(
+        tempfile.mkdtemp(prefix=GEN_KEYFILES_DIRECTORY_PREFIX),
+        f"{GEN_KEYFILE_PREFIX}{credential.get('id')}",
+    )
+    private_key_fp = open(private_keyfile_path, "w+")
+    private_key_fp.write(ssh_keyvalue)
+    private_key_fp.write("\n")
+    private_key_fp.close()
```
infinitewarp, line 199: Also, whether we use `tempfile` or not, at least these lines should use the `open` context manager to better guarantee that `close` is always automatically called. Like
```suggestion
    with open(private_keyfile_path, "w+") as private_key_fp:
      private_key_fp.write(ssh_keyvalue)
      private_key_fp.write("\n")
```

quipucords/api/credential/model.py
abellotti: Yes, the CharField are Varchars internally and PG only allocates what it needs. I think I've tackled the largest (16kbit rsa), 
 some info in this comment: https://github.com/quipucords/quipucords/pull/2390/commits/da45b52ea7d860632953980d933732e71122febc.  No preference CharField vs TextField (both Varchars). The docs do mention using CharField for smaller data sizes. We could use a TextField with no max_length. Although we should still set an upper limit so that users don't abuse it.

quipucords/scanner/network/utils.py
infinitewarp: To clarify and expand on what I mean here, I think _each file_ we write to disk should have its own `NamedTemporaryFile` inside its own `TemporaryDirectory`.  We _could_ have them all share one `TemporaryDirectory`, but directories are cheap. We would need to retain in local memory a reference to the `NamedTemporaryFile` and `TemporaryDirectory` that we would use later for deletion.

Using `tempfile` objects like that would significantly reduce the risk of multiple threads/processes stomping on each others' files, and it means _we_ aren't responsible for choosing arbitrary path names. We can completely drop `GEN_KEYFILES_DIRECTORY_PREFIX` and `GEN_KEYFILE_PREFIX` since the names of these files don't really matter as long as we can pass the name to Ansible. Python's library is already smart enough to create _very_ unique names that will almost never conflict. It will actually test the name it generates, and it will keep trying a new random names if the last name it generated already exists.

abellotti: I think that's fine.  Other than the deletion upon context exit which we don't use, TemporaryDirectory is the same as (and actually uses) mkdtemp. for the filename itself, I had used the {GEN_KEYFILE_PREFIX}{credential.get('id')} pattern thinking of not creating thousands of files if our scans of 1K or 10K system use the same credential/ssh key (most likely).  Then again, if mkdtemp() is random, it's a moot point.  Hmm, Maybe we do need a unique directory for this process/thread/scan. thoughts ?

quipucords/api/credential/model.py
abellotti: Hmm, I did a Migration to TextField, even though max_length is taken by Django, this translate to just text in Postgres with no limit (other than the DB's 1GB limit).

I verified this by doing a Migration to TextField with the max_length set to 65536 and I had it here. DB just shows `text`:

image

And the following proving it's wide open:

```
In [3]: c = Credential.objects.get(id=26)

In [4]: c.ssh_keyvalue = "1234567890" * 100000

In [5]: c.save()
```

```
In [6]: d = Credential.objects.get(id=26)

In [7]: len(d.ssh_keyvalue)
Out[7]: 4050355
```

So, yea, here's a potential abuse of someone feeding 1MB in and Ansible encrypting bumping it up by 4. 

We need that max_length for our sanity so we're safer to stick with CharField.

quipucords/scanner/network/connect.py
abellotti: I know, I didn't like it either 😢 , luckily it was only the two places.

quipucords/scanner/network/utils.py
abellotti: If we're ok with thousands of SSH key files, we could stick them all in one directory (not random named) but with random filenames, at least they're not all over the place and the cleanup code will be simpler (no need to delete the parent directory), kind'a like a qpc .ssh directory but for dynamically generated files.

quipucords/scanner/network/utils.py
abellotti: From what it looks like, tempfile.TemporaryDirectory === tempfile.mkdtemp, so I think we're safe. But, yea going to random filenames would also protect this case.

quipucords/scanner/network/utils.py
infinitewarp: > Other than the deletion upon context exit which we don't use, TemporaryDirectory is the same

I think having the automatic cleanup on exit would still be good free extra protection in case we fail to delete any individual file. I just tested this locally to confirm that automatic cleanup does not actually require using `TemporaryDirectory` as a context manager. It looks like cleanup also happens on garbage collection on normal process shutdown:

```
>>> import tempfile
>>> d = tempfile.TemporaryDirectory()
>>> d.name
'/var/folders/71/5v1_8dbd03j_nfxbb8bkb1q00000gn/T/tmpq2zufxkw'
>>> from pathlib import Path
>>> Path(f'{d.name}/test.txt').touch()
```

Meanwhile in a separate shell while that REPL is still open, just to be extra sure:

```
❯ ls /var/folders/71/5v1_8dbd03j_nfxbb8bkb1q00000gn/T/tmpq2zufxkw
test.txt
```

and back to the first REPL:

```
>>> glob.glob(f"{d.name}/*")
['/var/folders/71/5v1_8dbd03j_nfxbb8bkb1q00000gn/T/tmpq2zufxkw/test.txt']
>>> ^d

❯ ls /var/folders/71/5v1_8dbd03j_nfxbb8bkb1q00000gn/T/tmpq2zufxkw
ls: /var/folders/71/5v1_8dbd03j_nfxbb8bkb1q00000gn/T/tmpq2zufxkw: No such file or directory
```

It's gone! 🎉 

`mkdtemp` does not do similar cleanup on exit:

```
>>> import tempfile
>>> d = tempfile.mkdtemp()
>>> d
'/var/folders/71/5v1_8dbd03j_nfxbb8bkb1q00000gn/T/tmpicmk21_c'
>>> from pathlib import Path
>>> Path(f'{d}/test.txt').touch()
>>> import glob
>>> glob.glob(f'{d}/*')
['/var/folders/71/5v1_8dbd03j_nfxbb8bkb1q00000gn/T/tmpicmk21_c/test.txt']
>>> ^d

❯ ls /var/folders/71/5v1_8dbd03j_nfxbb8bkb1q00000gn/T/tmpicmk21_c
test.txt
```

It's still there! 💥 

> then I used the {GEN_KEYFILE_PREFIX}{credential.get('id')} pattern thinking of not creating thousands of files

In a simpler case, I wouldn't oppose using a regular pattern like `{GEN_KEYFILE_PREFIX}{credential.get('id')}`, but with our scanning process, the risk here is that if you have two scans running in parallel using the same credentials but for maybe different sources, you run the risk of one deleting the file before the other is finished using it. So, we would have to get "clever" and add some randomness, at which point we might simply use full random name generation, thus `NamedTemporaryFile`.

I think having thousands of files in a directory should not be a problem here. As long as we're cleaning up after ourselves, those files should be relatively short-lived anyway, and I _doubt_ (but woudl love to be proven wrong) many of our current users actually have thousands of credentials or thousands of sources (not to be confused with thousands of hosts in a source) given how difficult that would be to define in the GUI or CLI. If I read our code correctly, and please correct me if I'm wrong, this new temporary file is written to disk effectively once per source, not once per host, which is good for keeping the file count low.

* once per source+credential combination:
  * connect task -> construct_inventory -> _construct_vars -> _credential_vars -> generate_ssh_keyfile
  * inspect task -> construct_inventory -> _construct_vars -> _credential_vars -> generate_ssh_keyfile

Is my reading here correct?

On a general note, I've worked on filesystems with upwards of 10k files in one directory without any major problems. Any problems I've seen typically arise from having too many _open_ file handles at once, but I don't think we're in that space here yet.

quipucords/scanner/network/utils.py
infinitewarp: Sorry for crossing the streams!  😆  See the _other_ comment thread with a local demo comparing `tempfile.TemporaryDirectory` against `tempfile.mkdtemp`. `TemporaryDirectory` gives an automatic cleanup on process shutdown as a nice bonus.

abellotti: Thank you for verifying 👍  from what I've read, except for the cleanup done with TemporaryDirectory, they are identical.

quipucords/scanner/network/utils.py
abellotti: That's what I thought so too (one per source), but it does create one per target host. So for now, I will update the code to create them all in one directory (no need to create/delete thousands of directories) but will move to NamedTemporaryFile for guaranteeing uniqueness and additional protection (their deletion upon process termination in case something got missed (error/exception)). 

I will push an update shortly.
Thank you.

abellotti: https://github.com/quipucords/quipucords/pull/2390/commits/48899b1ba6715b9de413e3e3c267225a1a70aae4

quipucords/scanner/network/utils.py
```diff
@@ -5,6 +5,7 @@
 import tempfile
 from functools import cache
 from multiprocessing import Value
+from pathlib import Path
```
bruno-fs, line 8: pathlib ftw

quipucords/scanner/network/utils.py
```diff
@@ -183,16 +184,17 @@ def raw_facts_template():
...
 
 
 def generate_ssh_keyfile(credential, ssh_keyvalue):
     """Generate an ssh private keyfile for the ssh key content specified."""
-    private_keyfile_path = os.path.join(
-        tempfile.mkdtemp(prefix=GEN_KEYFILES_DIRECTORY_PREFIX),
-        f"{GEN_KEYFILE_PREFIX}{credential.get('id')}",
-    )
+    ssh_pkey_dir = os.path.join(tempfile.gettempdir(), GEN_KEYFILES_DIRECTORY)
+    Path(ssh_pkey_dir).mkdir(parents=True, exist_ok=True)
```
bruno-fs, line 194: we can use just pathlib here (I'm almost enabling a ruff check for preferring pathlib XD)

```
    ssh_pkey_dir = Path(tempfile.gettempdir()) / GEN_KEYFILES_DIRECTORY
    ssh_pkey_dir.mkdir(parents=True, exist_ok=True)
```

quipucords/tests/api/credential/test_credential.py
```diff
@@ -18,14 +16,23 @@
...
+
+def alt_cred_type(cred_type):
+    """Given a credential type, return an alternate random one."""
+    cred_types = DataSources.values.copy()
+    cred_types.remove(cred_type)
+    return random.choice(cred_types)
+
+
+@pytest.mark.django_db
+class TestCredential(LoggedUserMixin):
```
bruno-fs, line 30: I'm almost sure we don't need `LoggedUserMixin` anymore. I created this mixin for unittest based tests.

quipucords/tests/api/credential/test_credential.py
```diff
@@ -18,14 +16,23 @@
...
     """Tests against the Credential model and view set."""
 
-    def setUp(self):
+    def setup_method(self, _test_method):
         """Create test case setup."""
         management.call_command("flush", "--no-input")
```
bruno-fs, line 35: now that this test is not a unittest anymore I think we can skip this setup.

pytest-django tests already teardown data between each test.

quipucords/tests/api/credential/test_credential.py
```diff
@@ -165,11 +211,74 @@ def test_hc_create_err_ssh_bad(self):
...
+        with tempfile.NamedTemporaryFile() as ssh_keyfile:
+            ssh_keyfile.write(b"\n")
+            data = {
+                "name": "cred1",
+                "username": "user1",
+                "cred_type": DataSources.NETWORK,
+                "ssh_keyfile": str(ssh_keyfile.name),
+                "ssh_keyvalue": "A" * 2048,
+            }
+            response = django_client.post(url, json=data)
```
bruno-fs, line 234: FYI there's a builtin pytest fixture for creation of temporary files: [tmp-path](https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmp-path-fixture). It's a random directory and pytest will take care of cleaning that up after a test is done. I like this fixture because you can save the files for debugging passing a parameter to pytest (`--basetemp=somedir` ).

I don't see the need to use `tmp_path` fixture here, but I thought it would be cool to share the info.

T

quipucords/tests/scanner/network/test_network_inspect.py
```diff
@@ -91,6 +92,22 @@ def setUp(self):
...
+        self.cred_ssh.save()
+        self.cred_ssh_data = model_to_dict(self.cred_ssh)
+
+        # setup source_ssh for scan
+        self.source_ssh = Source(name="source_ssh", port=22, hosts=["1.2.3.5"])
+        self.source_ssh.save()
+        self.source_ssh.credentials.add(self.cred_ssh)
+
+        self.host_ssh_list = [("1.2.3.5", self.cred_ssh_data)]
+
```
bruno-fs, line 110: while it's OK to have this here...
1. only the new tests depend on this setup: the "pytest way" to deal with it is creating a local fixture and have those tests consume it
2. both Credential and Source models have an equivalent factory boy... well, factory.

Take these as optional comments - especially 2.

bruno-fs (commented):
Nice job and THANK YOU for converting good old credential tests to pytest.

quipucords/tests/api/credential/test_credential.py
abellotti: Without this, the following 2 tests fail:

```
test_hostcred_update_view
test_hostcred_update_double
```

They assume id 2 for the update, which is true if things are torned down, but they are not somehow.  I will update the tests to use the response's id to be more reliable. But there may be something else not setup right.

quipucords/tests/api/credential/test_credential.py
abellotti: looks like `tmp_path` is already being used in this file, I'll update to use the same, we should be consistent.

quipucords/tests/scanner/network/test_network_inspect.py
abellotti: 👍  @pytest.fixture ftw,  nice!!

quipucords/scanner/network/connect.py
```diff
@@ -339,13 +339,18 @@ def _handle_ssh_passphrase(credential):
...
+        ssh_keyfile is not None or ssh_keyvalue is not None
     ):
-        keyfile = credential.get("ssh_keyfile")
         passphrase = decrypt_data_as_unicode(credential["ssh_passphrase"])
-        cmd_string = f"ssh-add {keyfile}"
+        if ssh_keyfile is not None:
+            cmd_string = f'ssh-add "{ssh_keyfile}"'
+        else:
+            ssh_keyvalue = decrypt_data_as_unicode(ssh_keyvalue)
+            cmd_string = f"/bin/bash -c 'echo \"{ssh_keyvalue}\n\" | ssh-add -'"
```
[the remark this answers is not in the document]
abellotti: Not that I recall seeing.  Just ran again with an ssh key with a passphrase (the above if case), and the output has those there, but no private keys shown:

```
<192.168.30.131> ESTABLISH SSH CONNECTION FOR USER: abellotti
<192.168.30.131> SSH: EXEC ssh -C -o ControlMaster=auto -o ControlPersist=60s -o StrictHostKeyChecking=no -o Port=22 -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o 'User="abellotti"' -o ConnectTimeout=120 -o 'ControlPath="/Users/abellotti/DevHome/init_scripts/.ansible/cp/803dbc9628"' -tt 192.168.30.131 'if [ -f /etc/machine-id ] ; then ls --full-time /etc/machine-id 2>/dev/null | grep -o '"'"'[0-9]\{4\}-[0-9]\{2\}-[0-9]\{2\}'"'"' ; fi'
<192.168.30.131> (0, b'2023-01-19\r\n', b'Shared connection to 192.168.30.131 closed.\r\n')
```

Also looked at /var/log/system.log and ~/.ansible/, not there either.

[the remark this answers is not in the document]
abellotti: also, looking at `pexpect.spawn`, it does not look like it would log unless the logfile parameter is specified. so I think we're good.  We can always change this code in the future to generate a temporary ssh_key file for running the ssh-add.

quipucords/scanner/network/utils.py
```diff
@@ -173,3 +182,58 @@ def get_fact_names():
...
+    keyfile_pattern = (
+        f"^{tempfile.gettempdir()}"
+        f"/{GEN_KEYFILES_DIRECTORY}/{GEN_KEYFILE_PREFIX}_[0-9]+_.*$"
+    )
+    return True if re.match(keyfile_pattern, private_keyfile_path) else False
+
+
+def delete_ssh_keyfile(private_keyfile_path):
+    """Remove a generated ssh_keyfile."""
+    if is_gen_ssh_keyfile(private_keyfile_path) and os.path.isfile(
```
infinitewarp, line 236: I would put `and os.path.isfile` inside the `is_gen_ssh_keyfile` so that `is_gen_ssh_keyfile` handles _all_ the relevant checks.

quipucords/scanner/network/utils.py
```diff
@@ -173,3 +182,58 @@ def get_fact_names():
...
+                delete_ssh_keyfile(ssh_pkf)
+
+
+def is_gen_ssh_keyfile(private_keyfile_path):
+    """Return True if the file path specified is a generated ssh keyfile."""
+    keyfile_pattern = (
+        f"^{tempfile.gettempdir()}"
+        f"/{GEN_KEYFILES_DIRECTORY}/{GEN_KEYFILE_PREFIX}_[0-9]+_.*$"
+    )
+    return True if re.match(keyfile_pattern, private_keyfile_path) else False
```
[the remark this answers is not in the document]
abellotti: It's a real check,  whether we specify an ssh_keyfile, or the actual ssh_keyvalue, both methods use the one and only Ansible mechanism that is file driven and driven by the `ansible_ssh_private_key_file` parameter.  We need to make sure we don't zap the user-provided ssh_keyfiles.  Otherwise, the code would have been a bit simpler.

infinitewarp (approved):
This looks like a good step forward!

quipucords/scanner/network/utils.py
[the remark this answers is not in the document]
infinitewarp: Acknowledged!

quipucords/scanner/network/utils.py
infinitewarp: I _think_ the recent changes have worked around this risk, if I'm reading the diff correctly. Thanks!

quipucords/api/migrations/0032_add_ssh_keyvalue_to_credential.py
```diff
@@ -0,0 +1,18 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("api", "0031_alter_scantask_systems_count_and_more"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="credential",
+            name="ssh_keyvalue",
```
bruno-fs, line 15: given this migration does not exist on main branch, I'd change "ssh_keyvalue" right here - and purge migration 33 from git history 🔥 🔥 🔥

quipucords/scanner/network/utils.py
```diff
@@ -173,3 +182,60 @@ def get_fact_names():
...
+def generate_ssh_keyfile(credential, ssh_key):
+    """Generate an ssh private keyfile for the ssh key content specified."""
+    ssh_pkey_dir = Path(tempfile.gettempdir()) / GEN_KEYFILES_DIRECTORY
+    ssh_pkey_dir.mkdir(parents=True, exist_ok=True)
+    private_keyfile_path = tempfile.NamedTemporaryFile(
+        prefix=f"{GEN_KEYFILE_PREFIX}_{credential.get('id')}_", dir=ssh_pkey_dir
+    ).name
+    with open(private_keyfile_path, "w+") as private_key_fp:
+        private_key_fp.write(ssh_key)
+        private_key_fp.write("\n")
```
bruno-fs, line 200: w8... so the key has 2 line breaks in the end? (the one preserved on the serializer + this)

abellotti: I had that NL added early on because the SSH key file MUST be NL terminated or it will fail to authenticate, so in the cases where someone adds/edits one without (API or through QPC), it will guarantee it working and it does not hurt to have the extra NL.

quipucords/api/migrations/0032_add_ssh_keyvalue_to_credential.py
abellotti: done.

infinitewarp (approved):

PR #2392, opened 2023-06-12
fix(ocp): properly encode potential OCP error entities

bruno-fs:
One of the premisses of OCP scanner "extra facts" is that they should not break a scan if an API error is caught. When that happens, an OCPError entity would be collected instead of the extra fact. 

This behavior was broken because OCPErrors were not able to be properly encoded as json.

### DIFF
--- a/quipucords/api/inspectresult/model.py
+++ b/quipucords/api/inspectresult/model.py
@@ -10,7 +10,7 @@ from django.utils.translation import gettext as _
 
 from api import messages
 from api.source.model import Source
-from compat.pydantic import BaseModel
+from compat.pydantic import BaseModel, PydanticErrorProxy
 
 
 class JobInspectionResult(models.Model):
@@ -85,7 +85,7 @@ class RawFactEncoder(JSONEncoder):
 
     def default(self, o):
         """Update the default Encoder to handle types beyond just the basic ones."""
-        if isinstance(o, BaseModel):
+        if isinstance(o, (BaseModel, PydanticErrorProxy)):
             return o.dict()
         if isinstance(o, datetime):
             return o.isoformat()
--- a/quipucords/scanner/openshift/inspect.py
+++ b/quipucords/scanner/openshift/inspect.py
@@ -5,6 +5,7 @@ from django.db import transaction
 
 from api.models import RawFact, ScanTask, SystemInspectionResult
 from scanner.exceptions import ScanFailureError
+from scanner.openshift.api import OpenShiftApi
 from scanner.openshift.entities import OCPBaseEntity, OCPCluster, OCPError, OCPNode
 from scanner.openshift.runner import OpenShiftTaskRunner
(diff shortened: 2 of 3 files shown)


quipucords/api/inspectresult/model.py
```diff
@@ -85,7 +85,7 @@ class RawFactEncoder(JSONEncoder):
 
     def default(self, o):
         """Update the default Encoder to handle types beyond just the basic ones."""
-        if isinstance(o, BaseModel):
+        if isinstance(o, (BaseModel, PydanticErrorProxy)):
```
infinitewarp, line 88: What exactly does adding `PydanticErrorProxy` do for us here? It looks like the class definition is empty. So, I assume something somewhere is creating an object and using that type, but I don't know where or how. Where would that manifest?

bruno-fs: Well, that will need some time to explain, hold my beer while I pick the context in other parts of the code

bruno-fs: OCPError is an pydantic model that also behaves as an exception.
That should not be possible due to metaclasses clashes, but I did an overly complicated thing to make it behave as so.

Let's start with the pretty part:

```python
@raises(SomeException)
class SomePydanticModel(BaseModel):
    ...
```
By wrapping "SomePydanticModel" with "raises" I can make it behave as an exception.
This was how OCPError was created
https://github.com/quipucords/quipucords/blob/1172fedaffe128c35dcf87342292a84e4ca9ec8e/quipucords/scanner/openshift/entities.py#L153-L160

Problem is, OCPError does not get matched to be encoded as json when saving it because `raises` does some dark magic involving, as you probably guessed, `PydanticErrorProxy`. And with this, I can introduce the ugly part:
https://github.com/quipucords/quipucords/blob/1172fedaffe128c35dcf87342292a84e4ca9ec8e/quipucords/compat/pydantic.py#L34-L74

bruno-fs: TLDR: `PydanticErrorProxy` is the base class for OCPError WHEN it is raised as an exception.

infinitewarp: Okay, I think I am understanding this more. Thanks for the background!

infinitewarp (commented):
I had to read it a few times to get it, but I think this makes sense to me. Patching the retrieve method _not_ selected by parametrize (line 205) is clever but also confused me a bit at first.

:+1:

infinitewarp (approved):

bruno-fs:
> I had to read it a few times to get it

Mmmmm I'll try to think of a more simple way to do that 🤔

infinitewarp:
> Mmmmm I'll try to think of a more simple way to do that 🤔

I wouldn't sweat it. This is probably just showing my lack of familiarity of this area of code.

PR #2396, opened 2023-06-14
chore: follow psycopg2 mantainers recommendation

bruno-fs:
Follow psycopg2 mantainers recommendation [1] and use `psycopg2` package instead of `psycopg2-binary`.

Not sure if related (didn't bother further investigating) but replacing the binary with the proper source version fixed an issue I was experiencing on my local environment: a very consistent segfault when trying to restart the server after populating the database with some scan data. It's important to note this is likely only affecting upstream versions (since all dependencies are built from source on downstream versions)

[1]: https://www.psycopg.org/docs/install.html#psycopg-vs-psycopg-binary


infinitewarp (approved):
I find it curious that they say the binary package "is meant for beginners" when the real benefit, or so I assumed, was that using the binary version meant you didn't have to install extra libs from other external projects like OpenSSL and PostgreSQL to a system that otherwise has no use for them. Or maybe I misunderstood and my memory is cloudy. 🤷‍♂️ 

Either way, if this works, it works. I probably already have those libs installed on my dev machine anyway. If I don't and things start breaking, we'll discuss then.

PR #2227, opened 2023-01-09
[DISCOVERY-239] - update server-version upstream

bruno-fs:
Fix: DISCOVERY-239

## How to test

`curl --insecure  https://localhost:9443/api/v1/status/ | jq '.server_version'`

the final part of the version should match current commit version instead of "master"

PR #2402, opened 2023-06-30
Build and upload container from GitHub Actions

mirekdlugosz:
Part of DISCOVERY-360.

This will result in 3 types of tags present in quay:

- `main`, which follows tip of `main` branch - image rebuilt on PR merge
- <version tag> - every time you create a new tag. Usually we tag commits in branches, so this PR might need backporting to any release branches we care about
- `pr-<number>` - for every PR (maybe including this one?), set to be automatically removed after 3 days.


.github/workflows/build.yml
```diff
@@ -4,9 +4,13 @@ on:
...
     branches: [ main ]
     tags:
-      - "*"
+      - '*'
   pull_request:
-    branches: [main, "release/*"]
+    branches: [main, 'release/*']
+
+env:
+  STABLE_TAG: ${{ github.event_name == 'push' && github.ref_name || format('pr-{0}', github.event.pull_request.number) }}
```
mirekdlugosz, line 12: "push" event is only when you push (merge) to main branch, or you push (create) tag. When you (force) push to PR branch, it generates "pull_request" event of "synchronize" type.

This effectively means:
- if you push to main branch, then use main tag
- if you create tag, then use this git tag as container image tag
- if you push to PR, then use pr-NUMBER as a tag

.github/workflows/build.yml
```diff
@@ -4,9 +4,13 @@ on:
...
+  EXPIRATION_LABEL: ${{ github.event_name == 'push' && 'quipucords.source=github' || 'quay.expires-after=3d' }}
```
mirekdlugosz, line 13: I had trouble with GitHub setting quay expiration label if I merged to main. No matter if I skipped condition, set variable to empty string or null. So I just put any meanigless value as a workaround.

.github/workflows/build.yml
```diff
@@ -18,20 +22,23 @@ jobs:
...
-      - name: Set dynamic variables
-        # this might seem odd but is the recommended way to set variables
-        # reference: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable
-        run: |
-          pip install poetry
-          BUILD_VERSION=$(poetry version -s)
-          echo "BUILD_VERSION=$BUILD_VERSION+$GITHUB_SHA" >> $GITHUB_ENV
-          echo "IMAGE_NAME_WITH_TAG=quipucords:$BUILD_VERSION" >> $GITHUB_ENV
-
-      - name: Docker Layer Caching
```
mirekdlugosz, line 30: Since we move to podman / buildah, docker layer caching is no longer useful for us. If image building performance becomes a bottleneck, we might need to think about bringing this back.

.github/workflows/build.yml
```diff
@@ -18,20 +22,23 @@ jobs:
...
-        uses: satackey/action-docker-layer-caching@v0.0.11
-        # Ignore the failure of a step and avoid terminating the job.
-        continue-on-error: true
-
       - name: Build quipucords image
-        run: docker build --build-arg BUILD_COMMIT=$GITHUB_SHA . -t $IMAGE_NAME_WITH_TAG
```
mirekdlugosz, line 36: Looks like `BUILD_COMMIT` build arg has been removed in #2227.

.github/workflows/build.yml
```diff
@@ -18,20 +22,23 @@ jobs:
...
           fetch-depth: 0  # fetches all commits/tags
 
-      - name: Set dynamic variables
-        # this might seem odd but is the recommended way to set variables
-        # reference: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable
-        run: |
-          pip install poetry
-          BUILD_VERSION=$(poetry version -s)
-          echo "BUILD_VERSION=$BUILD_VERSION+$GITHUB_SHA" >> $GITHUB_ENV
-          echo "IMAGE_NAME_WITH_TAG=quipucords:$BUILD_VERSION" >> $GITHUB_ENV
```
mirekdlugosz, line 28: Just for the record, there were multiple problems with this:

1. It doesn't work as intended. Idea was to create something like "1.2.3+sha" as git tag. But this new variable was only visible since next step, and instead images were tagged with "1.2.3"
2. Container tags can't use "+" sign
3. Expiration time is assigned to container image, not tag. If we create tags with git hash when building from main, then we either apply expiration to both tags (which is not ideal, because we want main to always resolve), or neither tag (which is not ideal, because now we have "1.2.3+sha" tag that never expires and takes space in quay).

.github/workflows/build.yml
```diff
@@ -18,20 +22,23 @@ jobs:
...
-        uses: satackey/action-docker-layer-caching@v0.0.11
-        # Ignore the failure of a step and avoid terminating the job.
-        continue-on-error: true
-
       - name: Build quipucords image
-        run: docker build --build-arg BUILD_COMMIT=$GITHUB_SHA . -t $IMAGE_NAME_WITH_TAG
+        id: build-image
+        uses: redhat-actions/buildah-build@v2
+        with:
+          image: quipucords/quipucords
```
mirekdlugosz, line 29: It's easy to also build arm image: just install qemu-user-static and add `platforms` here. See https://github.com/mirekdlugosz/quipucords/commit/b2aaf991a8fa5a2a593252fbeaa5a61d992d5673

HOWEVER building amd64 image takes about 2 minutes. Building arm image takes 10 minutes. And we upload only once both finished building.

Jenkins pipeline will need to pull image built here, and if we spend 12 minutes building images, and then 30 minutes running Camayoc, you will get Camayoc results about 45 minutes after opening PR. I set it as a self-imposed goal to have Camayoc results within 30 minutes after opening PR.

There might be a way to build and upload amd64, and then build and upload arm separately. I have not investigated. I suggest we made that a separate task.

.github/workflows/build.yml
bruno-fs: I think it's worth commenting this on the code (I will certainly forget this haha)

bruno-fs (commented):
LGTM!

bruno-fs (approved):
I'll leave approved, but please consider adding a comment here https://github.com/quipucords/quipucords/pull/2402/files#r1250927279

.github/workflows/build.yml
mirekdlugosz: Done in 21f90b7ed31d706668f7727c0830ca9419da2900

PR #2407, opened 2023-07-04
Create script to generate list of commands that require sudo

nicolearagao:
This PR introduces a new script that analyzes Ansible playbooks to collect all commands
requiring sudo permissions for execution. Additionally, it includes the following enhancements:

Integration of a step in GitHub Actions to ensure the generated list is up to date.
Addition of commands to the Makefile for generating the sudo command list and performing list comparison tests.

This PR implements [[DISCOVERY-366]](https://issues.redhat.com/browse/DISCOVERY-366)


abellotti:
The generated file docs/sudo_cmd_list.txt seems to be what is run (e.g. one-liner shell scripts).  Do we need the more refined list for the sudoers file, i.e. just the list of fully qualified commands? i.e. 

```
/usr/bin/find
/usr/bin/cat
/usr/bin/unzip
/sbin/chkconfig
/usr/bin/systemctl
...
```

Not sure if this can be automated or just manually created out of the generated docs/sudo_cmd_list.txt

scripts/generate_sudo_list.py
```diff
@@ -0,0 +1,64 @@
...
+functions = {
+    "docs": create_document_file,
+    "compare": compare_sudo_cmds_lists,
+}
+
+if __name__ == "__main__":
+    selected_function = functions[sys.argv[1]]
+    function_arguments = sys.argv[2:]
+
+    sys.exit(selected_function(*function_arguments))
```
bruno-fs, line 64: [optional] I think we should establish a pattern for "quipucords scripts" - with this being the first one.

Requirements for quipucords scripts:
- should be [django management commands](https://docs.djangoproject.com/en/4.2/howto/custom-management-commands/)
- they should be covered in tests as the rest of the application
- exceptions can be made for some specific purpose commands, like celery entrypoint or a custom script that should run independently from quipucords dependencies.

Thoughts @abellotti & @infinitewarp ?

abellotti: I think for this purpose, a Django management command would be overkill. Not doing DB, etc.  This is more of a static analysis/linting/man page thing, so a Makefile entry point sounds about right.

infinitewarp: I agree with @abellotti. As discussed earlier today at standup, a Django management command might be overkill for this script since it's not using our models or other internal quipucords logic. I would not strongly recommend it in this case.

Mirek also pointed out that making this a Django management command would require the user to have the full quipucords source and config locally installed in order to use this script. Because we are committing the output to the repo and docs anyway, the user should not need to run this script directly, but we should not impose that kind of impediment unless actually necessary.

infinitewarp:
> The generated file docs/sudo_cmd_list.txt seems to be what is run (e.g. one-liner shell scripts). Do we need the more refined list for the sudoers file, i.e. just the list of fully qualified commands these scripts run? i.e.
> 
> ```
> /usr/bin/find
> /usr/bin/cat
> /usr/bin/unzip
> /sbin/chkconfig
> /usr/bin/systemctl
> ...
> ```
> 
> Not sure if this can be automated or just manually created out of the generated docs/sudo_cmd_list.txt

I don't think we want to find the absolute paths for commands in the output _if we are not already using absolute paths_ in our code. Maybe we should change all commands in our code and playbooks to use absolute paths, but I think that's outside the scope of this change. (For the record, I _do_ believe we should make all commands use absolute paths where possible.) We should list the commands exactly as we call them.

I do believe that the intent of the original request was to get only the program names, though, not all of their arguments. The current list makes some commands not obvious. For example:

```
subscription-manager facts --list | grep '^virt.host_type.' | sed -n -e 's/^.*virt.host_type.\s*//p'
```

In this example, does only `subscription-manager` require `sudo`? Or do `grep` and `sed` _also_ require `sudo`?

I would _guess_ that only `subscription-manager` actually requires `sudo`, but I don't know if that's how the current code _actually behaves_. Maybe we are naively expecting `grep` and `sed` to also run through `sudo` if this entire command is sent into a `sudo` subshell. 🤷

infinitewarp:
> I do believe that the intent of the original request was to get only the program names, though, not all of their arguments. The current list makes some commands not obvious. For example:
> 
> ```
> subscription-manager facts --list | grep '^virt.host_type.' | sed -n -e 's/^.*virt.host_type.\s*//p'
> ```
> 
> In this example, does only `subscription-manager` require `sudo`? Or do `grep` and `sed` _also_ require `sudo`?
> 
> I would _guess_ that only `subscription-manager` actually requires `sudo`, but I don't know if that's how the current code _actually behaves_. Maybe we are naively expecting `grep` and `sed` to also run through `sudo` if this entire command is sent into a `sudo` subshell. 🤷

To be clear, I'm stating the **ideal** output here. In practice, I don't know if this is a reasonable request because I don't know of a clean and straightforward way to _parse_ shell commands in Python to extract only `subscription-manager`, `grep`, and `sed` from that example. We could use regular expressions, but the pattern will get very complicated very quickly if we want to support all of `bash`'s syntax (assuming a `bash`-like shell).

We might need to make a judgment call here and decide how "clean" and concise we really want the output to be. The current output might have to be good enough, and we may leave the parsing as an exercise for the end user.

bruno-fs:
> In this example, does only subscription-manager require sudo? Or do grep and sed also require sudo?

considering that sudo is required for the whole "pipeline", I'd say all commands require sudo (the list is already filtering for commands that are requiring sudo

TBD: will add links

scripts/generate_sudo_list.py
```diff
@@ -0,0 +1,64 @@
...
+import yaml
+
+
+def collect_sudo_commands():
+    """Collect all ansible cmds that require sudo."""
+    network_runner_path = Path("quipucords/scanner/network/runner")
+    list_cmds = []
+    for playbook in network_runner_path.glob("roles/*/tasks/main.yml"):
+        for task in yaml.safe_load(playbook.open()):
+            if "user_has_sudo" in task.get("when", ""):
```
bruno-fs, line 16: we are only collecting raw commands that require sudo.

I'd check for this in a slightly different way.
Instead of checking for our "user_has_sudo" I'd check for "become" == yes/true as this is the actual way to have a role run with a privilege escalation.

infinitewarp (approved):
Assuming we're okay with the output file containing the full commands, which I think we agreed last week is okay since parsing out the individual program names is much more difficult, I approve! 👍

Makefile
```diff
@@ -186,3 +186,9 @@ build-container:
 
 check-db-migrations-needed:
 	$(PYTHON) quipucords/manage.py makemigrations --check
+
+generate-sudo-list:
+	@$(PYTHON) scripts/generate_sudo_list.py docs "docs/sudo_cmd_list.txt"
+
+test-sudo-list:
```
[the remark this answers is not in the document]
nicolearagao: @ruda It's possible, but I am not sure if this is the best approach here. This test should serve more like a warning like "Hey, something was updated in the code, the current report you have is outdated, generate the report again." But I am not sure. what do u think @infinitewarp @bruno-fs and @abellotti ?

scripts/generate_sudo_list.py
```diff
@@ -0,0 +1,64 @@
...
+
+def compare_sudo_cmds_lists(file_path):
+    """Compare the most updated sudo list with an existing one."""
+    generated_doc = generate_sudo_cmds_document()
+
+    with tempfile.NamedTemporaryFile(delete=False) as temp_file:
+        temp_file.write(generated_doc.encode())
+        try:
+            if filecmp.cmp(temp_file.name, file_path):
+                sys.stdout.write("The sudo command list document is up to date.")
```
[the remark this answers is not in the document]
nicolearagao: I will fix that!

docs/sudoers_required_cmds.txt
```diff
@@ -0,0 +1,30 @@
...
+/usr/bin/systemctl
+/sbin/chkconfig
+stat
+df
+ps
+tr
+readlink
+subscription-manager
+awk
+virt-what
```
infinitewarp, line 30: I know this file was manually created, but you may want to pass it through `| sort | uniq` before committing it to improve readability. Here's what I got:

```suggestion
/sbin/chkconfig
/sbin/ifconfig
/usr/bin/systemctl
/usr/sbin/dmidecode
awk
cat
command
df
dirname
echo
egrep
fgrep
find
grep
java
locate
ls
ps
readlink
sed
sort
stat
subscription-manager
tail
test
tr
unzip
virt-what
xargs
yum
```

docs/sudo_cmd_list.txt
```diff
@@ -0,0 +1,80 @@
...
+subscription-manager facts --list | grep '^cpu.cpu_socket(s).' | sed -n -e 's/^.*cpu.cpu_socket(s).\s*//p'
+subscription-manager facts --list | grep '^virt.host_type.' | sed -n -e 's/^.*virt.host_type.\s*//p'
+subscription-manager facts --list | grep '^virt.is_guest.' | sed -n -e 's/^.*virt.is_guest.\s*//p'
+subscription-manager facts --list 2>/dev/null | grep '^virt.uuid.' | sed -n -e 's/^.*virt.uuid.\s*//p'
+LANG=C subscription-manager status | grep -e 'Overall Status:'| cut -d ":" -f2
+subscription-manager list --consumed 2>/dev/null | grep -e '^SKU' -e '^Subscription Name' | sed -n -e 's/SKU\s*.\s*//p' -e 's/Subscription Name\s*.\s*//p' | awk '{ ORS = (NR%2 ? " - " {{":"}} RS) } 1'
+subscription-manager identity 2>/dev/null | 
+grep 'system identity:' |
+sed 's/system identity:\s*//'
+
```
infinitewarp, line 76: I just noticed this blank line. I wonder if there's a bug in `collect_sudo_commands` that results in something adding an empty string to `list_cmds`. I have an idea that I'll comment in that function…

Makefile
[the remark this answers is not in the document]
ruda: No worries, if this becomes important, then you can improve it in the future.

scripts/generate_sudo_list.py
```diff
@@ -0,0 +1,69 @@
...
+def collect_sudo_commands(path=None):
+    """Collect all ansible cmds that require sudo."""
+    if not path:
+        path = Path("quipucords/scanner/network/runner")
+    list_cmds = []
+    for playbook in path.glob("roles/*/tasks/main.yml"):
+        for task in yaml.safe_load(playbook.open()):
+            if task.get("become", ""):
+                list_cmds.append(task.get("raw", ""))
+    return list_cmds
```
infinitewarp, line 20: Before we return the list, we may want to drop any empty strings. I wonder if there's a task somewhere that put an empty string resulting in that stray blank line in the output. 🤔 

Consider:
```
    return [cmd.strip() for cmd in list_cmds if isinstance(cmd, str) and cmd.strip()]
```

That list comprehension will strip whitespace from each item and return only non-empty items. For example:

```
>>> list_cmds = ['hello', '', '1', 2, 'world', 'world', ' ', '0', 'False', None, False, True]
>>> [cmd.strip() for cmd in list_cmds if isinstance(cmd, str) and cmd.strip()]
['hello', '1', 'world', 'world', '0', 'False']
```

We may _also_ want to sort the final list, though, because `path.glob` **does not guarantee order**. Running the same command at different times or on different machines can result in a different ordered list, even if the actual content would be the same. So, consider this suggestion to unique (via `set`) and sort the list:

```suggestion
    list_cmds = [cmd.strip() for cmd in list_cmds if isinstance(cmd, str) and cmd.strip()]
    list_cmds = sorted(set(list_cmds))
    return list_cmds
```

That _could_ be a one-line return statement, but breaking it into a few lines might help with future debugging if we encounter issues. That's just a subjective opinion, though. :shrug: Anyhow, using the previous example with this new code, it would return:

```
['0', '1', 'False', 'hello', 'world']
```

No empty values, no duplicates, and order is sorted. 🎉

nicolearagao: good idea! thanks

infinitewarp (approved):

ruda (approved):

scripts/generate_sudo_list.py
```diff
@@ -0,0 +1,74 @@
...
+
+def compare_sudo_cmds_lists(file_path):
+    """Compare the most updated sudo list with an existing one."""
+    generated_doc = generate_sudo_cmds_document()
+
+    with tempfile.NamedTemporaryFile(delete=False) as temp_file:
+        temp_file.write(generated_doc.encode())
+        try:
+            if filecmp.cmp(temp_file.name, file_path):
+                sys.stdout.write("The sudo command list document is up to date.\n")
```
bruno-fs, line 51: why write to stdout instead of stderr?

scripts/test_generate_sudo_list.py
```diff
@@ -0,0 +1,180 @@
+"""Tests for generate sudo list script."""
```
bruno-fs, line 1: mmm didn't we had an agreement on having all tests located in a single directory? cc @infinitewarp

scripts/generate_sudo_list.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def collect_sudo_commands(path=None):
+    """Collect all ansible cmds that require sudo."""
+    if not path:
+        path = Path("quipucords/scanner/network/runner")
+    list_cmds = []
+    for playbook in path.glob("roles/*/tasks/main.yml"):
+        for task in yaml.safe_load(playbook.open()):
+            if task.get("become", ""):
```
bruno-fs, line 18: just FYI, technically become could be "falsy", but since we don't do that I think it's OK to leave it as is.

scripts/test_generate_sudo_list.py
infinitewarp: True… good catch!

scripts/generate_sudo_list.py
nicolearagao: Since `sys.stdout `is used for normal program output, while `sys.stderr `is used for error messages, I thought would be nice to differentiate the messages. Was that a bad choice ?

scripts/test_generate_sudo_list.py
bruno-fs: as we agreed in a slack conversation... let's keep this as is for now

scripts/generate_sudo_list.py
bruno-fs: unix way is to output any type of message to stderr
you only output to stdout if this the "output" of the program. this is a log for a script without an output

we could also just print nothing if nothing is wrong (I tend to prefer that XD, but that's just me)

bruno-fs: if it were to write anything to stdout, I'd say write the contents it would write to a file

not saying to do that, tho. just something that would make sense on stdout.

nicolearagao: I will change to print nothing if everything is ok.

scripts/generate_sudo_list.py
```diff
@@ -48,7 +48,7 @@ def compare_sudo_cmds_lists(file_path):
         temp_file.write(generated_doc.encode())
         try:
             if filecmp.cmp(temp_file.name, file_path):
-                sys.stdout.write("The sudo command list document is up to date.\n")
+                pass
             else:
```
bruno-fs: replace `if <somethings>/pass/else` pattern with `if not <something>`

nicolearagao: done and rebased

bruno-fs (approved):

PR #2409, opened 2023-07-06
compatibility with rhel 5/6

bruno-fs:
UBI 9 has stricter criptographic policies for ssh. We need to explicitly set those to "legacy mode" for detection of rhel 5 and 6.

### DIFF
--- a/quipucords/scanner/network/utils.py
+++ b/quipucords/scanner/network/utils.py
@@ -61,7 +61,19 @@ def _construct_vars(connection_port, credential=None):
     :returns: a dict that can be used as the host variables in an
         Ansible inventory.
     """
-    ansible_vars = {"ansible_port": connection_port}
+    # ssh arguments for compatibility with defaults for older systems (like RHEL 5&6)
+    # (this is only meant to help quipucords running on baremetal since the container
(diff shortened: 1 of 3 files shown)


quipucords/scanner/network/utils.py
```diff
@@ -61,7 +61,19 @@ def _construct_vars(connection_port, credential=None):
...
+    legacy_ssh_args = (
+        "-o 'KexAlgorithms=diffie-hellman-group-exchange-sha1,diffie-hellman-group14-sha1,diffie-hellman-group1-sha1'"  # noqa: E501
+        " -o 'HostKeyAlgorithms=+ssh-rsa,ssh-dss'"
+        " -o 'PubkeyAcceptedKeyTypes=+ssh-rsa,ssh-dss'"
+    )
+
+    ansible_vars = {
+        "ansible_port": connection_port,
+        "ansible_ssh_common_args": legacy_ssh_args,
+    }
```
bruno-fs, line 76: I'm on the fence whether we should add this to the final product.

I don't like explicitly setting less secure configurations. OTOH, discovery won't run as a openssh server and this would make bare-metal setups slightly easier.

PR #2413, opened 2023-07-07
perf: remove all dunder str methods from models

bruno-fs:
some `__str__` methods are making calls to related models, which can quickly cascade into a nasty N+1 problem.

We don't need all this info in `__str__` anyways, just remove it and log explicitly when required.

relates to JIRA: DISCOVERY-220

### DIFF
--- a/quipucords/api/details_report/model.py
+++ b/quipucords/api/details_report/model.py
@@ -22,13 +22,3 @@ class DetailsReport(models.Model):
     )
     cached_csv = models.TextField(null=True)
     cached_masked_csv = models.TextField(null=True)
-
-    def __str__(self):
-        """Convert to string."""
-        return (
-            "{"
-            f'"id":{self.id},'
-            ' "report_type":"details", '
-            f' "sources":{self.sources}'
-            "}"
-        )
--- a/quipucords/api/connresult/model.py
+++ b/quipucords/api/connresult/model.py
@@ -13,10 +13,6 @@ from api.source.model import Source
 class JobConnectionResult(models.Model):
     """The results of a connection scan."""
 
-    def __str__(self):
-        """Convert to string."""
-        return f"{{ id:{self.id},task_results:{self.task_results} }}"
-
     class Meta:
         """Metadata for model."""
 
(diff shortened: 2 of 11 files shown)


infinitewarp (changes requested):
In isolation, this seems fine, but https://issues.redhat.com/browse/DISCOVERY-220 mentions

> If specific information is needed for logging, it should be queried and logged explicitly.

If there are logs throughout quipucords that were relying on the old `__str__` methods to emit useful information, those logs would be effectively broken since they may no longer include enough useful information, and that could make diagnosing and helping to resolve customer issues (or even our own issues between environments or during development) much more difficult.

Before we merge this, we should check and evaluate the various logs that relied on the verbose `__str__` calls and, where appropriate, update them to log explicitly whatever relevant information we may need to understand the state of the objects at runtime.

quipucords/api/details_report/model.py
```diff
@@ -22,13 +22,3 @@ class DetailsReport(models.Model):
...
     cached_csv = models.TextField(null=True)
     cached_masked_csv = models.TextField(null=True)
-
-    def __str__(self):
-        """Convert to string."""
-        return (
-            "{"
-            f'"id":{self.id},'
-            ' "report_type":"details", '
-            f' "sources":{self.sources}'
```
bruno-fs, line 32: this is the change that will cause most impact in my evaluation.
I honestly don't think the info is worth for debugging - the tracebacks should be enough IMO


https://github.com/quipucords/quipucords/blob/9de10b9f7613c4a82a203e690b87c4e7eb0d6db7/quipucords/scanner/job.py#L358-L362
https://github.com/quipucords/quipucords/blob/9de10b9f7613c4a82a203e690b87c4e7eb0d6db7/quipucords/scanner/job.py#L374-L379

bruno-fs:
affected logs

https://github.com/quipucords/quipucords/blob/1508c3c751f1af1f6aff7fafd1349b7775022c52/quipucords/scanner/runner.py#L103-L107
https://github.com/quipucords/quipucords/blob/9de10b9f7613c4a82a203e690b87c4e7eb0d6db7/quipucords/scanner/job.py#L358-L362
https://github.com/quipucords/quipucords/blob/9de10b9f7613c4a82a203e690b87c4e7eb0d6db7/quipucords/scanner/job.py#L374-L379
https://github.com/quipucords/quipucords/blob/9de10b9f7613c4a82a203e690b87c4e7eb0d6db7/quipucords/scanner/job.py#L112-L122
https://github.com/quipucords/quipucords/blob/9de10b9f7613c4a82a203e690b87c4e7eb0d6db7/quipucords/scanner/tasks.py#L101

PR #2418, opened 2023-07-18
fix: allow clients to send empty strings for optional credential fields

infinitewarp:
This is required because some clients unexpectedly send a blank string instead of ommiting the field when sending data to our API. This may happen if a user enters and then removes optional values before saving or when changing an existing credential's auth method (e.g. from using a username+password to using an SSH key).

Changing the base model might seem like overkill here, but since we already use custom validation logic and expicitly reset some of these these field values to `None` wheniever they are not actually relevant, I think this should be a reasonable compromise.

Relates to JIRA: DISCOVERY-359
https://issues.redhat.com/browse/DISCOVERY-377


quipucords/tests/api/credential/test_serializer.py
```diff
@@ -125,7 +125,11 @@ def test_openshift_cred_empty_auth_token():
     }
     serializer = CredentialSerializer(data=data)
     assert not serializer.is_valid(), serializer.errors
-    assert serializer.errors["auth_token"]
+    assert serializer.errors["non_field_errors"] == [messages.TOKEN_OR_USER_PASS]
+    # The above assertion is technically correct but maybe misleading. We allow the
+    # `auth_token` field to be blank; so, technically there is no error on that field.
+    # Instead, we rely on the serializer's `validate` method to identify the problem
+    # and raise the non-field-error TOKEN_OR_USER_PASS.
```
bruno-fs, line 132: 👍🏽

bruno-fs (approved):
LGTM

PR #2435, opened 2023-08-02
feature(ocp): gather acm metrics

nicolearagao:
Collect metrics on ACM managed clusters.

This feature implements [DISCOVERY-349](https://issues.redhat.com/browse/DISCOVERY-349)


quipucords/scanner/openshift/api.py
```diff
@@ -291,3 +309,33 @@ def _init_cluster(self, cluster) -> OCPCluster:
...
+        is_cluster_available = any(
+            condition.get("reason") == "ManagedClusterAvailable"
+            for condition in conditions
+        )
+        managed_cluster_metrics["available"] = is_cluster_available
+
+        capacity = cluster["status"]["capacity"]
+        if capacity:
+            managed_cluster_metrics["core_worker"] = capacity.get("core_worker", "")
+            managed_cluster_metrics["socket_worker"] = capacity.get("socket_worker", "")
```
bruno-fs, line 337: could core_worker/socket_worker not be available when capacity is present? and do you know why capacity might not be present?

quipucords/tests/scanner/openshift/test_api.py
```diff
@@ -178,6 +186,18 @@ def test_retrieve_operators(ocp_client: OpenShiftApi):
...
 
 
+@pytest.mark.vcr(
+    VCRCassettes.OCP_ACM,
+    VCRCassettes.OCP_DISCOVERER_CACHE,
+)
+def test_retrieve_acm_metrics(ocp_client: OpenShiftApi):
+    """Test retrieve acm metrics method."""
+    acm_metrics = ocp_client.retrieve_acm_metrics()
+    assert isinstance(acm_metrics, list)
```
bruno-fs, line 196: the comment below refers to this line. we are usually writing comments before the relevant code, not after

bruno-fs (commented):
I have some comments and questions, but looks pretty good =D

bruno-fs:
please change the body of the commit message.

Instead of "This feature implements DISCOVERY-349", use "Relates to JIRA: DISCOVERY-349"

quipucords/scanner/openshift/api.py
nicolearagao: Capacity might not be present because we are looking at things from the hub perspective. So this endpoint will retrieve all the managed clusters, regardless their state: clusters that were connected in the past but now are disconnected, active clusters, and clusters that you wanted to import but had any sort of failure.If you were not able to fully connect to a cluster, then you couldn't collect detailed info about it, this is when conditions is not available. 
I haven't seen any case when capacity is present but not info on socket_workers and core_workers.

Looking at the code from the repo that generates those metrics. If capacity is present, then it will return some sort of value for core_worker and socket_worker.

https://github.com/stolostron/clusterlifecycle-state-metrics/blob/495ed8be90cf4a27ae9fac0c2518320462c1e46a/pkg/generators/cluster/managedclusterinfo.go#L160

quipucords/scanner/openshift/api.py
```diff
@@ -291,3 +309,33 @@ def _init_cluster(self, cluster) -> OCPCluster:
...
+        conditions = cluster["status"]["conditions"]
+        is_cluster_available = any(
+            condition.get("reason") == "ManagedClusterAvailable"
+            for condition in conditions
+        )
+        managed_cluster_metrics["available"] = is_cluster_available
+
+        capacity = cluster["status"]["capacity"]
+        if capacity:
+            managed_cluster_metrics["core_worker"] = capacity.get("core_worker", "")
```
infinitewarp, line 336: Why do we default to empty string `""` instead of `None`?

Asking because I'm not familiar with the logic of how we normally process these metrics.

quipucords/scanner/openshift/api.py
bruno-fs, line 337: None of the other facts collected on OCP scans can be absent and I don't see the need for an exception here.

```suggestion
        capacity = cluster["status"]["capacity"] or {}
        managed_cluster_metrics["core_worker"] = capacity.get("core_worker")
        managed_cluster_metrics["socket_worker"] = capacity.get("socket_worker")
```

quipucords/scanner/openshift/inspect.py
```diff
@@ -67,16 +67,24 @@ def _extra_cluster_facts(
...
-                extra_facts[fact_name] = api_method(
+                extra_facts_data = api_method(
                     timeout_seconds=settings.QPC_INSPECT_TASK_TIMEOUT
                 )
+                if fact_name == "acm_metrics":
+                    for managed_cluster in extra_facts_data:
+                        managed_cluster["hub_cluster_id"] = cluster.uuid
+                # the way to retrieve a hub cluster id is through clusterClaims,
+                # however if the managed cluster is the hub itself,
+                # this info won't be there.
```
bruno-fs, line 86: just noticed now that this comment is also after the relevant code. please move it.

quipucords/scanner/openshift/inspect.py
```diff
@@ -67,16 +67,24 @@ def _extra_cluster_facts(
...
             try:
                 if fact_name == "workloads" and not collect_ocp_workloads_enabled:
                     continue
-                extra_facts[fact_name] = api_method(
+                extra_facts_data = api_method(
                     timeout_seconds=settings.QPC_INSPECT_TASK_TIMEOUT
                 )
+                if fact_name == "acm_metrics":
+                    for managed_cluster in extra_facts_data:
+                        managed_cluster["hub_cluster_id"] = cluster.uuid
```
bruno-fs, line 83: what do we lose if we dont add this fact? to me this seems to be redundant (and remember this is raw facts)

quipucords/scanner/openshift/api.py
nicolearagao: You have a point.  I will add this modification.

quipucords/tests/scanner/openshift/test_api.py
```diff
@@ -178,6 +186,18 @@ def test_retrieve_operators(ocp_client: OpenShiftApi):
...
+@pytest.mark.vcr(
+    VCRCassettes.OCP_ACM,
+    VCRCassettes.OCP_DISCOVERER_CACHE,
+)
+def test_retrieve_acm_metrics(ocp_client: OpenShiftApi):
+    """Test retrieve acm metrics method."""
+    # OCP instance used on VCR preparation should have ACM operator installed,
+    # with active multiClusterHub instance
+    acm_metrics = ocp_client.retrieve_acm_metrics()
+    assert isinstance(acm_metrics, list)
```
bruno-fs, line 198: I just realized now this function will need more proper testing since we are not using a dedicated entity for it.

quipucords/scanner/openshift/api.py
```diff
@@ -291,3 +309,32 @@ def _init_cluster(self, cluster) -> OCPCluster:
...
+        )
+        managed_cluster_metrics["available"] = is_cluster_available
+
+        capacity = cluster["status"]["capacity"] or {}
+        managed_cluster_metrics["core_worker"] = capacity.get("core_worker")
+        managed_cluster_metrics["socket_worker"] = capacity.get("socket_worker")
+        managed_cluster_metrics["created_via"] = cluster["metadata"]["annotations"][
+            "open-cluster-management/created-via"
+        ]
+        return managed_cluster_metrics
```
bruno-fs, line 340: this function is probably covered in test but only with a integration test.
I think we 1) need either some tests test around it (each branch here spawn one test)

The other OCP related functionalities were kinda fine without because 1) they didn't had branches and 2) the entities initialization was ensuring the data was loaded and coerced.

bruno-fs:
@nicolearagao can you add to the commit message what was the command used to update the cassettes?

quipucords/tests/scanner/openshift/test_api.py
bruno-fs, line 198: Since we are only starting a simple list, we need more assertions on this test.

moreover, since the implementation has branches, I would not rely solely on VCR: each branch will require its own test (and to do that you will probably need a mock output).

quipucords/tests/scanner/openshift/test_api.py
```diff
@@ -178,6 +186,137 @@ def test_retrieve_operators(ocp_client: OpenShiftApi):
...
+
+    # The following assertion is based on the assumption that the provisioned cluster
+    # has not undergone any additional config modifications. When creating the
+    # multiClusterHub instance, it is expected that at least one managed cluster will
+    # be discovered, the hub itself. An associated set of keys is expected.
+    # The "available" key within these metrics is expected to be True, indicating that
+    # the cluster is available and configured properly.
```
bruno-fs, line 205: based on the comment, we can add the following assertion

```python
assert len(acm_metrics) > 2, "This test requires at least 2 clusters managed through ACM"
```

nicolearagao: assert len(acm_metrics) >= 1, "This test requires at least 1 cluster managed through ACM"
If we don't add any clusters, then the only managed cluster found will be the hub itself. This would be enough for this test to work.

nicolearagao: it won't be hub + local-cluster(name for hub cluster when found on managed clusters)

quipucords/tests/scanner/openshift/test_api.py
```diff
@@ -178,6 +186,137 @@ def test_retrieve_operators(ocp_client: OpenShiftApi):
...
+    ]
+    expected_available_value = True
+
+    found_matching_dict = False
+    for managed_cluster in acm_metrics:
+        if all(key in managed_cluster for key in expected_keys):
+            if managed_cluster["available"] == expected_available_value:
+                found_matching_dict = True
+                break
+    assert found_matching_dict
```
bruno-fs, line 225: Question about the data: why only 1 out of 4 clusters is available on the setup?

Regardless, here's a cleaner way to check the data and test ALL the output (not only the "available" cluster should have the expected facts)
```suggestion
    expected_data = [{key: mock.ANY for key in expected_keys}] * len(acm_metrics)
    assert acm_metrics == expected_data
    # at least one cluster is available
    assert any(cluster_metrics["available"] for cluster_metrics in acm_metrics)
```

nicolearagao: When I wrote this I was thinking about not expecting extra clusters in case we would have to rewrite the vcr cassettes. But what you proposed above works really well to check the expected facts regardless of the numbers of clusters present. we should guarantee that all the expected_facts are there for every single cluster + at least one cluster which available key is True.

bruno-fs (approved):
lgtm

PR #2436, opened 2023-08-03
chore: update poetry lock and requirements files, update github workflows

infinitewarp:
I'm not sure how we missed updating Poetry's specified Python version from `3.9` to `3.11` before.

If you aren't already using 3.11 locally, consider using `pyenv` to install it, and then recreate your poetry environment like:

```
poetry env remove --all
poetry env use $(pyenv which python)
poetry install
```


.github/workflows/test.yml
```diff
@@ -56,10 +63,10 @@ jobs:
       - name: Install poetry
         run: pipx install poetry
 
-      - name: Set up Python 3.11
-        uses: actions/setup-python@v4.5.0
+      - name: Set up Python
+        uses: actions/setup-python@v4.7.0
         with:
-          python-version: "3.11"
+          python-version-file: pyproject.toml
```
bruno-fs, line 69: Do you know if github actions parse python version from pyproject.toml? wow

bruno-fs: mmm but ou pyproject.toml still starts at 3.9
https://github.com/quipucords/quipucords/blob/main/pyproject.toml#L14

bruno-fs (approved):

PR #2438, opened 2023-08-04
Adding support for querying OpenShift Prometheus metrics

abellotti:
feat: Adding support for querying OpenShift Prometheus metrics

- Leveraging the OpenShiftApi class so we don't re-invent initialization/auth/etc.
- We'd be using the same OCP Cluster Credentials as quipucords, no additional configuration is needed.
- Failure to access the Prometheus metrics route/host/query will not fail quipucords connects/inspects, it will simply return empty queries.
- Capability exposed via a new metric_query method.

_Note_:  exposing this core capability as a separate PR from the introduction of node hyperthreading fact/fingerprints.

Sample usage:

```
from scanner.openshift.api import OpenShiftApi

conn_info = { 
    "host": "api.samplecluster1.lab.samplecorp.com",
    "port": "6443",
    "protocol": "https",
    "ssl_verify": False,
    "auth_token": "sha256~...................",
}

ocp_api = OpenShiftApi.with_config_info(**conn_info)

query = ''' 
count by(instance, label_node_hyperthread_enabled, label_node_role_kubernetes_io_master)
(max
    by(node, instance, label_node_hyperthread_enabled, label_node_role_kubernetes_io_master)
    (cluster:cpu_core_node_labels)
)
'''
all_node_metrics = ocp_api.metrics_query(query)

for node in ocp_api._list_nodes().items:
  node_metrics = next((metric for metric in all_node_metrics if metric["instance"] == node.metadata.name), {}) 
  print(f"Node: {node.metadata.name}  Metrics: {node_metrics}")
```

Todos:
 - [x] add tests


bruno-fs:
I know this is still a draft but I'd like to share an early feedback: I'd avoid conciliating what's collected on prometheus API with ocp node API. IMO inspection phase should only care about gathering facts. If an extra node is found (or is missing), that should not interfere. Conciliating data should be something for the upcoming "normalization phase" XD

abellotti:
No problem @bruno-fs, it's a good thing I separated the PRs :) in the second part, I was adding a `hyperthread_enabled` fact in the OCPNode. I'll rework that.

quipucords/scanner/openshift/api.py
```diff
@@ -138,6 +139,54 @@ def can_connect(self, raise_exception=False, **kwargs):
...
+                url=f"https://{self._metrics_host}/api/v1/query?query={query}",
+                headers=kube_config.api_key,
+            )
+        except ApiException as err:
+            logger.error(
+                "Failed to execute the Openshift metrics query: '%s' error: '%s'",
+                query,
+                err.reason,
+            )
+            return []
```
bruno-fs, line 162: consider wrapping this function with `catch_k8s_exception` instead as it is our "standard way" to deal with exceptions on ocp API calls.

We could improve how errors are logged there.

bruno-fs (commented):
Pretty good! Just left one optional suggestion

quipucords/scanner/openshift/api.py
abellotti: I was using catch_k8s_exception, but other than the logging, I did not want the error to raise an exception but rather just log the error/query, and return [] to be minimally intrusive.

quipucords/scanner/openshift/api.py
```diff
@@ -138,6 +139,40 @@ def can_connect(self, raise_exception=False, **kwargs):
...
+    @catch_k8s_exception
+    def metrics_query(self, query):
+        """Execute an OpenShift Prometheus Query."""
+        kube_config = self._configuration
+        metrics_host = self._metrics_host()
+        if not metrics_host:
+            return []
+        response = self._api_client.request(
+            "GET",
+            url=f"https://{metrics_host}/api/v1/query?query={query}",
```
infinitewarp, line 154: Will the `query` value always be _extremely_ simple, or should we URL encode it?

My hunch is that we should either URL encode it, _or_ better yet, use the optional `query_params` argument for `request`:

```
>>> from kubernetes.client import ApiClient
>>> help(ApiClient.request)
Help on function request in module kubernetes.client.api_client:

request(self, method, url, query_params=None, headers=None, post_params=None, body=None, _preload_content=True, _request_timeout=None)
    Makes the HTTP request using RESTClient.
```

If you want to check how that argument works, I think this class is defined here: https://github.com/kubernetes-client/python/blob/76ec6f5ce5572f66d82717e296480dc42584f0b7/kubernetes/client/api_client.py#L368

And that look like a thin wrapper for this: https://github.com/kubernetes-client/python/blob/76ec6f5ce5572f66d82717e296480dc42584f0b7/kubernetes/client/rest.py#L113

quipucords/tests/scanner/openshift/test_api.py
```diff
@@ -440,3 +456,35 @@ def test_retrieve_workloads(ocp_client: OpenShiftApi):
...
+@pytest.fixture
+def node_metrics_query():
+    """Return a simple node metrics query."""
+    return "count by(instance) (max by(node, instance) (cluster:cpu_core_node_labels))"
+
+
+@pytest.mark.vcr_primer(VCRCassettes.OCP_METRICS_CACHE)
+@pytest.mark.vcr(
+    VCRCassettes.OCP_ROUTE,
+    VCRCassettes.OCP_METRICS_CACHE,
```
bruno-fs, line 470: please add VCRCassettes.OCP_DISCOVERER_CACHE and recreate the cassette

abellotti: worked like a champ, Thanks. https://github.com/quipucords/quipucords/pull/2438/commits/931cc883819bb47e4f2c50a279c40b04e1c01156.

bruno-fs (approved):
LGTM

PR #2439, opened 2023-08-04
chore: Pinning debugpy to version 1.6.7

abellotti:
chore: Pinning debugpy to version 1.6.7

- Apparently 1.6.8 wasn't a fully-blown/published release, it was just a link to the 1.6.7 commit.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,7 +53,6 @@ requests-mock = "^1.10.0"
 httpretty = "^1.1.4"
 black = "^22.12.0"
 beautifulsoup4 = "^4.11.1"
-debugpy = "^1.6.5"
 pip-tools = "^6.12.1"
 pytest-docker-tools = "^3.1.3"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -37,6 +37,7 @@ more-itertools = "^9.1.0"
 packaging = "^23.1"
 setuptools = "^68.0.0"
 gitpython = "^3.1.32"
+debugpy = "1.6.7"
```
infinitewarp, line 40: I think this should go under `dev.dependencies` but it probably doesn't matter much since we all just install everything. I don't really use `debugpy` in my personal environment; so, I don't know. Thoughs  @bruno-fs ?

bruno-fs: this should go under dev dependencies, unless running with debugger becomes an "user facing" feature

infinitewarp (approved):

bruno-fs (approved):

PR #2442, opened 2023-08-14
RFC: Normalization phase

bruno-fs:
This is a RFC for the normalization phase.

DO NOT MERGE any of this before the following is solved:
- [ ] `DetailsReport` and `DeploymentsReport` models don't need to exist (as models) 
- [x] ubiquitous `report_id` field used in many models is just an integer, not an actual FK.
- [ ] Remove the `HostEntity` proxy and make `YupanaSerializer` use `Host` instead.

Open questions
- Should we support a data normalizing raw facts from old reports?
- Will we deduplicate data?
  - from the most recent discussions, the answer seems to be yes, but we need a FIRM yes or no to influence this implementation
- Should `Host` replace `SystemFingerprint` model? (this would help with the issue of data duplication)


quipucords/scanner/network/normalizer.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def infrastructure_type_normalizer(*, virt_what_type, virt_type):
+    """Normalize infrastructure type."""
+    # https://github.com/quipucords/quipucords/blob/e65d0e3deb23adca41ceed701e4d002187e593f6/quipucords/api/common/entities.py#L92-L112
+    if virt_what_type == "bare metal":
+        return "physical"
+    elif virt_type or virt_what_type:
+        return "virtualized"
+    return "unknown"
```
bruno-fs, line 17: I think this implementation is wrong (and will need to be updated after current @infinitewarp work on this fact), so don't need to consider the logic here. The important thing is the "demo" for a normalizer with multiple raw facts dependencies.

quipucords/scanner/normalizer.py
```diff
@@ -0,0 +1,169 @@
...
+                if not validator(normalized_fact):
+                    logger.error(
+                        "'{fact_name}' is invalid (value={normalized_fact})",
+                        fact_name=fact_name,
+                        normalized_fact=normalized_fact,
+                    )
+                    return None, None
+        except Exception:
+            logger.exception("Unexpected error during '%s' validation.", fact_name)
+            return None, None
```
bruno-fs, line 110: Catching these exceptions so one borked fact/normalizer/validator won't cause issues for other facts.

I plan on cleaning up this implementation replacing `return None, None` on exceptions with a specialized exception to be captured later. With this, we could add a "has_error" metadata for fingerprinting/dedup decision.

quipucords/scanner/normalizer.py
```diff
@@ -0,0 +1,169 @@
...
+        # for now, assume all facts contributed
+        return normalized_fact, field.raw_facts
```
bruno-fs, line 112: TODO: implement the API for specifying which raw facts/dependencies) were used for a fact

quipucords/scanner/normalizer.py
```diff
@@ -112,23 +120,31 @@ class FactMapper:
...
         validators=None,
         dependencies: list[str] = None,
     ) -> None:
         """Instantiate FactMapper.
 
-        :param raw_fact_key: name of the raw fact collected during inspection phase
-        :param normalizer: function used to normalize raw fact (signature MUST be
-            fn(raw_fact_value, **dependencies) where dependencies are kwargs from
-            dependencies)
+        :param raw_fact: raw fact name or list of raw fact names used to normalize a
```
infinitewarp, line 132: I prefer the old name or something like it. Just calling this "raw fact" implies that it is the fact itself, not the name of the fact, and that's an anti-pattern that has caused me confusion elsewhere.

bruno-fs: OK! I shall keep it as `raw_fact_key`

infinitewarp: Also, mixing types here feels weird to me. If we expect this to sometimes handle a list of strings, I'd make it always expect a list of strings, even if it's often a list of length one. We don't have to change this; just noting an odd feeling.

infinitewarp: …aaaand after reading a later commit that adds lots of FactMapper instantiations, I see why you made this type flexible since lists are in the minority. I guess I wouldn't want to make them all lists for no other reason.

Makefile
```diff
@@ -195,3 +195,18 @@ generate-sudo-list:
 
 test-sudo-list:
 	@$(PYTHON) scripts/generate_sudo_list.py compare "docs/sudo_cmd_list.txt" || exit 1
+
+# shamelessly copied/modified from https://github.com/RedHatInsights/insights-host-inventory/blob/64fa5698c88cfb92258c532cfbd26a657faa07f6/Makefile#L48-L57
+# hence this role follows GPLv2 license (I guess... HELP)
```
infinitewarp, line 200: 😅

bruno-fs: help

bruno-fs: this is the closest I found for an answer, but it's for stack overflow/exchange/&friends https://meta.stackexchange.com/a/371460

quipucords/utils/system_profile.py
```diff
@@ -0,0 +1,60 @@
...
+            .contents
+        )
+        # required for $ref resolution
+        field_schema["$defs"] = (
+            _system_profile().resolver().lookup(f"{URI}#/$defs").contents
+        )
+        # jsonschema.validate returns None - it raises exceptions on validation failures
+        jsonschema.validate(
+            value, schema=field_schema, cls=jsonschema.validators.Draft4Validator
+        )
```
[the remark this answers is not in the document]
bruno-fs: yes, that's the exact place!

quipucords/api/scanjob/model.py
```diff
@@ -328,7 +332,10 @@ def queue(self):  # noqa: C901
             for source in self.sources.all():
                 source.most_recent_connect_scan = self
                 source.save()
-
+        # Create report
+        # TODO(?): while Deployments/Details report are not removed, replace their
+        # report_id with report.id (?)
```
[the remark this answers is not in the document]
abellotti: Are we migrating Deployment/Details reports to the new Report?  Or do we not care (i.e. just zap them) since we're essentially ephemeral until DiscoPop time?

quipucords/scanner/job.py
```diff
@@ -211,6 +212,15 @@ def run(self) -> AsyncResult:
...
             celery_signatures_by_source[source_id].append(signature)
+            # Normalization phase is exclusive to celery task runner and (at least for
+            # now) is not tied to a specific scan task. It will use a inspection task id
+            # to grab all the raw facts related to it.
+            if (
+                settings.QPC_ENABLE_NORMALIZATION_PHASE
+                and scan_type == ScanTask.SCAN_TYPE_INSPECT
+            ):
+                normalization_signature = normalize.si(scan_task_id, source_type)
+                celery_signatures_by_source[source_id].append(normalization_signature)
```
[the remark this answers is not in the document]
abellotti: that makes sense. `scanning`, grabbing the whole lot, `fingerprinting`, you're this host and you're that host (deduplication implied).

quipucords/scanner/normalizer.py
abellotti: Just to confirm the behavior, if a fact normalization is borked (returning None, None, or exception), it won't be deduplicated and will just be sitting there as another raw fact  (either just duplicated or really unique).

abellotti:
Looks good @bruno-fs 👍 , lotsa good moving parts here.  Looking forward to the zapping of Deployment/Details reports. 

Re: your open questions above, initial hunch answers in `BOLD`
- Should we support a data normalizing raw facts from old reports? `NO?` - they are old reports and we're not PoP yet.
- Will we deduplicate data? `YES` - raw facts deduped into Host.
from the most recent discussions, the answer seems to be yes, but we need a FIRM yes or no to influence this implementation
- Should Host replace SystemFingerprint model? (this would help with the issue of data duplication) `YES` - deduped and also keeps the model simple (and closer to insights inventory).

quipucords/api/scanjob/model.py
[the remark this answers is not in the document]
bruno-fs: we will 🪓 them when the time comes =D

Makefile
bruno-fs: After clarifying this with @kahowell, I believe this is fine =D

PR #2461, opened 2023-09-06
feat: collect IPv4 and MAC addresses via `ip address show` command

infinitewarp:
Since RHEL has deprecated `ifconfig` and it may not be present on modern target systems, we should also try using `/sbin/ip` to find IPv4 and MAC addresses. This PR adds collection of raw facts using `ip`, and when processing facts into system fingerprints, if the `ifconfig` tasks did not return anything, we now substitute with the values that the `ip` tasks returned.

I manually verified this by performing network scans against a RHEL9 VM. Between scans, I would `sudo dnf install net-tools` to make `ifconfig` available or `sudo dnf remove net-tools` to make `ifconfig` unavailable, and I compared the different reports to verify which raw facts were populating the `ip_addresses` and `mac_addresses` fingerprints and that their values appeared to be correct.

Relates to JIRA: DISCOVERY-412
https://issues.redhat.com/browse/DISCOVERY-412


quipucords/scanner/network/runner/roles/ip/tasks/main.yml
```diff
@@ -0,0 +1,21 @@
...
+  ignore_errors: yes
+  become: '{{user_has_sudo}}'
+  when: 'internal_have_ip'
+
+- name: extract result value for 'ip address show'
+  set_fact:
+    # NOTE: We want to process the related output in two different ways, but the current
+    # quipucords Processor class behavior requires us to include is as two facts.
+    ip_address_show_ipv4: "{{ internal_ip_address_show_cmd | default(None) }}"
+    ip_address_show_mac: "{{ internal_ip_address_show_cmd | default(None) }}"
```
bruno-fs, line 19: what if we go NUTS and save only one raw fact, skip the processor, and do that processing on fingerprinting instead? please PUSH BACK XD

bruno-fs: that way the output in details report will truly be a raw fact heh

infinitewarp: Hmm… That's not a bad idea. I might be eager to make a more aggressive change like that _if_ we were dropping the old `ifconfig` code, _but_ since we are keeping it, I want to minimize the blast radius here, and I don't want to touch any of the `ifconfig` code unless strictly necessary.

This assumes we _do_ want to keep the old `ifconfig`, though, which I assume we do since it probably still works on older versions of RHEL. Lots of qualifiers here because I'm afraid of breaking support for old systems as part of a bug fix for newer systems. 🙂

bruno-fs (approved):
image

infinitewarp:
continuous-integration/jenkins tests are currently failing because camayoc does not expect these new facts `{'ip_address_show_ipv4', 'ip_address_show_mac', 'internal_have_ip'}`. We agreed to merge this change, accepting that risk, and we will update camayoc with these new facts very soon.

PR #2462, opened 2023-09-07
feat: add ansible logs to report tarball

bruno-fs:
Add support to report tarball view for including scan job-related logs.
For now this feature is limited to ansible logs collected during inspection phase of network scans.

Also update factories to support the generation of a valid tarball with fake-powered facts.

Most of this work will be backported to 1.3 release branch, excluding the commit that updates our factories and the unit test for report tarball view (which depends on the updated factories). Considering there's also an integration test that will be backported to 1.3, I think that's fine.

Reviewing commit per commit is recommended.


.gitignore
```diff
@@ -113,10 +112,6 @@ ENV/
 # mypy
 .mypy_cache/
 
-# sqlite
-db.sqlite3
-db.sqlite3-journal
-
```
infinitewarp, line 119: Why are we removing these from gitignore?

quipucords/quipucords/settings.py
```diff
@@ -283,9 +283,12 @@ def app_secret_key_and_path():
 VERBOSE_FORMATTING = (
     "%(levelname)s %(asctime)s %(module)s %(process)d %(thread)d %(message)s"
 )
-
-LOG_DIRECTORY = Path(env.str("LOG_DIRECTORY", str(BASE_DIR)))
+LOG_DIRECTORY = Path(env.str("QPC_LOG_DIRECTORY", str(DEFAULT_DATA_DIR / "logs")))
+# for convenience, create the default log directory
+if LOG_DIRECTORY.parent == DEFAULT_DATA_DIR:
+    LOG_DIRECTORY.mkdir(exist_ok=True, parents=True)
```
infinitewarp, line 289: I understand the convenience of this, but altering the filesystem in the settings module feels icky to me. 🤔

.gitignore
bruno-fs: because of the move from their default location (under quipucords/quipucords) to ./var, which is already on .gitignore

quipucords/quipucords/settings.py
bruno-fs: fair enough.

.gitignore
infinitewarp: oh nice I didn't catch that!

quipucords/quipucords/settings.py
bruno-fs: this was handled in a separate PR

quipucords/tests/integration/conftest.py
```diff
@@ -1,4 +1,13 @@
-"""Common fixtures for integration tests."""
+"""
+Common fixtures for integration tests.
+
+pytest_docker_tools is used to create container images and container fixtures.
+Besides some specifics, it is mostly passing kwargs to docker sdk.
+
+Links to relevant documentation
+- https://github.com/Jc2k/pytest-docker-tools#containers
+- https://docker-py.readthedocs.io/en/stable/containers.html
```
bruno-fs, line 9: Added this here because during development I completely forgot how to play around with pytest-docker-tools

conftest.py
```diff
@@ -19,6 +20,16 @@ def disable_celery_until_we_are_ready(settings):
     settings.QPC_ENABLE_CELERY_SCAN_MANAGER = False
 
 
+@pytest.fixture(autouse=True)
+def default_data_dir(settings, tmp_path: Path) -> Path:
+    """Set a clean data dir for storing files."""
+    data_dir = tmp_path / "data"
+    settings.DEFAULT_DATA_DIR = data_dir
+    settings.LOG_DIRECTORY = data_dir / "logs"
```
infinitewarp, line 28: Instead of monkey-patching settings directly, did you consider the `override_settings` context manager from `django.test`? That way it cleans up after itself if we ever _don't_ want this fixture (I did noticed the `autouse=true` above).

bruno-fs: huh. I thought the settings fixture was exactly the same thing as override_settings

bruno-fs: > This fixture **[settings]** will provide a handle on the Django settings module, and automatically revert any changes made to the settings (modifications, additions and deletions).

https://pytest-django.readthedocs.io/en/latest/helpers.html#settings

quipucords/api/details_report/model.py
```diff
@@ -15,7 +16,7 @@ class DetailsReport(models.Model):
     )
     report_version = models.CharField(max_length=64, null=False)
     report_platform_id = models.UUIDField(default=uuid.uuid4, editable=False)
-    sources = models.JSONField(null=False, default=list)
+    sources = models.JSONField(null=False, default=list, encoder=RawFactEncoder)
```
infinitewarp, line 19: Why is this change part of this PR? How exactly does the sources model encoder affect collection of Ansible logs?

Are this and its related changes (like the test) what you suggested would be skipped when we port this change to 1.3?

conftest.py
infinitewarp: Ah, I overlooked that you were using the settings fixture, not the settings module. Got it!

bruno-fs: We actually do want the `autouse` behavior because tests will mess with the default path.
They already were. They were reading `./quipucords/quipucords/secret.txt`, for instance.

I believe all places that can store data should be hermetically isolated in tests to avoid mixing local development files.

quipucords/api/details_report/model.py
bruno-fs: this is here on the same commit that introduces the automatic generation of raw facts to details report. as I pointed in the commit message
>refactor(DetailsReportFactory): generate faker-powered raw facts on the fly
>
>For convenience, DetailsReport.source is now using the same RawFact
    encoder used for raw facts. Even though this is not required for production
    code, it is technically the correct thing to do, since that field
    holds raw facts as well.
    
    
And yes, this is part of what I'd skip when backporting to 1.3

infinitewarp: I was reviewing the whole PR, not individual commits, and didn't see this message. Thank you for pointing that out.

bruno-fs: To be more concise: without this addition the factory for details report would break sometimes (when it randomly decide to pick OCP facts)

bruno-fs:
ouch, I forgot to add the "relates to Jira" thingy 🤦🏽

quipucords/api/reports/reports_gzip_renderer.py
```diff
@@ -53,6 +55,16 @@ def render(self, data, accepted_media_type=None, renderer_context=None):
             deployments_csv_name: encode_content(deployments_csv, "csv"),
         }
 
+        # collect associated logs
+        scan_job_id = reports_dict.get("scan_job_id")
+        log_directory: Path = settings.LOG_DIRECTORY
+        log_files = list(log_directory.glob(f"scan-job-{scan_job_id}-*"))
+        if not log_files:
+            logger.warning(f"No logs were found for {report_id=}")
```
infinitewarp, line 63: Not necessarily something we need to change here, but normally all variables should be parameterized in logger calls so that their execution/concatenation is deferred or not run at all if the log level isn't appropriate.

Feel free to resolve this comment thread.

infinitewarp (approved):

quipucords/api/reports/reports_gzip_renderer.py
bruno-fs: ouch, I forgot about it. TY

PR #2467, opened 2023-09-12
slightly improve test_reports.py

infinitewarp:
Among other recent changes, I was trying to update `test_reports.py` but struggled with some test assertions. This PR very slightly improves this test class:

* do not use plain `assert` style assertions inside a unittest test; you will lose the explanation message
* set maxdiff to increase the diff verbosity on assertion equality failures


nicolearagao (approved):
LGTM

PR #2468, opened 2023-09-12
refactor(insights): remove rh_installed_products

nicolearagao:
This function was deprecated on hbi side some time ago. Currently, if we want to send instaled_products information we have to do it through system profile.


infinitewarp (approved):
LGTM! 👍

bruno-fs (approved):

PR #2469, opened 2023-09-13
fix(insights): report correct cloud provider tag

nicolearagao:
Inventory UI expects GCP tag instead of GOOGLE and Discovery should use the same value.

Relates to JIRA: DISCOVERY-414
https://issues.redhat.com/browse/DISCOVERY-414


infinitewarp (approved):
🚀

PR #2470, opened 2023-09-13
refactor: drop support for masking report values

infinitewarp:
See also related change in qpc: https://github.com/quipucords/qpc/pull/276


bruno-fs (approved):
ty so much for this

PR #2471, opened 2023-09-13
fix: avoid unexpected stdout in network scan Ansible result

infinitewarp:
This changes some processing to read the last line of stdout instead of the first or to process stdout lines in reverse order. This should avoid some problems where a badly configured target system emits unexpected lines to stdout upon establishing the ssh connection.

Relates to JIRA: DISCOVERY-407
https://issues.redhat.com/browse/DISCOVERY-407

### DIFF
--- a/quipucords/scanner/network/processing/jws.py
+++ b/quipucords/scanner/network/processing/jws.py
@@ -12,10 +12,8 @@ class ProcessJWSInstalledWithRpm(process.Processor):
     @staticmethod
     def process(output, dependencies=None):
         """Determine if jws was installed with rpm. Version 3 and up."""
-        stdout = output.get("stdout_lines", [])
-        if len(stdout) == 1 and "Red Hat JBoss Web Server" in stdout[0]:
-            return True
-        return False
+        stdout_lines = output.get("stdout_lines", [])
+        return bool(stdout_lines and "Red Hat JBoss Web Server" in stdout_lines[-1])
 
 
 class ProcessHasJBossEULA(process.Processor):
@@ -26,8 +24,8 @@ class ProcessHasJBossEULA(process.Processor):
     @staticmethod
     def process(output, dependencies=None):
         """Check if JBossEULA.txt exists in JWS_Home directory."""
-        stdout = output.get("stdout", "false")
-        return stdout == "true"
+        stdout_lines = output.get("stdout_lines", [])
+        return bool(stdout_lines and stdout_lines[-1] == "true")
 
 
 class ProcessTomcatPartOfRedhatProduct(process.Processor):
@@ -38,10 +36,8 @@ class ProcessTomcatPartOfRedhatProduct(process.Processor):
     @staticmethod
     def process(output, dependencies=None):
         """Return either True or False."""
-        result = output.get("stdout_lines", False)
-        if result and result[0] == "True":
-            return True
-        return False
+        stdout_lines = output.get("stdout_lines", [])
+        return bool(stdout_lines and stdout_lines[-1] == "True")
 
 
 class ProcessJWSHasCert(process.Processor):
@@ -53,6 +49,5 @@ class ProcessJWSHasCert(process.Processor):
     @staticmethod
     def process(output, dependencies=None):
         """Return either True or False."""
-        if output.get("stdout_lines", False):
-            return True
-        return False
+        stdout_lines = output.get("stdout_lines", [])
+        return bool(stdout_lines and stdout_lines[-1].endswith(".pem"))
--- a/quipucords/scanner/network/processing/system_purpose.py
+++ b/quipucords/scanner/network/processing/system_purpose.py
@@ -2,6 +2,7 @@
 
 import json
 import logging
+import re
 
 from scanner.network.processing import process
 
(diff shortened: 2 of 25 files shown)


quipucords/tests/scanner/network/processing/test_cloud_provider.py
```diff
@@ -50,13 +50,19 @@ class TestProcessDmiSystemProductName(unittest.TestCase):
...
+        we always return only one of them: the content from the last line.
+        """
         dependencies = {"internal_dmi_system_product_name": ansible_result("a\nb\nc")}
         self.assertEqual(
             cloud_provider.ProcessDmiSystemProductName.process(
                 "QPC_FORCE_POST_PROCESS", dependencies
             ),
-            "a",
+            "c",
         )
```
bruno-fs, line 66: I'm wondering if in a real world scenario we actually need the first line tho. the test expects more lines here 🤔

quipucords/tests/scanner/network/processing/test_dmi.py
```diff
@@ -31,12 +37,13 @@ def test_success_case(self):
...
         """Test dmi_system_uuid too long."""
-        # stdout_lines looks like ['a', 'b', 'c']
-        dependencies = {"internal_dmi_system_uuid": ansible_result(f"{'a' * 37}\nb\nc")}
+        # stdout_lines looks like ['a', 'b', 'ccccccccccccccccccccccccccccccccccccc']
+        dependencies = {"internal_dmi_system_uuid": ansible_result(f"a\nb\n{'c' * 37}")}
         self.assertEqual(
-            dmi.ProcessDmiSystemUuid.process("QPC_FORCE_POST_PROCESS", dependencies), ""
+            dmi.ProcessDmiSystemUuid.process("QPC_FORCE_POST_PROCESS", dependencies),
+            "b",
         )
```
bruno-fs, line 45: this one shows that the output is wrong IMO. it should be "" considering "b" is not a valid uuid 🤔

quipucords/tests/scanner/network/processing/test_cloud_provider.py
infinitewarp: Yes, ***this is a bad test***, plain and simple. In this change, I am trying to resist rewriting all the related bad legacy code. I am focusing acutely on applying my fix and preserving their existing functionality and tests — warts and all — without thinking too hard about them. ☹️

quipucords/tests/scanner/network/processing/test_dmi.py
infinitewarp: See previous comment. I know this is a bad test, but I am not trying to reinvent legacy code or assumptions.

infinitewarp: This doesn't make me a happy developer, by the way, but I am setting limits on the scope of my work because I could spend all day/week/month/year improving bad old code like this.

bruno-fs: I see. I was concerned this test was reflecting some real world scenario.

quipucords/scanner/network/runner/roles/cpu/tasks/main.yml
```diff
@@ -45,7 +45,7 @@
   ignore_errors: yes
 
 - name: gather internal_cpu_model_ver fact
-  raw: cat /proc/cpuinfo 2>/dev/null | grep '^model\s*.' | sed -n -e 's/^.*model\s*.\s*//p'
+  raw: cat /proc/cpuinfo 2>/dev/null | grep '^model\s*:' | sed -n -e 's/^.*model\s*:\s*//p'
```
infinitewarp, line 48: Why did this regex change? Well, if you look at what the old command normally produces:

```
$ cat /proc/cpuinfo 2>/dev/null | grep '^model\s*.' | sed -n -e 's/^.*model\s*.\s*//p'
63
ame	: Intel(R) Xeon(R) CPU E5-2676 v3 @ 2.40GHz
```
See how there are TWO lines in the output? **This is a bug in the original implementation**, but we rarely (but probably _not never_) noticed it because normally we would just grab the first line of output, resulting in the value `63`. Order is not guaranteed, though, and it's possible that some reports out in the wild contained the wrong line with the model name instead of the model ID.

Remove the `sed`, and you can see more of the original output:

```
$ cat /proc/cpuinfo 2>/dev/null | grep '^model\s*.'
model		: 63
model name	: Intel(R) Xeon(R) CPU E5-2676 v3 @ 2.40GHz
```

Now that we are grabbing the _last_ line of output for commands like this, it's more important that this command only outputs the one line that is actually relevant. After changing the regex, the new output looks like:

```
$ cat /proc/cpuinfo 2>/dev/null | grep '^model\s*:' | sed -n -e 's/^.*model\s*:\s*//p'
63
```

Much better.

quipucords/scanner/network/processing/system_purpose.py
```diff
@@ -16,12 +17,19 @@ class ProcessSystemPurpose(process.Processor):
...
             except json.decoder.JSONDecodeError as error:
                 logger.warning("system_purpose was not valid JSON.  Error: %s", error)
-                return None
+            # We anticipate in some edge cases that unwanted non-JSON outputs may be
+            # included in the stdout stream before the actual JSON output, but we expect
+            # they are on their own lines. So, chop the first line, try, and repeat.
+            # It's a brute-force solution, but this should help in those edge cases!
+            if "\n" not in system_purpose_str:
+                break
+            system_purpose_str = re.sub(r"^.*?\n", "", system_purpose_str)
```
bruno-fs, line 34: :big_brains:

quipucords/scanner/network/processing/jws.py
```diff
@@ -53,6 +49,5 @@ class ProcessJWSHasCert(process.Processor):
     @staticmethod
     def process(output, dependencies=None):
         """Return either True or False."""
-        if output.get("stdout_lines", False):
-            return True
-        return False
+        stdout_lines = output.get("stdout_lines", [])
+        return bool(stdout_lines and stdout_lines[-1].endswith(".pem"))
```
gtanzillo, line 53: This is good, I think the extra check for a `.pem` file is important

bruno-fs (commented):
:shipit:

quipucords/scanner/network/runner/roles/date/tasks/main.yml
```diff
@@ -11,7 +11,7 @@
 
 - name: set date_date fact
   set_fact:
-    date_date: "{{internal_date_date}}"
+    date_date: "{{internal_date_date }}"
```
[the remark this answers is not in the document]
infinitewarp: Fixed in the latest push with an amended commit.

nicolearagao (approved):

PR #2472, opened 2023-09-13
chore(reports-view): small cleanup

bruno-fs:
Since the url for reports tarball view gives no margin for a None report id, the error treatment the view had for report_id is completely useless (django is already doing that for us).

### DIFF
--- a/quipucords/api/reports/view.py
+++ b/quipucords/api/reports/view.py
@@ -14,10 +14,7 @@ from rest_framework.decorators import (
 )
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
-from rest_framework.serializers import ValidationError
 
-from api import messages
-from api.common.util import is_int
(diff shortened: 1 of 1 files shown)


quipucords/api/reports/view.py
```diff
@@ -34,35 +31,33 @@
...
 @permission_classes(perm_classes)
 @renderer_classes((ReportsGzipRenderer,))
-def reports(request, report_id=None):
+def reports(request, report_id):
     """Lookup and return reports."""
     reports_dict = {}
-    if report_id is not None:
-        if not is_int(report_id):
-            error = {"report_id": [_(messages.COMMON_ID_INV)]}
-            raise ValidationError(error)
```
bruno-fs, line 43: this part of the code is unreachable given how the URL for this view is set.

let Django do it's job and move on

abellotti (approved):

infinitewarp (approved):

PR #2473, opened 2023-09-14
chore: add a vagrant config to spin multiple vms

bruno-fs:
QoL for development. Easily spin multiple vms to run network scans locally.

My personal cheat sheet for vagrant:
```
# spin all vms
vagrant up
# get the ips
vagrant ssh-config | grep HostName | awk '{print $2}'
# tear everything down
vagrant destroy -f
```

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,21 @@
+Vagrant.configure("2") do |config|
+  # unnoficial RHEL images
+  (6..9).each do |n|
+    config.vm.define "rhel%d" % n do |vmconfig|
+      vmconfig.vm.box = "generic/rhel%d" % n
+    end
+  end
+  # unnoficial centos stream images (for testing purposes we can consider them RHEL-like)
+  (8..9).each do |n|
+    config.vm.define "stream%d" % n do |vmconfig|
+      # we could use the official "centos/stream%d" images, but I'm unable to boot stream9 :'(
+      vmconfig.vm.box = "generic/centos%ds" % n
+    end
+  end
+  # allow ssh'ing with password authentication (username and password is "vagrant")
+  config.vm.provision "shell", inline: <<-EOF
+    sed -i 's/PasswordAuthentication no/PasswordAuthentication yes/g' /etc/ssh/sshd_config
+    # fallback to service for systems that don't use systemd
+    systemctl restart sshd.service || service sshd restart
+  EOF
+end
--- a/.gitignore
+++ b/.gitignore
@@ -128,3 +128,6 @@ quipucords/roles/
 
 # local files
 var/
+
+# vagrant data
+.vagrant/


Vagrantfile
```diff
@@ -0,0 +1,21 @@
...
+  # unnoficial RHEL images
+  (6..9).each do |n|
+    config.vm.define "rhel%d" % n do |vmconfig|
+      vmconfig.vm.box = "generic/rhel%d" % n
+    end
+  end
+  # unnoficial centos stream images (for testing purposes we can consider them RHEL)
+  (8..9).each do |n|
+    config.vm.define "stream%d" % n do |vmconfig|
+      # we could use the official "centos/stream%d" images, but I'm unable to boot stream9 :'(
```
bruno-fs, line 11: also, the download for the official images is pretty slow

bruno-fs:
Ah, the oneliner for grabbing IPs might not be enough depending on provider.
You might need to grab port too.

mirekdlugosz (approved):
LGTM :+1:

infinitewarp (changes requested):
I took this branch for a test drive, and your `Vagrantfile` mostly works for me!

[asciicast](https://asciinema.org/a/612231)

The RHEL6 VM encountered an error running `systemctl`, and when everything was up, I wasn't able to SSH into the RHEL6 VM, but I didn't spend any time troubleshooting it. Is RHEL6 too old for the `systemctl` command? Did the RHEL6 VM setup work for you? Maybe I need to add a local `~/.ssh/config` to allow the old weak `ssh-rsa`.

I successfully SSHed into all the other VMs, though! 🎉  I needed to include the port numbers when grepping the ssh-config like this:

```
vagrant ssh-config 2>/dev/null | grep -e '^Host\|\sHostName\|\sPort'
```

Vagrantfile
```diff
@@ -0,0 +1,21 @@
...
+      vmconfig.vm.box = "generic/centos%ds" % n
+    end
+  end
+  # allow ssh'ing with password authentication (username and password is "vagrant")
+  config.vm.provision "shell", inline: <<-EOF
+    sed -i 's/PasswordAuthentication no/PasswordAuthentication yes/g' /etc/ssh/sshd_config
+    # fallback to service for systems that don't use systemd
+    systemctl restart sshd.service || service sshd restart
```
bruno-fs, line 19: rhel6 compatibility

bruno-fs:
> Is RHEL6 too old for the systemctl command? 

It is. `service` is there for RHEL6 (the error message can be safely ignored).
https://github.com/quipucords/quipucords/pull/2473/files#r1347832504

> Did the RHEL6 VM setup work for you? Maybe I need to add a local ~/.ssh/config to allow the old weak ssh-rsa.

yup, it works, but I had to enable support for less secure kex algorithms

because this is so frequent (testlab running rhel6, ec2 running rhel6), I added to all my hosts on ~/.ssh/config
```
Host *
   HostkeyAlgorithms +ssh-rsa
   PubkeyAcceptedAlgorithms +ssh-rsa
   KexAlgorithms +diffie-hellman-group1-sha1
```
This is virtually what we have on discovery by enabling legacy compatibility 

https://github.com/quipucords/quipucords/blob/2e908b73048bd54df5f166bc6a1ffe34e7f6ac9e/Dockerfile#L37-L38

Vagrantfile
bruno-fs: I'm expecting it to fail on rhel 6 because it doesn't use systems. then it will fallback to service.

infinitewarp: D'oh! I completely overlooked that "or" half of that command. Good thinking.

infinitewarp (approved):
🚀

PR #2474, opened 2023-09-14
Port API tests from Camayoc

mirekdlugosz:
I would like to reduce duplication between Camayoc and Quipucords unit tests. If something can be tested at unit level, it should be tested at unit level.

This is the port of https://github.com/quipucords/camayoc/blob/main/camayoc/tests/qpc/api/v1/credentials/test_manager_creds.py 

I am opening this to get early feedback. Mostly: how do people feel about extending existing unit tests to cover wider combination of things?   
What about readability and short variable names here? Should I move these parameters generation to separate functions?

I intend to add more ported tests before opening it for merge. And I assume there are going to be multiple PRs like this one.

### DIFF
--- a/quipucords/tests/api/credential/test_credential.py
+++ b/quipucords/tests/api/credential/test_credential.py
@@ -95,11 +95,14 @@ class TestCredential:
         host_cred = self.create_credential()
         assert isinstance(host_cred, Credential)
 
-    def test_hostcred_create(self, django_client):
+    @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


bruno-fs (commented):
LGTM 👍🏽

abellotti:
I totally agree on unit test first and if not appropriate/doable there, then camayoc. No need to duplicate. 👍

bruno-fs (approved):

infinitewarp (approved):

ruda (approved):
LGTM.

PR #2481, opened 2023-09-23
test: randomize db sequences

bruno-fs:
Randomizing database sequences to:
1. Force tests to not hardcode ids
2. Avoid false positive tests passing only because models had coinciding ids
3. Attempt to simulate a more real world like scenario

In the future this fixture could be an util used to highlight and help debugging the UI issues with incorrect ids being used.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -11,7 +11,18 @@ from urllib.parse import urljoin
 
 import pytest
 from django.contrib.auth import get_user_model
+from django.core import management
 from django.test import override_settings
+
+
+@pytest.fixture(scope="session")
+def django_db_setup(django_db_setup, django_db_blocker):
+    """Force db sequences to use random numbers."""
+    with django_db_blocker.unblock():
+        # randomize db sequences to avoid ID coincidences causing
+        # false positive tests, avoid hardcoded IDs, and attempt
+        # to simulate a real world-like scenario.
+        management.call_command("randomize_db_sequences")
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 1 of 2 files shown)


conftest.py
```diff
@@ -5,15 +5,45 @@
...
+@pytest.fixture(scope="session")
+def django_db_setup(django_db_setup, django_db_blocker):
+    """Force db sequences to use random numbers."""
+    with django_db_blocker.unblock():
+        randomize_db_sequences()
+
+
+def randomize_db_sequences():
+    """Randomize db sequences."""
+    # shamelessly inspired on https://pytest-django.readthedocs.io/en/latest/database.html#randomize-database-sequences
```
mirekdlugosz, line 29: I don't have any qualm on which one is better, but this popped up in my feed recently: https://adamj.eu/tech/2023/07/04/django-test-random-order-querysets/

conftest.py
```diff
@@ -5,15 +5,45 @@
...
 
 
+@pytest.fixture(scope="session")
+def django_db_setup(django_db_setup, django_db_blocker):
+    """Force db sequences to use random numbers."""
+    with django_db_blocker.unblock():
+        randomize_db_sequences()
+
+
+def randomize_db_sequences():
```
infinitewarp, line 27: How would you feel about moving this outside of `conftest` and making a small Django management command wrapper for it? I would love for our initial setup instructions to include randomizing the sequences immediately after database creation because it's not just our tests that suffer from ugly ID synchronization.

infinitewarp (approved):

PR #2483, opened 2023-09-25
DISCOVERY-435: Authorize GitHub requests if token is available

mirekdlugosz:
Context: On internal Jenkins, `make build-container` may fail because GitHub rate-limits API requests, and a lot of internal traffic appears as a single IP to GitHub.

With this change, we can use GitHub token to authenticate API request during `make build-container`. Jenkins pipeline will use that capability, but it remains optional for everyone else.

The easy way would be to pass environment variable using `podman --env`, but then token will be embedded in final image layer. The correct way is to use docker / podman `--secret` flag. Unfortunately, this flag requires file with secret to exist when running `podman build`, so I have to add new empty file to repo.

Since secrets are not part of built layer, changing secrets file content will not invalidate cached layers. So if you run `make build-container`, change content of `.github_api_token` and run `make build-container` again, you will end up with exactly the same image. You can modify Makefile to run `podman build --no-cache`, or clear locally cached layers through other means.

I have not tested this with real Docker, but `podman` claims compatibility with Docker in this use case. Docker supports some other options that podman does not.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,11 @@ TEST_OPTS := -n $(PARALLEL_NUM) -ra -m 'not slow' --timeout=15
 QUIPUCORDS_CONTAINER_TAG ?= quipucords
 QUIPUCORDS_UI_PATH ?= ../quipucords-ui
 QUIPUCORDS_UI_RELEASE ?= latest
+GITHUB_API_TOKEN_SECRET ?= /run/secrets/gh_api_token
+GITHUB_API_TOKEN := $(file < $(GITHUB_API_TOKEN_SECRET))
+ifneq ($(GITHUB_API_TOKEN),)
+	GITHUB_API_AUTH = -H "Authorization: Bearer $(GITHUB_API_TOKEN)"
+endif
 
 ifndef DOCKER_HOST
 	PODMAN_SOCKET := $(shell podman machine inspect --format '{{.ConnectionInfo.PodmanSocket.Path}}' 2> /dev/null || podman info --format '{{.Host.RemoteSocket.Path}}' 2> /dev/null || echo -n "")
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -6,12 +6,30 @@ DIRS	= test bin locale src
...
+GITHUB_API_TOKEN := $(file < $(GITHUB_API_TOKEN_SECRET))
+ifneq ($(GITHUB_API_TOKEN),)
+	GITHUB_API_AUTH = -H "Authorization: Bearer $(GITHUB_API_TOKEN)"
+endif
+
+ifndef DOCKER_HOST
+	PODMAN_SOCKET := $(shell podman machine inspect --format '{{.ConnectionInfo.PodmanSocket.Path}}' 2> /dev/null)
+	ifneq ($(.SHELLSTATUS), 0)
+		PODMAN_SOCKET := $(shell podman info --format '{{.Host.RemoteSocket.Path}}' 2> /dev/null)
+	endif
```
infinitewarp, line 25:
```suggestion
	PODMAN_SOCKET := $(shell podman machine inspect --format '{{.ConnectionInfo.PodmanSocket.Path}}' 2> /dev/null || podman info --format '{{.Host.RemoteSocket.Path}}' 2> /dev/null)
```

…due to issue I discussed in Slack here: https://redhat-internal.slack.com/archives/C02QSNF1UKE/p1695931278695499?thread_ts=1695910544.300239&cid=C02QSNF1UKE

But we need to verify that this version also works on podman _without_ a machine VM running. 😕

mirekdlugosz: I took a slightly different version of this approach, with additional `|| echo -n ""`. The idea is that if you don't have podman installed, then PODMAN_SOCKET will be empty, and next conditional will be skipped.

On machine without VM, `podman machine inspect` returns 125 and reports that VM is missing. So it will fall back to `podman info`.

One problem I noticed is that if you have podman machine set up, but VM is shut down, `podman machine inspect` will still report socket, which does exist. So it will take VM path, and all tests are going to fail. Since macOS users are likely going to have VM started (or will start it when they notice all tests failing), and Linux users don't really have a reason to use podman with VM, I think it's alright to just leave it as it is.

PR #2485, opened 2023-09-25
feat: randomize sequences after migrating database

infinitewarp:
Thanks to @bruno-fs for making this possible with quipucords/pull/2481! 🎉 

I want this in the make target so I don't need an extra step for it in my sandbox DB rebuild routine…

```
dropdb -U qpc qpc \
&& createdb -U qpc qpc \
&& poetry install \
&& make server-migrate \
&& poetry run python ./quipucords/manage.py randomize_db_sequences \
&& make server-set-superuser serve
```

Since the makefile is chiefly for us developers/maintainers anyway, I think we should *always* be running the sequence randomizer to ensure we don't miss any unfortunate ID collisions. (I'm looking at you, report IDs…)

Now my local command is simply:

```
dropdb -U qpc qpc \
&& createdb -U qpc qpc \
&& poetry install \
&& make server-migrate server-randomize-sequences server-set-superuser serve
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,6 +132,9 @@ server-makemigrations:
 server-migrate:
 	$(PYTHON) quipucords/manage.py migrate --settings quipucords.settings -v 3
 
+server-randomize-sequences:
+	$(PYTHON) quipucords/manage.py randomize_db_sequences --settings quipucords.settings
+
 celery-worker:
 	$(PYTHON) -m celery --app quipucords --workdir quipucords worker
 


Makefile
```diff
@@ -131,6 +131,7 @@ server-makemigrations:
 
 server-migrate:
 	$(PYTHON) quipucords/manage.py migrate --settings quipucords.settings -v 3
+	$(PYTHON) quipucords/manage.py randomize_db_sequences
```
bruno-fs, line 134: I'd change put this in a separate `make role` because production does depend on `make server-migrate`

https://github.com/quipucords/quipucords/blob/61610c14c5d757b59569c34624ced23d16adad8c/deploy/entrypoint_web.sh#L15

bruno-fs: and the reason I insist this to be a separate command is because we know that randomizing db sequences will horribly bork the UI

infinitewarp: Fair enough. The UI is already horribly broken, but I guess we shouldn't make it _obvious_ that it's already horribly broken. 😕 

FWIW, it seems trivially easy to encounter this ugly UI ID bug _without_ randomizing the database sequences. I've been hitting it somewhat regularly since I joined the project many months ago.

infinitewarp: Done!

It's `make server-randomize-sequences` to try to follow the current `server-` naming convention.

bruno-fs (approved):

PR #2488, opened 2023-09-26
chore(refact): Removing the sync merge report API

abellotti:
It is no longer used by the UI and the CLI reports merge has already been updated to using the async merge report API.

Related PRs:
- https://github.com/quipucords/quipucords-ui/pull/243
- https://github.com/quipucords/camayoc/pull/448


bruno-fs (approved):
:shipit:

mirekdlugosz:
I'm looking at tests removed in Camayoc PR and unit tests for async merge.

It's not clear to me what is the difference between "Test Async Report Merge" and "Test PUT Async Report Merge", except for input JSON format.

I see following cases covered in Camayoc, but I can't find their equivalent in unit tests. I would suggest to add them as part of this PR, just to keep the coverage:

* "Test Async Report Merge" (one with `sources` structure) - a test that tries to add 2 identical sources
* "Test PUT Async Report Merge" (one with `reports`) - a test where `reports` is empty list (right now this is the same as `test_by_id_merge_jobs_list_too_short`, but the idea is to detect if empty list is ever treated differently than one-element list)

abellotti:
Thanks @mirekdlugosz for the review, I'll add the tests. Also, since I'm touching this file, I'll update it to pytest.

mirekdlugosz (approved):
I really only looked into 17e9efa41b5fc24ac0e2a6552199d9988bfe26c7 . Thanks for including these tests :+1:

PR #2490, opened 2023-09-29
refactor(source): unittest to pytest.

nicolearagao:
If you touch it, you refactor it. =)


bruno-fs (approved):
I must admit I'd rewrite it from scratch and use factories, but it's a fine conversion anyways. We can refactor to use factory boy in the future.

PR #2492, opened 2023-10-02
feature: base normalizer

bruno-fs:
Adds a base normalizer that will work as a foundation for the normalization phase.
Also, add a normalizer for network scans to have a more practical example.

This PR is a spiritual successor for its original RFC (#2442)

Relates to JIRA: DISCOVERY-226

### DIFF
--- a/quipucords/scanner/network/normalizer.py
+++ b/quipucords/scanner/network/normalizer.py
@@ -1,0 +1,78 @@
+"""normalizer for network scans."""
+
+import ipaddress
+
+from constants import DataSources
+from fingerprinter import formatters
+from scanner.normalizer import BaseNormalizer, FactMapper, NormalizedResult, str_or_none
+
+
+def infrastructure_type_normalizer(
+    *, virt_what_type=None, virt_type=None, subman_virt_is_guest=None
+):
+    """Normalize infrastructure type."""
+    if virt_what_type == "bare metal":
+        return NormalizedResult("physical", raw_fact_keys=["virt_what_type"])
+    elif virt_type:
+        return NormalizedResult("virtualized", raw_fact_keys=["virt_type"])
+    elif subman_virt_is_guest:
+        # We don't know virt_type, but subscription-manager says it's a guest.
+        # So, we assume it's virtualized. See also: DISCOVERY-243.
+        return NormalizedResult("virtualized", raw_fact_keys=["subman_virt_is_guest"])
+    return NormalizedResult(None, raw_fact_keys=None)
+
+
+def ip_address_normalizer(ifconfig_ip_addresses: list[str]):
+    """Normalize ip_address."""
+    # https://github.com/quipucords/quipucords/blob/e65d0e3deb23adca41ceed701e4d002187e593f6/quipucords/api/common/entities.py#L124-L136
+    if ifconfig_ip_addresses is None:
+        return []
+    addresses = []
+    for raw_ip in ifconfig_ip_addresses:
+        ip_addr = ipaddress.ip_address(raw_ip)
+        # raw fact is only collecting ipv4 addresses
+        if isinstance(ip_addr, ipaddress.IPv4Address):
+            addresses.append(str(ip_addr))
+    return addresses
+
+
+def network_interfaces(ip_addresses: list[str]):
+    """Normalize network interfaces."""
+    interface = {
+        "name": "unknown",  # required by yuptoo
+        "ipv4_addresses": ip_addresses or [],
+        "ipv6_addresses": [],  # required by yuptoo
+    }
+    return [interface]
+
+
+class Normalizer(BaseNormalizer):
+    """Network Scan normalizer."""
+
+    source_type = DataSources.NETWORK.value
+
+    mac_addresses = FactMapper(
+        "ifconfig_mac_addresses", formatters.format_mac_addresses
+    )
+    ip_addresses = FactMapper("ifconfig_ip_addresses", ip_address_normalizer)
+    bios_uuid = FactMapper("dmi_system_uuid", str_or_none)
+    insights_id = FactMapper("insights_client_id", str_or_none)
+    subscription_manager_id = FactMapper("subscription_manager_id", str_or_none)
+    number_of_cpus = FactMapper("cpu_core_count", int)
+    number_of_sockets = FactMapper("cpu_socket_count", int)
+    cores_per_socket = FactMapper("cpu_core_per_socket", int)
+    system_memory_bytes = FactMapper("system_memory_bytes", int)
+    infrastructure_type = FactMapper(
+        ["virt_what_type", "virt_type", "subman_virt_is_guest"],
+        infrastructure_type_normalizer,
+    )
+    infrastructure_vendor = FactMapper("virt_type", str_or_none)
+    os_release = FactMapper("etc_release_release", str_or_none)
+    arch = FactMapper("uname_processor", str_or_none)
+    cloud_provider = FactMapper("cloud_provider", str_or_none)
+    system_purpose = FactMapper("system_purpose_json", dict)
+    network_interfaces = FactMapper(
+        None, network_interfaces, dependencies=["ip_addresses"]
+    )
+    # ----- non canonical/system profile facts ---
+    etc_machine_id = FactMapper("etc_machine_id", str_or_none)
--- a/quipucords/scanner/normalizer.py
+++ b/quipucords/scanner/normalizer.py
@@ -1,0 +1,175 @@
+"""normalizer module."""
+
+from logging import getLogger
+
+logger = getLogger(__name__)
+
+
+class NormalizerMeta(type):
+    """
+    Type for Normalizer classes.
+
+    Customize class creation to provide the necessary magic for Normalizers.
+
+    This metaclass is meant to be used ONLY on BaseNormalizer.
+    """
+
+    def __new__(cls, name, bases, attrs):
+        """Blueprint for Normalizer classes."""
+        # we need fields to be accessible as a class attribute. This is the WHOLE reason
+        # for the existence of this metaclass (I wish python had a simpler api for this
+        # type of thing).
+        # Usually just creating a attribute on a base class would be enough
+        # for this behavior, but not when you are dealing with dicts or lists.
+        # (Creating 'fields' as a normal attribute on 'BaseNormalizer' class would
+        # result on it being shared between the base class and its subclasses.)
+        attrs["fields"] = {}
+        return super().__new__(cls, name, bases, attrs)
+
+
+class BaseNormalizer(metaclass=NormalizerMeta):
+    """Base class for Normalizer classes."""
+
+    # must be set for other implementations
+    source_type = None
+
+    def __init__(self, raw_facts: dict, server_id: str = None) -> None:
+        """Create a normalizer instance.
+
+        :param raw_facts: dict of raw facts from inspection phase
+        :param server_id: server_id of quipucords server that collected these
+        raw facts.
+        """
+        self.raw_facts = raw_facts
+        self._metadata = {}
+        self.server_id = server_id
+
+    @property
+    def facts(self):
+        """Return normalized facts."""
+        if not hasattr(self, "_normalized"):
+            raise AssertionError(
+                "You must call '.normalize()' before acessing `.facts`."
+            )
+        return self._normalized
+
+    @property
+    def metadata(self):
+        """Return metadata for normalized facts."""
+        if not hasattr(self, "_normalized"):
+            raise AssertionError(
+                "You must call '.normalize()' before acessing `.metadata`."
+            )
+        return self._metadata
+
+    def normalize(self):
+        """Normalize and validate raw facts.
+
+        :return: dict of normalized facts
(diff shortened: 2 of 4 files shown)


quipucords/scanner/normalizer.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def __new__(cls, name, bases, attrs):
+        """Blueprint for Normalizer classes."""
+        # we need fields to be accessible as a class attribute. This is the WHOLE reason
+        # for the existence of this metaclass (I wish python had a simpler api for this
+        # type of thing).
+        # Usually just creating a attribute on a base class would be enough
+        # for this behavior, but not when you are dealing with dicts or lists.
+        # (Creating 'fields' as a normal attribute on 'BaseNormalizer' class would
+        # result on it being shared between the base class and its subclasses.)
```
mirekdlugosz, line 27: I don't understand the part about dicts and lists, but can't we just have a [base object implementing `__getattr__`](https://docs.python.org/3/reference/datamodel.html#object.__getattr__)? That will be called when you try to access `object.field_name`, assuming there's no `field_name` in `object`, so you can check in `object.fields` or whatever before failing.

mirekdlugosz: I think I understand the dicts lists comment. This depends on when attribute is defined - it if is defined as class attribute (outside of init or any method), then yes, it will be shared between all instances. If it is defined in method, it is exclusive to instance:

```python
class A:
    class_fields = []
    def __init__(self):
        self.fields = []

    def add_to_fields(self, value):
        self.fields.append(value)
        print("{}.fields of obj {} is set to {}".format(
            self.__class__.__name__, str(id(self)), str(self.fields)
        ))

    def add_to_class_fields(self, value):
        self.class_fields.append(value)
        print("{}.class_fields of obj {} is set to {}".format(
            self.__class__.__name__, str(id(self)), str(self.class_fields)
        ))


class B(A):
    pass

one = B()
two = B()

one.add_to_fields(1)
two.add_to_fields(2)

one.add_to_class_fields(1)
two.add_to_class_fields(2)
```
```
B.fields of obj 139825975519056 is set to [1]
B.fields of obj 139825975519120 is set to [2]
B.class_fields of obj 139825975519056 is set to [1]
B.class_fields of obj 139825975519120 is set to [1, 2]
```

quipucords/scanner/normalizer.py
```diff
@@ -0,0 +1,225 @@
...
+        self.raw_facts = raw_facts
+        self._metadata = {}
+        self.server_id = server_id
+
+    @property
+    def facts(self):
+        """Return normalized facts."""
+        assert hasattr(
+            self, "_normalized"
+        ), "Call 'normalize()' before asking for facts."
```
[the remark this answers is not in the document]
bruno-fs: I forgot about that. Some linters even complain about asserts in the middle of non test code because of that...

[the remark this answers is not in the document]
bruno-fs: this is where we need to iterate over fields following the order they were created

https://github.com/quipucords/quipucords/blob/fd552f9f95af83c617f7d152d823cc9528f860fe/quipucords/scanner/normalizer.py#L73

And this is required because I considered preserving order + checking at creation time if the dependency field do exist (implementation https://github.com/quipucords/quipucords/blob/fd552f9f95af83c617f7d152d823cc9528f860fe/quipucords/scanner/normalizer.py#L201-L205 + test https://github.com/quipucords/quipucords/blob/911326a4c7cd4446a6b4c5dc48b06e9cad7b868b/quipucords/tests/scanner/test_normalizer.py#L28-L38) the best heuristics to ensure field dependencies would be respected.

I like this idea because if a normalizer is "incorrectly" implemented it would fail ASAP - and is relatively simple.

But if you have an idea that would completely avoid the metaclass I'm all ears =D

quipucords/scanner/normalizer.py
bruno-fs: Also, FactMapper depends on `fields` existing at class level.

quipucords/scanner/normalizer.py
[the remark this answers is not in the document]
bruno-fs: exluding tests from selection, I'll have to change some code to remove assertions from production code
```
ruff check --select=S101 .
quipucords/api/source/serializer.py:516:17: S101 Use of `assert` detected
quipucords/scanner/manager.py:43:9: S101 Use of `assert` detected
quipucords/scanner/manager.py:143:9: S101 Use of `assert` detected
quipucords/scanner/openshift/api.py:214:9: S101 Use of `assert` detected
quipucords/scanner/openshift/entities.py:59:9: S101 Use of `assert` detected
quipucords/scanner/openshift/entities.py:60:9: S101 Use of `assert` detected
Found 6 errors.
```

quipucords/scanner/normalizer.py
mirekdlugosz: I gave it a little thought and there _might_ be three ways to avoid metaclass:

1. Keep `fields` shared class attribute, but instead of it being key -> fact mapping, make it child class "path" (say: `tuple(self.__class__.mro())`) -> key -> fact multi-level mapping.

2. Let `FactMapper.__set_name__` set normalizer class `fields`.

3. Make `FactMapper` lazy - they would not look into normalizer `fields` until specific field is actually requested.

Now, all three of them have major drawbacks, and I am not convinced these drawbacks aren't making them actually _worse_ than using metaclass...

So let's see if anyone else has something to add, but I think we will leave that part as is. At least now I have better grasp _why_ this metaclass exists :+1:

quipucords/scanner/normalizer.py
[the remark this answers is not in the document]
mirekdlugosz: While I haven't checked, I assumed this is not the only place where `assert` is used. So that was more of an off-hand comment "hey, maybe we shouldn't be doing that", instead of hard requirement to change your PR.

quipucords/scanner/network/normalizer.py
```diff
@@ -0,0 +1,78 @@
...
+
+import ipaddress
+
+from constants import DataSources
+from fingerprinter import formatters
+from scanner.normalizer import BaseNormalizer, FactMapper, NormalizedResult, str_or_none
+
+
+def infrastructure_type_normalizer(
+    *, virt_what_type=None, virt_type=None, subman_virt_is_guest=None
```
infinitewarp, line 11: Nice use of `*`. I wonder if there are more places we should use this to force better readability.

quipucords/scanner/normalizer.py
bruno-fs: 2. seems a pretty solid option.

I'm not sure I understood 1, though 😅

bruno-fs: got it

bruno-fs: I wrote a small test to confirm the behavior I was expecting is actually happening at class construction (should've written it in the first place).

Test is passing
```python
def test_multiple_normalizers():
    """Ensure one normalizer class does not leak info to another."""
    assert BaseNormalizer.fields == {}

    class N1(BaseNormalizer):
        fact1 = FactMapper("some_raw_fact", str)

    class N2(BaseNormalizer):
        fact2 = FactMapper("another_raw_fact", str)

    assert BaseNormalizer.fields == {}
    assert N1.fields == {"fact1": N1.fact1}
    assert N2.fields == {"fact2": N2.fact2}
```

Will be a good benchmark while I experiment other approaches.

Great thing about getting rid of the metaclass is making BaseNormalizer an abstract class 😬

quipucords/scanner/normalizer.py
```diff
@@ -0,0 +1,226 @@
...
+    """
+
+    def __init__(
+        self,
+        raw_fact_key: str | list[str] | None,
+        normalizer: callable,
+        *,
+        validators=None,
+        dependencies: list[str] = None,
+    ) -> None:
```
bruno-fs, line 155: normalizer should probably be called something else here

quipucords/scanner/normalizer.py
```diff
@@ -0,0 +1,229 @@
...
+            # either because this method don't have access to it (we are dealing with
+            # class construction here).
+            # As odd as it seems to create something for another class here, given the
+            # symbiotic relationship between Normalizers and FactMappers, this seems to
+            # be the sweetspot for it and a good trade off.
+            normalizer.fields = {}
+        # ensure dependencies are created BEFORE this one
+        self._ensure_dependencies_exist(normalizer)
+        normalizer.fields[name] = self
+        self.fact_name = name
```
bruno-fs, line 202: how about this @mirekdlugosz ? got rid of the metaclass, but I still feel the wall of text might be helpful for future readers

quipucords/scanner/normalizer.py
bruno-fs: https://github.com/quipucords/quipucords/pull/2492/files#r1348003229

quipucords/scanner/normalizer.py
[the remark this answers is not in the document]
bruno-fs: done =D

quipucords/scanner/normalizer.py
infinitewarp: My gut instinct tells me `None` might be more appropriate than `NotImplemented`, but I've only very rarely actually used `NotImplemented`, and I don't know what side effects I may be missing from using one or the other here.

Otherwise I think I like this idea, and I am grateful for the wall of text comment!

bruno-fs: I never used not `NotImplemented` before but it felt adequate to this use case. Lemme check what happens if you don't add a fact mapper and try to call `.normalize`

bruno-fs: mmmm 

```
AttributeError                            Traceback (most recent call last)
Cell In[5], line 1
----> 1 n.normalize()

File ~/src/quipucords/quipucords/scanner/normalizer.py:60, in BaseNormalizer.normalize(self)
     58     return self._normalized
     59 self._normalized = {}
---> 60 for fact_name in self.fields.keys():
     61     norm_result = self._normalize_fact(fact_name)
     62     self._normalized[fact_name] = norm_result.value

AttributeError: 'NotImplementedType' object has no attribute 'keys'

```

bruno-fs: IDK, I'm on the fence. If u really prefer `None` I can gladly change it.

bruno-fs: replaced `NotImplemented` with `None`

infinitewarp (approved):
I suspect I won't _fully_ grok the scope of potential problems with this new design until I'm writing _my own_ normalizers and fact mappers, but at least in the isolated context of code review, this seems like a good approach. :shipit:

quipucords/tests/scanner/test_normalizer.py
```diff
@@ -0,0 +1,199 @@
...
+
+        norm = N(raw_facts, mocker.ANY)
+        norm.normalize()
+        assert norm.facts == {"fact": 5}
+        assert (
+            caplog.messages[-1] == "Unexpected raw facts used for fact 'fact': {'raw3'}"
+        )
+
+
+def test_multiple_normalizers():
```
mirekdlugosz, line 187: Would it be worth it to add some more cases in this vein?

- two normalizers, but with the same fact name
- one normalizer claiming a fact dependency on a fact that is only defined in another normalizer
- two normalizers inheriting from common parent, but other than BaseNormalizer (e.g. BaseNormalizer -> NetworkNormalizer -> N1 ; BaseNormalizer -> NetworkNormalizer -> N2). Assuming we expect to ever have normalizers inheriting indirectly from BaseNormalizer.

mirekdlugosz (approved):

quipucords/scanner/normalizer.py
mirekdlugosz: Late to the party, but:

- agreed that `None` is better than `NotImplemented`, which has pretty specific meaning for Python internals.
- nothing wrong with wall of text. We deal with relatively complex problem to begin with, it won't hurt to have a little bit of prose to better understand the rationale and possible pitfalls.

quipucords/scanner/normalizer.py
```diff
@@ -0,0 +1,228 @@
...
+        except Exception:
+            logger.exception("Unexpected error during '%s' normalization.", fact_name)
+            return NormalizedResult(None, has_error=True, raw_fact_keys=None)
+        if not isinstance(norm_result, NormalizedResult):
+            raw_fact_keys = self._get_raw_fact_keys(fact_name)
+            norm_result = NormalizedResult(
+                value=norm_result, raw_fact_keys=raw_fact_keys
+            )
+        else:
+            self._check_raw_fact_keys_for_fact(fact_name, norm_result.raw_fact_keys)
```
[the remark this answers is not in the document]
bruno-fs: > This will happen when I claim that fact_foo depends on fact_bar, but fact_bar is not defined, correct? 

Yes

> So this is really developer-facing. And as a developer, I would prefer this thing to blow up loudly and scream at me that I've got my dependencies defined wrong.

This is a really good argument. Gonna change it here.

infinitewarp (approved):

PR #2497, opened 2023-10-04
chore: addopt bandit static linter

bruno-fs:
Yet another linter enabled in ruff. This time, [bandit](https://pypi.org/project/bandit/).

The inspiration to add this check came from this comment: https://github.com/quipucords/quipucords/pull/2492#discussion_r1345776403

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,6 +88,7 @@ select = [
     "PL",  # Pylint
     "BLE", # flake8-blind-except
     "PTH", # flake8-use-pathlib
+    "S",   # flake8-bandit
 ]
 ignore = [
     "D107",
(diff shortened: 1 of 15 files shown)


infinitewarp (approved):
big 👏👏👏 for removing `assert`s from runtime code

pyproject.toml
```diff
@@ -112,9 +113,14 @@ known-first-party = [
 
 [tool.ruff.per-file-ignores]
 "__init__.py" = ["F401", "D104"]
-"**/test_*.py" = [
+"**test**" = [
```
abellotti, line 116: seems to be much more relaxed, what weren't we catching with **/test_*.py ?

bruno-fs: I changed to also capture `conftest.py`.

bruno-fs:
> big 👏👏👏 for removing asserts from runtime code

@infinitewarp thank @mirekdlugosz for calling it out in another PR ;D

pyproject.toml
abellotti: Can we do this in Ruff,   "**/test_*.py", "conftest.py" = [ .... ] so we don't duplicate the exceptions?

bruno-fs: I did try something similar and now that. no luck. Will check the docs tho, maybe there's a way 🤔

PR #2498, opened 2023-10-09
fix: `export LANG=C LC_ALL=C;` before all ansible raw commands

infinitewarp:
This should help make commands use a consistent locale since we can't trust the system default and we can't reliably set locale on the SSH connection.

Relates to JIRA: DISCOVERY-244
https://issues.redhat.com/browse/DISCOVERY-244

---

It feels ugly, but I confirmed locally it works. 🤷 

Note that I'm using `export` as a separate command instead of prefixing individual program commands with variables. I chose this approach because in _some_ of our `raw` commands, we execute _multiple_ programs, and each of them might need their own copy of these variables. By simply exporting to the environment at the start of each `raw`, we cover _all_ of its commands, and we have a single consistent pattern applied everywhere.

How did I test this? I started a RHEL9 EC2 instance, and I performed the following changes to it before starting a network scan. I think this is all the relevant commands I ran; this was a messy manual process as I was experimenting to make a "badly configured" target system.

```
sudo dnf install glibc-langpack-es
sudo localectl set-locale LANG=es_US
sudo dnf remove glibc-langpack-en
sudo vim /etc/profile.d/lang.sh
:%s/C.UTF-8/es_US/
:%s/en_US.UTF-8/es_US/
:%s/en_US.UTF-8/es_US/
:x
sudo reboot
```

After the VM restarted, I SSHed back into it and confirmed that `es_US` was the default locale.
```
$ id -u jboss
id: �jboss�: no existe ese usuario
```

I also checked `/etc/ssh/sshd_config` to see if connections were allowed to set `LANG` or `LC_*` variables, and it did not. You should check your `/etc/ssh/sshd_config`, remove or comment out anything like `AcceptEnv LANG LC_*`, then `sudo service sshd restart`, and reconnect any SSH sessions.

### DIFF
--- a/quipucords/scanner/network/runner/roles/check_dependencies/tasks/main.yml
+++ b/quipucords/scanner/network/runner/roles/check_dependencies/tasks/main.yml
@@ -5,7 +5,7 @@
     internal_host_started_processing_role: "check_dependencies"
 
 - name: test if user has sudo cmd
-  raw: echo "user has sudo" 2>/dev/null
+  raw: export LANG=C LC_ALL=C; echo "user has sudo" 2>/dev/null
   register: internal_user_has_sudo_cmd
   become: yes
   ignore_errors: yes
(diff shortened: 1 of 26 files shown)


bruno-fs (approved):
Looks great!

abellotti (approved):

nicolearagao (approved):

PR #2506, opened 2023-10-25
Improve acs connectivity errors

nicolearagao:


quipucords/scanner/acs/connect.py
```diff
@@ -48,11 +50,7 @@ def execute_task(self, manager_interrupt):
                     self.system_name,
                 )
         except RequestException as exception:
-            logger.exception("Unable to connect to '%s'.", self.system_name)
-            if isinstance(exception, RequestConnError):
-                conn_result = SystemConnectionResult.UNREACHABLE
-            else:
-                conn_result = SystemConnectionResult.FAILED
+            conn_result = self._handle_connection_error(exception)
```
mirekdlugosz, line 53: Just a suggestion:

```python
        try:
            response = self.client.get(
            ... # snip
        except (MaxRetryError, RetryError, RequestConnError):
            logger.exception(
                """Connection error while connecting to '%s'.
                Verify source information and try again.""",
                self.system_name,
            )
            conn_result = SystemConnectionResult.UNREACHABLE
        except RequestException:
            logger.exception(
                """Unable to connect to '%s'.
                Unexpected exception while handling connection.""",
                self.system_name,
            )
            conn_result = SystemConnectionResult.FAILED
```

You can put multiple `except` blocks for single `try`. If more specific exceptions are put first, they are handled and Python will skip further `except`s. If some generic exception is raised, it will fall through specific `except`s and get caught by more generic one.

nicolearagao: @mirekdlugosz I like the idea, but I was afraid it would get too polluted, but I guess not enough to need a separate function.

mirekdlugosz: I hear you. That's why it's only a suggestion. I would probably consider extracting `try` body into separate function, and leave exception handling here, but that's just my preference.

Let's see what other people have to say.

infinitewarp: I like @mirekdlugosz 's suggestion of keeping the multiple excepts here.

mirekdlugosz (approved):

bruno-fs (approved):

PR #2510, opened 2023-11-02
Update ruff to 0.1.3

abellotti:
chore: Updating ruff to 0.1.3
    
- Updating to latest ruff and keeping consistency with qpc
- Latest ruff 0.1.3 was not as forgiving for a couple of syntaxes as the earlier version. So, fixing those errors here which also helped eliminate a couple of noqa and lint exceptions.


infinitewarp (approved):
Good to condense those `else`+`if`s into `elif`s anyway. LGTM! 🚀

PR #2533, opened 2023-12-04
fix(source-create): prevent a race condition on triggered scans

bruno-fs:
Relates to JIRA: DISCOVERY-372


infinitewarp (approved):
LGTM! :shipit: 

And thank you for optimizing that unnecessary nested conditional logic. 🧠

PR #2535, opened 2023-12-06
Add required params to Jenkinsfile

mirekdlugosz:
The new pipeline skips execution if you didn't select any of API, CLI and UI. This is to not waste too much time in badly-configured standalone jobs.

But these params aren't set in PR-triggered runs, either, causing all PR pipelines to quickly finish; what's worse - they would appear as if they were successful.

Let's see if I put this directive in right place.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,13 @@
 @Library("smqe-shared-lib@master") _
 
+properties([
+    parameters([
+        booleanParam(defaultValue: true, name: 'TEST_API'),
+        booleanParam(defaultValue: true, name: 'TEST_CLI'),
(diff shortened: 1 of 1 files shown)


mirekdlugosz:
Alright, Jenkins is using `Jenkinsfile` from target branch / main. Which makes sense - we don't want to give full control over Jenkins to random people who happen to create PRs.

But that also means I can't test this before merge, and if I'm wrong, I would need X PRs until this is finally fixed.

So, instead, I did a fix in discovery-ci repo - now it detects if it's running against PR and fills Camayoc components based on that. Later, we will need to change values depending on OS anyway.

PR #2536, opened 2023-12-07
Feature/add last updated to credentials model

nicolearagao:


quipucords/tests/api/credential/test_credential.py
```diff
@@ -1122,7 +1122,13 @@ def test_masked_data_serialization_list(self, django_client):
         """Test if data is masked as expected for list method."""
         results = []
         for input_data, output, _ in self.INPUT_OUTPUT_ID:
-            CredentialFactory(**input_data)
+            credential = CredentialFactory(**input_data)
+            output["created_at"] = credential.created_at.isoformat()
+            output["updated_at"] = (
+                credential.updated_at.isoformat()
+                if hasattr(credential, "updated_at")
```
infinitewarp, line 1129: When could `credential.updated_at` not exist as an attribute? 

Is this just a weird side effect of using the factory? I think "real" credentials should always have this value.

nicolearagao: I guess u are correct. This field will be automatically updated whenever a change is made to the model instance. I will correct this.

quipucords/api/migrations/0044_credential_created_at_credential_updated_at.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def update_timestamps(apps, schema_editor):
+    """Populate new fields of the existing rows."""
+    Credential = apps.get_model("api", "Credential")
+
+    for credential in Credential.objects.all():
+        credential.created_at = django.utils.timezone.now()
+        credential.updated_at = django.utils.timezone.now()
+        credential.save()
```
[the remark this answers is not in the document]
infinitewarp: Just a quick test in a local REPL to confirm this general syntax works like I think it does:

```
In [2]: Credential.objects.all().update(become_method='sudo')
Out[2]: 4
```

(There are 4 credential obbjects in my DB.)

infinitewarp (approved):
🎉 👍

PR #2537, opened 2023-12-13
chore: sort environment variables when logging.

abellotti:
- When user selects to dump the environment variables via the boolean QPC_LOG_ALL_ENV_VARS_AT_STARTUP at start-up, let's sort them as it is tedious to find a variable by name otherwise.


infinitewarp (approved):
Excellent idea!

ruda (approved):
LGTM.

PR #2540, opened 2024-01-09
feat: Add support for bulk deleting credentials

abellotti:
- Adding support for deleting credentials in bulk via the /api/v1/credentials/bulk_delete/ POST URL. That URL takes the list if ids to delete via the ids: [ id1, id2, ... ] entry in the posted request.

- [x] Needs an updated openapi schema
- [x] Needs Unit tests

Related to: https://issues.redhat.com/browse/DISCOVERY-511

### DIFF
--- a/quipucords/api/credential/view.py
+++ b/quipucords/api/credential/view.py
@@ -20,18 +20,12 @@ from rest_framework.serializers import ValidationError
 from rest_framework.viewsets import ModelViewSet
 
 from api import messages
-from api.common.util import is_int
+from api.common.util import ids_to_str, is_int
 from api.exceptions import UnprocessableEntity
 from api.filters import ListFilter
 from api.models import Credential, Source
 from api.serializers import CredentialSerializer
 from api.user.authentication import QuipucordsExpiringTokenAuthentication
-
-
-def related_sources(pk):
-    """Return sources related to a credential."""
-    return Source.objects.filter(credentials__pk=pk).values("id", "name")
-
 
 auth_classes = (QuipucordsExpiringTokenAuthentication, SessionAuthentication)
 perm_classes = (IsAuthenticated,)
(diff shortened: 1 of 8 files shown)


abellotti:
Usage:

```
POST /api/v1/credentials/bulk_delete/
{
  "ids": [
       67,
       68,
       123,
       129,
       ...
   ]
}
```

quipucords/api/credential/view.py
```diff
@@ -15,12 +21,56 @@
...
+    if not ids:
+        raise ParseError(detail=_("Missing list of credentials to delete 'ids'"))
+
+    with transaction.atomic():
+        try:
+            for pk in ids:
+                cred = Credential.objects.get(pk=pk)
+                if related_sources(pk):
+                    raise ValidationError
+                cred.delete()
```
bruno-fs, line 63: I think there's room for improvement here because this has potential to create too many queries

1. start with a query.filter grabbing all Sources that are related to the credential id list. If the result is not an empty queryset, throw the error for CRED_ID_DELETE_NOT_VALID_W_SOURCES (the exact credential IDs can be checked on the resulting source list)

2. `qs = Credential.objects.filter(pk__in=ids)`. If qs.count() <  len(ids), then we throw CRED_ID_DOES_NOT_EXIST (compare qs.all() pks with ids to find the offending ones

3. if we didn't throw errors in the previous steps we can safely run `qs.delete()` 

Done, instead of 3 queries per credential we end up with 3 queries total (if no error is found).
Thoughts?

abellotti: 👍 That sounds good, I initially had the qs.delete() but it was always deleting credentials even  if sources were related, then I realized we do our own check in the destroy method (per DELETE /api/v1/credentials/##) thus this variant.  I will attempt the above. Right now the first failure (404 or related source 422) is the credential id we return in the error, as we don't attempt to proceed.  With the above, we should be able to return the list of all 404 (or 422) credential Ids.

PR #2544, opened 2024-01-11
DISCOVERY-340: ensure fingerprint-only scan jobs are properly working

bruno-fs:
Adds test coverage to fingerprint-only scan jobs as a pre-requisite for the upcoming refactors for upload & merge views.

Relates to JIRA: DISCOVERY-340

### DIFF
--- a/quipucords/fingerprinter/runner.py
+++ b/quipucords/fingerprinter/runner.py
@@ -508,8 +508,8 @@ class FingerprintTaskRunner(ScanTaskRunner):
                 fingerprint = self.process_facts_for_datasource(
                     source_type, source, fact
                 )
-            except KeyError:
-                self.scan_task.log_message.error(
+            except NotImplementedError:
+                self.scan_task.log_message(
                     "Could not process source, " f"unknown source type: {source_type}",
                     log_level=logging.ERROR,
                 )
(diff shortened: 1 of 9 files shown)


quipucords/fingerprinter/runner.py
```diff
@@ -508,8 +508,8 @@ def _process_source(self, source):
                 fingerprint = self.process_facts_for_datasource(
                     source_type, source, fact
                 )
-            except KeyError:
-                self.scan_task.log_message.error(
+            except NotImplementedError:
```
infinitewarp, line 511: Where did the `KeyError` transform into `NotImplementedError`? Or was this already wrong before? Did this result from changing `serializer.is_valid(raise_exception)`?

infinitewarp (approved):
LGTM, I think! It feels like an eternity since I last touched this Celery code, but I believe what I see here looks okay.

quipucords/fingerprinter/runner.py
bruno-fs: It was already wrong before. I should have saved this change in a separate commit.

here's the reasoning
- [this place](https://github.com/quipucords/quipucords/blob/10c64d08060ebace0a483b2b1b22980248f1b1e4/quipucords/fingerprinter/runner.py#L482-L487) is where the exception is thrown.
- [15 months ago](https://github.com/quipucords/quipucords/commit/dd705ba265fe0c93457c85541a7d4fc6a1968102) this was raising a `KeyError` instead of `NotImplementedError`.

PR #2547, opened 2024-01-15
feat(rhacs): use new RHACS v4.3+ API paths

infinitewarp:
Updated 2024-01-17: When fetching RHACS data, we now use the new API paths as introduced in RHACS 4.3. We expect requests to the older RHACS 4.2 API to fail.

~When fetching RHACS data, we now try the "new" API paths as introduced in RHACS 4.3. If a request to the new 4.3 API path fails, we seamlessly retry the request with the old RHACS 4.2 path.~

~This may be a good solution while 4.2 is still available and in use, but we should consider removing this retry logic and embrace the 4.3+ APIs exclusively after some amount of burn-in or if we learn that we have minimal or no customers using RHACS 4.2.~

Relates to JIRA: DISCOVERY-502
https://issues.redhat.com/browse/DISCOVERY-502


nicolearagao (approved):
LGTM

abellotti (approved):

PR #2549, opened 2024-01-18
feat(vcenter): Add support for installed_products.

abellotti:
Added support for installed_products for discovered vCenter VMs.

- Added the update-product-list directive to the Makefile
    The new command fetches the product file list and creates the needed JSON subset docs/products-prod.json for consumption by Discovery.
- Added utility code to query products and product names from the products-prod.json file.
- Added generation of the installed_products fact with the "id" and "name" for the operating system name reported with the vm.os property.

Relates To: https://issues.redhat.com/browse/DISCOVERY-316

Todo's:

- [ ] Verify uploaded Insights report shows the installed products for vCenter
- [ ] Investigate auto-update during builds
- [ ] Add tests

PR #2551, opened 2024-01-19
test(vcenter): Converting vCenter test to pytest

abellotti:
- Converting the vCenter test to pytest.

This is needed in https://github.com/quipucords/quipucords/pull/2549 and done in a separate PR to minimize its scope.


bruno-fs (approved):
LGTM

PR #2554, opened 2024-01-24
tests: pytestfy quipucords/tests/api/scanjob/test_scanjob.py

bruno-fs:
also fix a inconsistency in ScanFactory where related jobs were not fully associated to Scan instances.


nicolearagao (approved):
niiiice

PR #2558, opened 2024-01-25
tests: Updated tests with hardcoded /api/v1 paths to use reverse()

abellotti:
- Updated tests with hard coded /api/v1 paths to use django's reverse() function.
- Updated tests using test constants referencing /api/v1 paths.

### DIFF
--- a/quipucords/api/urls.py
+++ b/quipucords/api/urls.py
@@ -33,9 +33,11 @@ ROUTER.register(r"users", UserViewSet, basename="users")
 v1_urls = [
     path("credentials/bulk_delete/", credential_bulk_delete, name="cred-bulk-delete"),
     path("reports/<int:report_id>/details/", details, name="reports-details"),
-    path("reports/<int:report_id>/deployments/", deployments),
-    path("reports/<int:report_id>/insights/", insights),
-    path("reports/<int:report_id>/", reports),
+    path(
+        "reports/<int:report_id>/deployments/", deployments, name="reports-deployments"
+    ),
+    path("reports/<int:report_id>/insights/", insights, name="reports-insights"),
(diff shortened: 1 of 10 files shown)


quipucords/api/urls.py
```diff
@@ -31,13 +31,19 @@
...
 
 v1_urls = [
-    path("credentials/bulk_delete/", credential_bulk_delete),
-    path("reports/<int:report_id>/details/", details),
-    path("reports/<int:report_id>/deployments/", deployments),
-    path("reports/<int:report_id>/insights/", insights),
-    path("reports/<int:report_id>/", reports),
-    path("reports/merge/jobs/", async_merge_reports),
-    path("reports/merge/jobs/<int:scan_job_id>/", async_merge_reports),
+    path("credentials/bulk_delete/", credential_bulk_delete, name="cred-bulk-delete"),
```
infinitewarp, line 34: Unless there's a technical reason it needs to be shortened (bumping into max length somewhere?), I think we should keep the full `credentials-bulk-delete` as the name for clarity and consistency.

infinitewarp: Ooohhh, good point! I didn't see how the existing `cred` was already shortened. In that case, while you're touching so many lines, yeah, let's also rename `cred` to `credentials` to make it more consistent with the rest.

(There's still some inconsistency in how we do or don't pluralize the other nouns, but it's up to you if you might want to touch all of them. I don't have a super strong opinion there.)

abellotti: updated to credentials-* for basename prefixes.

infinitewarp (approved):
🚢 🇮🇹 !

PR #2559, opened 2024-01-25
fix: pin pip==23.3.1 due to 23.3.2 introducting unwanted setuptools-scm version

infinitewarp:
We need to pin `pip` to `23.3.1` because `23.3.2` (specifically https://github.com/pypa/pip/pull/12392/commits/9ec7e36dd6e1ef80c80c9fefde69bf97631f51c9) causes our `requirements-build.txt` to define `setuptools-scm`'s version using a URL instead of its semver string, and downstream Cachito cannot handle that URL.

This is a temporary crutch, but it means we don't have to manually revert/edit the version for `setuptools-scm` like I did in https://github.com/quipucords/quipucords/pull/2556.

See also these Slack discussions for context:

* https://redhat-internal.slack.com/archives/C02QSNF1UKE/p1706121152928859?thread_ts=1706120564.649749&cid=C02QSNF1UKE
* https://redhat-internal.slack.com/archives/C02QSNF1UKE/p1706211614206739


pyproject.toml
```diff
@@ -34,6 +34,7 @@ more-itertools = "^9.1.0"
 django-cors-headers = "^4.3.1"
 
 [tool.poetry.group.dev.dependencies]
+pip = "23.3.1"  # TODO BWS Add explanation
```
infinitewarp, line 37: Yes, I know. I have a local commit that I have not yet fixed that will update this, but I don't want to push it yet because I'm waiting on brew to build downstream, and I don't want to risk changing the upstream refs before it finishes.

It will instead look like this:
```
pip = "23.3.1"  # see https://github.com/quipucords/quipucords/pull/2559
```

abellotti (approved):

PR #2562, opened 2024-01-29
chore(migrations): make credentials timestamp attribute migration reversible

abellotti:
- Adding the reverse_code to the migration, while we would lose the *_at attributes added since those were added and we don't have those before the 0044 migration, we can at least reverse the migration.


bruno-fs (approved):
<3

PR #2563, opened 2024-01-30
fix(network-scan): don't overwrite ansible logs

bruno-fs:
scans with multiple "inspection groups" were saving only the latest inspection group data and hence impairing our ability to debug it.


quipucords/quipucords/settings.py
```diff
@@ -360,7 +360,7 @@ def app_secret_key_and_path():
 # Reverse default behavior to avoid host key checking
 os.environ.setdefault("ANSIBLE_HOST_KEY_CHECKING", "False")
 # Reverse default behavior for better readability in log files
-os.environ.setdefault("ANSIBLE_NOCOLOR", "False")
+os.environ.setdefault("ANSIBLE_NOCOLOR", "True")
```
infinitewarp, line 363: Should this change the preceding comment too?

infinitewarp (approved):

quipucords/quipucords/settings.py
bruno-fs: actually, the default behavior is having it set to False
When I added this I while ago I set it to False by mistake

abellotti (approved):

infinitewarp:
We're seeing one of our CI test machines fail unexpectedly on this PR check, but that failure seems unrelated to this change. Ignoring it and merging!

PR #2569, opened 2024-02-02
refactor(network-scan): decouple InspectCallback logic

bruno-fs:
Decouple logic saving data to database and postprocessing from
InspectCallback.

Now inspect callback only acts as a helper to parse and group together
facts collected from ansible events and mantains it's original
functionality for cancel callback. This will make understanding and
testing this class much, much easier.

InspectTaskRunner now holds the logic for post processing and storing results to db.

### DIFF
--- a/quipucords/scanner/network/inspect.py
+++ b/quipucords/scanner/network/inspect.py
@@ -5,9 +5,11 @@ from pathlib import Path
 import ansible_runner
 from ansible_runner.exceptions import AnsibleRunnerException
 from django.conf import settings
+from django.db import transaction
 from django.forms import model_to_dict
 
 import log_messages
+from api.inspectresult.model import RawFact
 from api.models import (
     ScanJob,
     ScanOptions,
@@ -19,11 +21,13 @@ from api.vault import write_to_yaml
 from constants import SCAN_JOB_LOG
 from scanner.exceptions import ScanFailureError
 from scanner.network.exceptions import ScannerException
-from scanner.network.inspect_callback import InspectResultCallback
+from scanner.network.inspect_callback import AnsibleResults, InspectCallback
+from scanner.network.processing.process import NO_DATA, process
 from scanner.network.utils import (
     check_manager_interrupt,
     construct_inventory,
     delete_ssh_keyfiles,
+    raw_facts_template,
 )
 from scanner.runner import ScanTaskRunner
 
@@ -204,7 +208,7 @@ class InspectTaskRunner(ScanTaskRunner):
                 f" {(idx + 1):d} of {len(group_names):d}"
             )
             self.scan_task.log_message(log_message)
-            call = InspectResultCallback(self.scan_task, manager_interrupt)
+            call = InspectCallback(manager_interrupt)
 
             # Build Ansible Runner Parameters
             runner_settings = {
@@ -258,6 +262,9 @@ class InspectTaskRunner(ScanTaskRunner):
 
             # Let's delete any private ssh key files that we generated
             delete_ssh_keyfiles(inventory)
+            # persist facts
+            for result in call.iter_results():
+                self._persist_results(result)
             # save stdout and stderr from ansible
             self._persist_ansible_logs(runner_obj)
 
@@ -283,10 +290,9 @@ class InspectTaskRunner(ScanTaskRunner):
                     error_msg = log_messages.NETWORK_TIMEOUT_ERR
                 else:
                     error_msg = log_messages.NETWORK_UNKNOWN_ERR
+                # TODO: refactor this - this logic is incorrect. The result of
+                # the whole scan is only taking into consideration the last group.
                 scan_result = ScanTask.FAILED
-
-            # Always run this as our scans are more tolerant of errors
-            call.finalize_failed_hosts()
 
             log_message = (
                 "INSPECT PROCESSING GROUP COMPLETED"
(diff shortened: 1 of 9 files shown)


quipucords/scanner/network/inspect.py
```diff
@@ -297,6 +303,89 @@ def _inspect_scan(  # noqa: PLR0912, PLR0915, C901
...
+                    fact_value=fact_value,
+                    host=ansible_results.host,
+                )
+            except Exception:
+                logger.exception(
+                    "[host=%s] Unexpected error ocurred during fact '%s' processing",
+                    ansible_results.host,
+                    fact_key,
+                )
+                continue
```
bruno-fs, line 356: in the original implementation this code wasn't wrapped in try/except.

this is probably overkill, but I'm still unsure on how well tested and resilient `process` is

quipucords/scanner/network/inspect_callback.py
```diff
@@ -1,239 +1,298 @@
...
-        event_data = event_dict.get("event_data")
-        host = event_data.get("host", UNKNOWN)
-        result = event_data.get("res")
-        task_action = event_data.get("task_action")
+    def task_on_ok(self, event_dict: dict):
+        """Handle successful ansible events."""
+        # example event
+        # {
+        #   "uuid": "9d097760-4d2c-43e5-8aba-d1a0e825aed7",
+        #   "counter": 140,
```
bruno-fs, line 58: I failed miserably in finding documentation for those events, so I decided to slap full examples here.
they will be used later (with way less data) in unit tests for this class

quipucords/scanner/network/inspect_callback.py
```diff
@@ -1,239 +1,298 @@
...
-        task_facts = result.get("ansible_facts")
-        if task_facts:
-            self.process_task_facts(task_facts, host)
+        role = event_data.get("role")
+        logger.info("[host=%s] processing role=%s task=%s", host, role, task)
+        self._process_task_facts(ansible_facts, host)
+
+    def _process_task_facts(self, ansible_facts, host):
+        for fact_name, _fact_value in ansible_facts.items():
+            fact_value = sanitize_for_utf8_compatibility(_fact_value)
```
bruno-fs, line 108: this is change of order compared with the original implementation. sanitization was happening after post processing.

I decided to do that as early as possible to be more cautious.

quipucords/scanner/network/inspect_callback.py
```diff
@@ -1,239 +1,298 @@
...
-            raise AnsibleRunnerException(err_msg) from err_msg
+            event = event_dict["event"]
+        except KeyError:
+            logger.error(
+                log_messages.TASK_UNEXPECTED_FAILURE,
+                "event_callback",
+                "unknown event",
+                event_dict,
+            )
+            return
```
bruno-fs, line 275: from the few logs we got from the field since this was added I think we could either remove this log or make it debug level.

the events from this issue are pretty uninteresting and seem to be common since the upgrade to ansible-core 8

quipucords/scanner/network/inspect_callback.py
```diff
@@ -1,239 +1,298 @@
...
-                            log_messages.TASK_UNEXPECTED_FAILURE
-                            % ("event_callback", event, event_dict),
-                            log_level=logging.ERROR,
-                        )
-                # Save last role for task logging later
-                if event == "playbook_on_task_start" and event_data:
-                    event_role = event_data.get("role")
-                    if event_role != self.last_role:
-                        self.last_role = event_role
-        except Exception as err_msg:  # noqa: BLE001
```
bruno-fs, line 222: removed this generic try/except because the one wrapping `ansible_runner.run` in `InspectTaskRunner` already covers something unexpected here (and I would rather avoid having these blind excepts everywhere)

quipucords/scanner/network/inspect_callback.py
infinitewarp: In this case, I'd encourage including long example objects like this one _in full_ in the unit tests (maybe with a comment to explain how you captured them). Although we might think we need less data for the tests, I often look at tests as another form of documentation, _especially_ when it comes to dealing with external/third-party integration code, and these fuller examples might contain clues for future ideas.

quipucords/scanner/network/inspect_callback.py
```diff
@@ -1,239 +1,298 @@
...
+        task = event_data.get("task")
+        role = event_data.get("role")
         if result.get("rc") == TIMEOUT_RC:
-            logger.warning("Task %s timed out", event_data["task"])
+            logger.error("[host=%s] Task %s timed out", host, event_data.get("task"))
 
         # Only log an error when ignore errors is false
         if event_dict.get("ignore_errors", False):
             err_reason = result.get("msg", event_data)
+            # is it unexpected, tho? we have so many commands that will recurrently fail
```
infinitewarp, line 184: Agreed… maybe the message and log level should be `warning` instead. Maybe? I'm not 100% sure either. 🤔

bruno-fs: changed it to warning

quipucords/tests/scanner/network/test_inspect_callback.py
```diff
@@ -0,0 +1,299 @@
...
+        extravars=ScanOptions.get_default_extra_vars(),
+    )
+    assert runner_obj.status == "successful"
+    all_results = list(callback.iter_results())
+    assert len(all_results) == 1
+    results = all_results[0]
+    assert results.host == "localhost"
+    assert results.status == SystemInspectionResult.SUCCESS
+    # spot check at least one fact
+    assert results.facts["cpu_count"] == str(os.cpu_count())
```
bruno-fs, line 299: given a recent feedback, I expect this to work on mac, but I'd like someone to please run it to be sure @infinitewarp or @abellotti

quipucords/scanner/network/inspect_callback.py
bruno-fs: I moved all these comments to actual fixtures with instructions on how to run obtain them

quipucords/tests/scanner/network/test_inspect_callback.py
infinitewarp: Unfortunately, this test (but _only_ this test) is failing on my Mac. 😕 Details in Slack thread here: https://redhat-internal.slack.com/archives/C02QSNF1UKE/p1707410433413239?thread_ts=1707317786.877279&cid=C02QSNF1UKE

infinitewarp (approved):

PR #2572, opened 2024-02-05
refactor: Migrate SourceOptions to Source

abellotti:
- Migrating the SourceOptions attributes to Source
- Serialize the new attributes from Source
- Also Serialize the now "options" property to represent the SourceOptions attributes that were merged in for v1 compatibility.
- Expose both /api/v1/sources and /api/v2/sources endpoints
- Add V2 tests 

**Notes:**
- The docs/swagger.yml is /api/v1 specific, we don't want to clutter this but create a new OpenAPI 3.0 for the v2 API's, i.e. `docs/openapi-3-v2.json`
- The Migration is currently forward only, while we can reverse the migration, it's a no-op, the One-To-One relationships to SourceOptions are not recreated essentially losing the ssl options. https://github.com/quipucords/quipucords/blob/b47f64802c802a8580fadfc49aefa31afd5285df/quipucords/api/migrations/0045_merge_source_options_with_source.py

Relates to: DISCOVERY-314


abellotti:
With this PR, the `SourceOptions` SSL type attributes are migrated to `Source` and the `option` representation is preserved for these attributes for v1 compatibility.

Both /api/v1/sources and /api/v2/sources are now available and note the differences  in the following example for the _use_paramiko_ attribute:


GET /api/v1/sources/15

```
{
  "id": 15,
  "name": "aabtest3",
  "source_type": "network",
  "port": 22,
  "hosts": [
    "192.169.110.4",
    "193.42.21.5",
    "105.123.29.6",
    "105.123.28.9"
  ],
  "options": {
    "use_paramiko": false
  },
  "credentials": [
    {
      "id": 89,
      "name": "aabtest3"
    }
  ]
}
```

GET /api/v2/sources/15

```
{
  "id": 15,
  "name": "aabtest3",
  "source_type": "network",
  "port": 22,
  "hosts": [
    "192.169.110.4",
    "193.42.21.5",
    "105.123.29.6",
    "105.123.28.9"
  ],
  "use_paramiko": false,
  "credentials": [
    {
      "id": 89,
      "name": "aabtest3"
    }
  ]
}
```

Both driven by the newer model.

quipucords/api/source/model.py
```diff
@@ -52,6 +52,22 @@ class Source(models.Model):
...
+        if self.ssl_protocol is not None:
+            result_options["ssl_protocol"] = self.ssl_protocol
+        if self.ssl_cert_verify is not None:
+            result_options["ssl_cert_verify"] = self.ssl_cert_verify
+        if self.disable_ssl is not None:
+            result_options["disable_ssl"] = self.disable_ssl
+        if self.use_paramiko is not None:
+            result_options["use_paramiko"] = self.use_paramiko
+        # if result_options == {}:
+        #    return None
```
bruno-fs: I think we can get rid of this comment, right?
(this reminded me, I'm almost sure there's a checker on ruff for commented code)

bruno-fs: ah, nevermind. it was removed in a subsequent commit.

quipucords/api/source/serializer.py
```diff
@@ -50,6 +50,11 @@ def display_value(self, instance):
 class SourceSerializerBase(NotEmptySerializer):
```
bruno-fs, line 50: I'd argue `NotEmptySerializer` adds a nasty behavior we might not want on API v2. Do we need automatic removal of null fields? Lets keep this behavior on v1 for sure, for compatibility, but I'd get rid of it on v2 and embrace DRF standard.

Suggestion: use `NotEmptySerializerMixin` on SourceSerializerV1.

quipucords/api/source/serializer.py
```diff
@@ -606,3 +621,33 @@ class Meta:
...
+            instance.save()
+        return instance
+
+
+# Let's define the SourceSerializer based on the API_VERSION
+match API_VERSION:
+    case 2:
+        SourceSerializer = SourceSerializerV2
+    case _:
+        SourceSerializer = SourceSerializerV1
```
bruno-fs, line 653: I'm a bit confused here. Won't we use `request.version` to find the version at runtime? why do we need API_VERSION constant? 🤔

quipucords/tests/api/source/test_source.py
```diff
@@ -1710,3 +1710,1595 @@ def test_update_rhacs_green_path(self, django_client, rhacs_cred):
...
+            source_info = {"name": data["name"], "id": response["id"]}
+            source_list.append(source_info)
+        return source_list
+
+    def test_validate_opts(self):
+        """Test the validate_opts function."""
+        source_type = DataSources.SATELLITE
+        options = {"use_paramiko": True}
+        with pytest.raises(ValidationError):
+            SourceSerializer.validate_opts(options, source_type)
```
bruno-fs, line 1796: mmmm in this test case my confusion intensifies about the case for v1 vs v2. Isn't this `SourceSerializerV1`? 🤔

bruno-fs (commented):
Excellent PR. Such DRY, so many things clean.

I have some questions, though.

quipucords/tests/api/source/test_source.py
bruno-fs: if it is, shouldn't those tests live on the other `TestCase`, for serializer v1?

quipucords/api/source/serializer.py
```diff
@@ -74,7 +63,7 @@ class SourceSerializer(NotEmptySerializer):
     port = IntegerField(required=False, min_value=0, allow_null=True)
     hosts = JSONField(required=True)
     exclude_hosts = JSONField(required=False)
-    options = SourceOptionsSerializer(required=False, many=False)
+
     credentials = CredentialsField(many=True, queryset=Credential.objects.all())
```
bruno-fs, line 67: I have a PTSD everytime I see something that COULD be paginated in a single response.

Maybe we could (in subsequent PRs) force a limit on the backend to not allow more than an arbitrary number of credentials to be attached to a source. If not possible, then remove this field (sorry UI/CLI who will be making some extra requests).

quipucords/api/source/serializer.py
bruno-fs: gotcha. It's a nice way to have a smaller diff.

bruno-fs: leave it as is, :shipit:

quipucords/api/source/serializer.py
bruno-fs: I'm being admittedly paranoid with this one. we can deal with it later.

bruno-fs (approved):

PR #2574, opened 2024-02-06
fix: ignore unexpected fields in credential serializer

infinitewarp:
This simply removes the custom `CredentialSerializer.validate` method to restore standard DRF behaviors.

Relates to JIRA: DISCOVERY-534
https://issues.redhat.com/browse/DISCOVERY-534


bruno-fs (approved):
👍🏽

PR #2577, opened 2024-02-08
feat(Dockerfile): run server in container as non-root "discovery" user

infinitewarp:


Dockerfile
```diff
@@ -57,24 +59,31 @@ COPY deploy/ssl /etc/ssl/qpc
 # Create /deploy
 COPY deploy  /deploy
 
+# Copy server code
+COPY . .
+
+# Install quipucords as package
+RUN pip install -v -e .
```
bruno-fs, line 66: I'd rather have this layer as "external" / late as possible - to optimize build time for developers (me, hehe) rebuilding the image frequently.

I THINK this can happen right after `USER discovery`. If not, before `RUN chown ...`

bruno-fs (approved):

PR #2580, opened 2024-02-13
Port Scan API tests from Camayoc

mirekdlugosz:
Adding unit tests for things that we have in Camayoc.

### DIFF
--- a/quipucords/tests/api/scan/test_scan.py
+++ b/quipucords/tests/api/scan/test_scan.py
@@ -1,5 +1,6 @@
 """Test the API application."""
 
+import random
 import re
 from datetime import timedelta
 
@@ -227,6 +228,27 @@ class TestScanCreate:
             }
         }
 
+    @pytest.mark.parametrize("option_value", ("hamburger", "87", 42, "*", []))
+    def test_create_disable_optional_products_value_type(
+        self, django_client, faker, option_value
+    ):
+        """Test invalid type for disabled_optional_products value."""
+        source = SourceFactory()
+        payload = {
+            "name": faker.bothify("Scan ????-######"),
+            "sources": [source.id],
+            "options": {"disabled_optional_products": {"jboss_eap": option_value}},
+        }
+        response = django_client.post(reverse("v1:scan-list"), json=payload)
+        assert response.status_code == status.HTTP_400_BAD_REQUEST
+        assert response.json() == {
+            "options": {
+                "disabled_optional_products": {
+                    "jboss_eap": ["Must be a valid boolean."]
+                }
+            }
+        }
+
 
 @pytest.mark.django_db
 class TestScanRetrieve:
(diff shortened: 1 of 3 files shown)


quipucords/tests/api/reports/test_reports.py
```diff
@@ -290,3 +290,75 @@ def test_report_deployments_csv(django_client, deployment_with_logs: Deployments
...
+
+
+def test_report_details_consistency(
+    django_client, deployment_with_logs: DeploymentsReport
+):
+    """Test report details response is driven by request headers and not cached."""
+    deployments_report = deployment_with_logs
+    report = deployments_report.report
+
+    for _ in range(2):
```
infinitewarp, line 302: I don't understand why we are running this code twice. Both iterations use the same report IDs, and both assert the same results. How does this exercise or prove that the response is not cached, assuming that is the intent based on the docstring? Am I overlooking or misunderstanding something?

quipucords/tests/api/reports/test_reports.py
```diff
@@ -290,3 +290,75 @@ def test_report_deployments_csv(django_client, deployment_with_logs: Deployments
...
+
+
+def test_report_deployments_consistency(
+    django_client, deployment_with_logs: DeploymentsReport
+):
+    """Test report deployments response is driven by request headers and not cached."""
+    deployments_report = deployment_with_logs
+    report = deployments_report.report
+
+    for _ in range(2):
```
infinitewarp, line 339: Same question as previous note about running this code twice. Just putting a comment here in case anything needs to change.

quipucords/tests/api/reports/test_reports.py
mirekdlugosz: The Camayoc test for that traces to https://github.com/quipucords/camayoc/issues/195 from 2018, with following rationale:

> Test that you receive the appropriate content back when specifying a content type.
> Test that after you selected both content types that caching doesn't cause any side-effects and you get the appropriate content type back.

I wasn't around back then, so I can only assume this was done in response to actual bug, or a risk considered by a team at the time. And honestly, "check that our caching is not too aggressive" does sound reasonable, especially when it's cheap.

I was on the fence with that one myself, so I can remove this if you think it doesn't provide enough value to keep around.

mirekdlugosz: For caching, I think the idea was "if I request JSON and CSV, then I get JSON and CSV respectively". I.e. I don't get back cached CSV content if I request JSON after CSV. It's run twice to ensure cache is loaded (if any) and we cover both "request JSON and then CSV" and "request CSV and then JSON".

infinitewarp: I keep looking at this, and although I think I understand the historic context you provided, I don't think the current test provides much value. I think my vote would be simply to remove these tests, or at least remove the loop while keeping one-time assertions, assuming the basic behavior is not already covered elsewhere.

I don't feel _super_ strongly one way or the other, and I'll trust your judgement about keeping or removing them. You may resolve this conversation (and the other) whenever you're ready to merge.

infinitewarp (approved):

quipucords/tests/api/reports/test_reports.py
mirekdlugosz: I removed them in 91f870b128fbac2de83290ee124e0605f6dba0ef (with intent it will be squashed before merge). Since nobody has strong opinion and reason to keep them around, I think we can take a risk.

With recent discussion about changes we want to do around generating and storing reports, hopefully we will get better tests for that anyway.

I'll keep that thread open for a little longer, in case anyone wants to voice their opinion.

PR #2582, opened 2024-02-13
feat: lock out user after multiple failed logins

infinitewarp:
This change enables the `django-axes` middleware for logging failed login attempts and subsequently locking out the user from further login attempts for a specified length of time.

See also:
* https://django-axes.readthedocs.io/en/latest/index.html
* https://github.com/jazzband/django-axes
* https://pypi.org/project/django-axes/

The UI isn't super pretty now, but I believe we can iterate on that once we get this basic functionality in place. When using the discovery-ui web app, on your Nth failure, you get a simple page that looks like this:

Screenshot 2024-02-13 at 3 46 01 PM

Repeated failed logins from the `qpc` CLI will also trigger `django-axes` and its lockout functionality, but the CLI doesn't show any additional details like that. It simply says `Error: Unable to log in with provided credentials.` same as the first time you entered bad credentials. We may want to iterate on that as well.

`django-axes` already has APIs *and Django management commands* to reset its failure counters. We may consider writing some docs around them in case some sysadmin accidentally locks themself out and cannot wait for the cooldown to expire.

Those management commands include:
```
[axes]
    axes_list_attempts
    axes_reset
    axes_reset_failure_logs
    axes_reset_ip
    axes_reset_ip_username
    axes_reset_logs
    axes_reset_username
```


infinitewarp:
In the most recent commit, I exposed several of the `AXES_*` settings as environment variables, and if the user chooses to *disable* the account lockout functionality, we now log a warning at startup that looks something like this:

```
❯ AXES_LOCK_OUT_AT_FAILURE=False poetry run make server-migrate serve

...

Account login failure lockout is DISABLED (AXES_ENABLED=True, AXES_LOCK_OUT_AT_FAILURE=False). This may be a security risk!
```

Also, that warning message is a plain English string because the translation engine isn't ready yet when settings are being loaded here. Trying to invoke it would result in an exception like this:
```
django.core.exceptions.AppRegistryNotReady: The translation infrastructure cannot be initialized before the apps registry is ready. Check that you don't make non-lazy gettext calls at import time.
```
I *could* try to move this check and message later into the app's startup process, but… 🤷‍♂️

bruno-fs (approved):
I don't love the commented code on the second commit, but LGTM

infinitewarp:
> I don't love the commented code on the second commit, but LGTM

Agreed! That comment was informational text to readers unfamiliar with axes before I decided to add the environment variables. I have amended the second commit to remove the commented code.

PR #2583, opened 2024-02-19
refactor(Scan): Migrating ScanOptions into the Scan Model

abellotti:
Migrating the following Models into Scan and removing them from the Database:
   - ScanOptions
   - DisabledOptionalProductsOptions
   - ExtendedProductSearchOptions
    
The DisabledOptionalProductsOptions and ExtendedProductSearch get moved into JSONFields in the Scan model for easier management and extensibility.
    
The DisabledOptionalProductOptions gets migrated into a flipped boolean equivalent enabled_optional_products for better readability.

- Moving the ScanOptions attributes to the Scan Model
- Updated the ScanJob model to represent the options from its attached Scan and not its copy which no longer exists.

This PR keeps the v1 compatibility by exposing an options property with the _disabled_optional_products_ dictionary which inverts the boolean from the new model.

Relates To: https://issues.redhat.com/browse/DISCOVERY-314

- [x] Testing with API (manually)
- [x] Testing with UI (manually)
- [x] Test with Camayoc
- [x] Update pytests


quipucords/api/scanjob/serializer.py
```diff
@@ -109,7 +111,7 @@ class ScanJobSerializerV1(NotEmptySerializer):
     status = ValidStringChoiceField(read_only=True, choices=ScanTask.STATUS_CHOICES)
     status_message = CharField(read_only=True)
     tasks = TaskField(many=True, read_only=True)
-    options = ScanOptionsSerializer(read_only=True, many=False)
+    options = DictField(required=False, default={})
```
bruno-fs, line 114: Using a literal `{}` might be problematic. This instance will be modified on the fly if default value is used.
Does `default` accepts a "factory" like `dict`?

quipucords/api/scan/model.py
```diff
@@ -130,6 +131,11 @@ def get_default_forks():
         """Create the default number of forks."""
         return DEFAULT_MAX_CONCURRENCY
 
+    @staticmethod
+    def get_max_forks():
+        """Create the maximum number of forks."""
+        return UPPER_MAX_CONCURRENCY
```
bruno-fs, line 135: I understand you are following the pattern established previously for `get_default_forks` but do you see an advantage in creating a static method on a class that returns a constant instead of simply creating a constant in it?

Not saying we should change it, just trying to see if there's an advantage I'm not seeing here.

bruno-fs (commented):
Awesome job! I bet it was pretty hard.

Please squash the first two commits together ("refactor(Scan): Migrating ScanOptions into the Scan Model" & "refactor(Scan): Update the Scan model to include ScanOptions") because one has the migration and the other one has the actual change to the models. We are unlikely to cherry-pick just one of those commits, but I'd always be extra careful to have atomic commits when making db migrations.

Also, there's a small typo on "refactor(Scan): Preferring to use a dictional for the supported products": dictional ->  dictionary/dict

quipucords/api/scanjob/serializer.py
abellotti: 👍 yes, I see we do that in other defaults too.

quipucords/api/scan/model.py
abellotti: Still using the method, I will update to use a Scan constant.

bruno-fs (approved):
LGTM

PR #2584, opened 2024-02-19
Detect JWS 6 installations

mirekdlugosz:
Make it possible to detect JWS 6 installation through systemd unit file.

JWS 6 supports 2 installation methods - through archive and through RPM. Both will create systemd unit file, which is a recommended way of starting a daemon.

Also fixed a glaring issue in the same file - see comment.

### DIFF
--- a/quipucords/scanner/network/runner/roles/jboss_ws/tasks/main.yml
+++ b/quipucords/scanner/network/runner/roles/jboss_ws/tasks/main.yml
@@ -21,7 +21,7 @@
 
 # If client has not renamed default JWS_HOME folder and followed recommended zip file install instructions, JWS_HOME should be in /opt.
 - name: search for JWS_HOME in search directories
-  raw: export LANG=C LC_ALL=C; find {{search_directories}} -type d 2>/dev/null | egrep '*(ews|jws)*[0-99]\.[0-99]$'
+  raw: export LANG=C LC_ALL=C; find {{search_directories}} -type d 2>/dev/null | egrep '(ews|jws).*[0-99]\.[0-99]$'
   register: internal_jws_find_home_from_search
   ignore_errors: yes
   when: 'jboss_ws_ext'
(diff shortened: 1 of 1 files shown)


quipucords/scanner/network/runner/roles/jboss_ws/tasks/main.yml
```diff
@@ -21,7 +21,7 @@
 
 # If client has not renamed default JWS_HOME folder and followed recommended zip file install instructions, JWS_HOME should be in /opt.
 - name: search for JWS_HOME in search directories
-  raw: export LANG=C LC_ALL=C; find {{search_directories}} -type d 2>/dev/null | egrep '*(ews|jws)*[0-99]\.[0-99]$'
+  raw: export LANG=C LC_ALL=C; find {{search_directories}} -type d 2>/dev/null | egrep '(ews|jws).*[0-99]\.[0-99]$'
```
mirekdlugosz, line 24: Someone clearly confused wildcard with regexp here. The previous version was "ews or jws, zero or more times", and it matched anything. This resulted in considering things like `/usr/lib64/python3.6`, `/usr/lib/firmware/ath10k/WCN3990/hw1.0` or even `/sys/devices/pci0000:00/0000:00:01.2`, which clearly do not contain JWS.

Paired with [DISCOVERY-544](https://issues.redhat.com/browse/DISCOVERY-544), it made extended search actually worse at detecting JWS than standard search.

quipucords/scanner/network/runner/roles/jboss_ws/tasks/main.yml
```diff
@@ -56,7 +56,7 @@
 
 # ------------------------------ DETECT JWS/EWS -------------------------------
 - name: check jws presence with yum
-  raw: export LANG=C LC_ALL=C; yum grouplist jws3 jws3plus jws5 2>/dev/null | grep -A1 'Installed Groups:' | grep 'Red Hat JBoss Web Server [0-99]'
+  raw: export LANG=C LC_ALL=C; yum grouplist jws3 jws3plus jws5 jws6 2>/dev/null | grep -A1 'Installed Groups:' | grep 'Red Hat JBoss Web Server [0-99]'
```
mirekdlugosz, line 59: This is affected by [DISCOVERY-456](https://issues.redhat.com/browse/DISCOVERY-456), but since I already modify this file...

infinitewarp (approved):

PR #2585, opened 2024-02-19
feat: enable request throttling with configurable rates

infinitewarp:
Relates to JIRA: DISCOVERY-573
https://issues.redhat.com/browse/DISCOVERY-573


infinitewarp:
It looks like we can't merge this until we refactor `qpc`'s delete functionality to use bulk calls which current spams the server with many requests to delete each object _individually_. We need to add "delete all" capabilities to our API, and we need to update `qpc` to use these "delete all" APIs so that `qpc`'s `clear --all` commands do not trigger the server's throttle thresholds.

I am setting this PR to draft until we complete that work.

bruno-fs (approved):

ruda (approved):

PR #2586, opened 2024-02-19
feat: add optional "all" input to credentials/bulk_delete/ API

infinitewarp:
Alter the behavior of the `credentials/bulk_delete/` API to add an optional `all` input and make its `ids` input optional. If the `all` input is set with a truthy value, we will attempt to delete _all_ credentials, effectively overriding the previously-required `ids` input. If you send only the `ids` input, the existing bulk delete behavior remains unchanged.

This change will enable a future enhancement to `qpc`'s `clear --all` command to reduce its API call count from O(n) to O(1), and that change is required for `quipucords` to implement effective request rate throttling.

Relates to JIRA: DISCOVERY-468
https://issues.redhat.com/browse/DISCOVERY-468


nicolearagao (approved):
LGTM

quipucords/api/credential/view.py
```diff
@@ -38,27 +38,40 @@ def credential_bulk_delete(request):
...
-        existing_ids = set([cred.id for cred in all_creds])
-        missing_ids = all_ids - existing_ids
-        if missing_ids:
-            raise NotFound(
-                detail=_(messages.CRED_IDS_DO_NOT_EXIST % ids_to_str(missing_ids))
-            )
+        if delete_all:
+            all_creds = Credential.objects.all()
+        else:
+            all_creds = Credential.objects.filter(id__in=ids)
```
mirekdlugosz, line 67: Since query objects are lazy, if you hate `else` like I do, you may:

```suggestion
        all_creds = Credential.objects.all()
        if not delete_all:
            all_creds = all_creds.filter(id__in=ids)
```

mirekdlugosz (approved):

mirekdlugosz:
I just realized: this is for credentials only, but `qpc` supports `clear all` for creds, sources and scans. So I guess we need something like that for all of them.

I am fine figuring out API details for creds here, and then including all the others before un-drafting, or in other PRs.

quipucords/api/credential/view.py
infinitewarp: New version of the code has shifted a bit, but it does something very like your suggestion, eliminating an `else` and `filter`ing the object as you suggested.

bruno-fs (approved):

PR #2589, opened 2024-02-21
ScanJob ViewSet for v2 API

bruno-fs:
Add a new ViewSet for ScanJobs. It has list and retrieve, replacing both
`/v1/jobs/<job-id>`, `/v1/scans/<scan-id>/jobs/` and `/v1/report/merge/jobs/<job-id>`.
Part of the rationale is to support upcoming scanless fingerprint type
jobs that will be created for upload reports (DISCOVERY-452). If we would
follow the original pattern, yet another endpoint for listing jobs would be
required instead.

Compared to the original endpoints, this one lacks: 
- `scan.name` (since we are planning to move away from it - and name don't seem a really useful info in this context)
- tasks (since neither UI nor CLI actually use this info - and it will very likely be populated with way more than just 3 tasks to support parallel celery tasks in the future)
- counts_fingerprint (also not used by UI/CLI - and will very likely be irrelevant in the future).

The new viewset also aims to avoid N+1 issues with related objects, using prefetch related to get info from sources and annotations to calculate system counts.

Relates to JIRA: DISCOVERY-452 DISCOVERY-340

### DIFF
--- a/quipucords/api/scanjob/queryset.py
+++ b/quipucords/api/scanjob/queryset.py
@@ -1,0 +1,80 @@
+"""Module for ScanTaskQuerySet."""
+
+from django.db.models import Case, F, Q, QuerySet, Sum, When
+
+from api.scantask.model import ScanTask
+
+
+class ScanJobQuerySet(QuerySet):
+    """Custom QuerySet for ScanJob."""
+
+    def with_counts(self):
+        """Annotate ScanJob with system counts from associated tasks."""
+        return self.annotate(
+            # inspection type jobs use connect type tasks as reference for systems_count
+            # Other types only use their own type.
+            systems_count=Case(
+                When(
+                    scan_type=ScanTask.SCAN_TYPE_INSPECT,
+                    then=Sum(
+                        "tasks__systems_count",
+                        filter=Q(tasks__scan_type=ScanTask.SCAN_TYPE_CONNECT),
+                        default=0,
+                    ),
+                ),
+                default=Sum(
+                    "tasks__systems_count",
+                    filter=Q(tasks__scan_type=F("scan_type")),
+                    default=0,
+                ),
+            ),
+            systems_scanned=Sum(
+                "tasks__systems_scanned",
+                # only counts of tasks with the same scan_type should be considered
+                filter=Q(tasks__scan_type=F("scan_type")),
+                default=0,
+            ),
+            # case required for following annotations because inpection type scansjobs
+            # sum failed/unreachable values from both connect and inspect tasks, while
+            # connection type scanjobs only look for connection tasks
+            systems_failed=Case(
+                When(
+                    scan_type=ScanTask.SCAN_TYPE_INSPECT,
+                    then=Sum(
+                        "tasks__systems_failed",
+                        filter=Q(
+                            tasks__scan_type__in=(
+                                ScanTask.SCAN_TYPE_CONNECT,
+                                ScanTask.SCAN_TYPE_INSPECT,
+                            )
+                        ),
+                        default=0,
+                    ),
+                ),
+                default=Sum(
+                    "tasks__systems_failed",
+                    filter=Q(tasks__scan_type=F("scan_type")),
+                    default=0,
+                ),
+            ),
+            systems_unreachable=Case(
+                When(
+                    scan_type=ScanTask.SCAN_TYPE_INSPECT,
+                    then=Sum(
+                        "tasks__systems_unreachable",
+                        filter=Q(
+                            tasks__scan_type__in=(
+                                ScanTask.SCAN_TYPE_CONNECT,
+                                ScanTask.SCAN_TYPE_INSPECT,
+                            )
+                        ),
+                        default=0,
+                    ),
+                ),
+                default=Sum(
+                    "tasks__systems_unreachable",
+                    filter=Q(tasks__scan_type=F("scan_type")),
+                    default=0,
+                ),
+            ),
+        )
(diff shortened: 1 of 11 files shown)


quipucords/tests/api/scanjob/test_scanjob_queryset.py
```diff
@@ -0,0 +1,89 @@
...
+    (
+        ("inspect_type_scanjob", 100, 10, 39, 51),
+        ("connect_type_scanjob", 100, 20, 30, 50),
+    ),
+)
+def test_with_counts(  # noqa: PLR0913
+    request, scanjob, count, scanned, failed, unreachable
+):
+    """Test qs.with_counts results against hardcoded values in fixtures."""
+    scanjob = request.getfixturevalue(scanjob)
```
bruno-fs, line 71: Ah, I specifically used `request.getfixturevalue` instead of `lazyfixtures` here since @infinitewarp mentioned we might need to drop it soon

quipucords/api/scanjob/queryset.py
```diff
@@ -0,0 +1,68 @@
...
+            ),
+            systems_scanned=Sum(
+                "tasks__systems_scanned",
+                # only counts of tasks with the same scan_type should be considered
+                filter=Q(tasks__scan_type=F("scan_type")),
+                default=0,
+            ),
+            # case required for following annotations because inpection type scansjobs
+            # sum failed/unreachable values from both connect and inspect tasks, while
+            # connection type scanjobs only look for connection tasks
```
infinitewarp, line 27: Although this sounds reasonable, in practice, do we ever _actually_ have unexpected scan tasks associated with a scan job? For example, do we ever create a connection-type scanjob that has an inspection-type scan type, and if we did, would it ever have non-zero count for systems failed? I'm just wondering if there are specific situations that prevent us from simplifying the `Sum` without any `tasks__scan_type` filter.

Full disclosure: I have not gone through and reviewed the scan job/task code recently. I'm just working out what makes logic sense in my head.

bruno-fs: I *believe* connection type scanjobs only have a single scan task.

It occurred to me a simple sum all, without the case, could be enough - specially because in production fingerprint type tasks always have the value zero in production. But then I started wondering we might want to change that so counts in fingerprint type task have a meaning.

infinitewarp: My instinct says KISS, but if you believe the counts may change in the near future, I'll defer to your judgment. 🙂

bruno-fs: KISS is always a good point. IF we need the complexity later, we defer to add it later.
I'll first confirm if a connection type scan only has a connection task first.

the side effect would be loss of compatibility with the legacy method that does counts, but it doesn't really matter on realistic scenarios.

bruno-fs: Took some time researching and experimenting. Connect type jobs only have connect type tasks, so we could get rid of the case and both connect/inspect type scans would work as intended.

However, for fingerprint type scans things would be a bit odd because we would change the behavior from showing meaningless zero values (which I'd like to change soon ™️ ) to now meaningless counts from other tasks.

bruno-fs: query in postgresql

```
SELECT
    "api_scanjob"."id",
    "api_scanjob"."scan_type",
    "api_scanjob"."status",
    "api_scanjob"."status_message",
    "api_scanjob"."options_id",
    "api_scanjob"."start_time",
    "api_scanjob"."end_time",
    "api_scanjob"."scan_id",
    "api_scanjob"."connection_results_id",
    "api_scanjob"."inspection_results_id",
    "api_scanjob"."report_id",
    CASE
        WHEN "api_scanjob"."scan_type" = inspect THEN COALESCE(
            SUM("api_scantask"."systems_count") FILTER (
                WHERE
                    "api_scantask"."scan_type" = connect
            ),
            0
        )
        ELSE COALESCE(
            SUM("api_scantask"."systems_count") FILTER (
                WHERE
                    "api_scantask"."scan_type" = ("api_scanjob"."scan_type")
            ),
            0
        )
    END AS "systems_count",
    COALESCE(
        SUM("api_scantask"."systems_scanned") FILTER (
            WHERE
                "api_scantask"."scan_type" = ("api_scanjob"."scan_type")
        ),
        0
    ) AS "systems_scanned",
    CASE
        WHEN "api_scanjob"."scan_type" = inspect THEN COALESCE(
            SUM("api_scantask"."systems_failed") FILTER (
                WHERE
                    "api_scantask"."scan_type" IN (connect, inspect)
            ),
            0
        )
        ELSE COALESCE(
            SUM("api_scantask"."systems_failed") FILTER (
                WHERE
                    "api_scantask"."scan_type" = ("api_scanjob"."scan_type")
            ),
            0
        )
    END AS "systems_failed",
    CASE
        WHEN "api_scanjob"."scan_type" = inspect THEN COALESCE(
            SUM("api_scantask"."systems_unreachable") FILTER (
                WHERE
                    "api_scantask"."scan_type" IN (connect, inspect)
            ),
            0
        )
        ELSE COALESCE(
            SUM("api_scantask"."systems_unreachable") FILTER (
                WHERE
                    "api_scantask"."scan_type" = ("api_scanjob"."scan_type")
            ),
            0
        )
    END AS "systems_unreachable"
FROM
    "api_scanjob"
    LEFT OUTER JOIN "api_scantask" ON ("api_scanjob"."id" = "api_scantask"."job_id")
GROUP BY
    "api_scanjob"."id"
```

infinitewarp (approved):

PR #2591, opened 2024-02-27
refactor(inspect-view): remove unused facts info

bruno-fs:
This view is only used by the UI, which isn't consuming nor will consume facts from this view. Removing facts from this view removes a potential issue for sources with multiple hosts/subnets, avoiding unnecessary queries and a unnecessary large json response.

Relates to JIRA: DISCOVERY-218


infinitewarp (approved):
🪓

PR #2592, opened 2024-02-28
Get rid of indermediary models between ScanJob and inspect results

bruno-fs:
As a preparation rite for DISCOVERY-218 refactor, streamline the connection between ScanJob and inspection results.
Also fix a inconsistency in how the result model is named (Inspection vs Inspect).

Reviewing commit per commit is recommend over looking at the whole thing. Post review some commits will be squashed together to ensure proper "atomic commits".


quipucords/api/scanjob/model.py
```diff
@@ -280,6 +267,28 @@ def queue(self):  # noqa: C901
...
+        """Delete InspectResults exclusively related to this ScanJob."""
+        # results_from_other_jobs = InspectResult.objects.exclude(
+        #     tasks__in=self.tasks.all()
+        # ).values(
+        #     "id",
+        # )
+        # InspectResult.objects.exclude(id__in=results_from_other_jobs).filter(
+        #     tasks__job_id=self.id
+        # ).delete()
+        # TODO: why the approach above didn't work? I had to use the m2m model to do it
```
bruno-fs, line 280: I need to drop this commented code, but first... Does anyone see a way to make this deletion work without using through model like below?

quipucords/api/scanjob/model.py
```diff
@@ -280,6 +267,28 @@ def queue(self):  # noqa: C901
...
+        results = InspectResult.objects.filter(tasks__in=self.tasks.all())
+        ThroughModel = InspectResult.tasks.through
+        results_from_other_jobs = (
+            ThroughModel.objects.filter(inspectresult__in=results)
+            .exclude(scantask__in=self.tasks.all())
+            .values(
+                "inspectresult_id",
+            )
+        )
+        results.exclude(id__in=results_from_other_jobs).delete()
```
infinitewarp, line 290:
```suggestion
        InspectResult.objects.filter(tasks__job_id=self.id).exclude(
            tasks__in=ScanTask.objects.exclude(job_id=self.id)
        ).delete()
```
I think this will do what you want. 🙂 I arrived at this odd-looking chain of excludes by breaking down the problem down to what I think are the basic requirements.

We want to find all `InspectResult`s that have a task that have "this" `ScanJob`. That part's easy:
```python
InspectResult.objects.filter(tasks__job_id=self.id)
```

Now, let's find all `ScanTask`s that belong do a _different_ `ScanJob`:
```python
ScanTask.objects.exclude(job_id=self.id)
```

By extending the first queryset to exclude any `InspectResult`s that have `ScanTask`s in the previous set, that effectively means we are saying "give me all results with tasks related to this job, and exclude any results that have tasks related to any _other_ jobs":
```python
InspectResult.objects.filter(tasks__job_id=self.id).exclude(tasks__in=ScanTask.objects.exclude(job_id=self.id))
```

I checked out your branch, make this change, and ran the tests in `test_scanjob.py`. Everything passes, but you may want to triple-check my work or add a couple _more_ tests out of an abundance of caution.

This new command still generates some gnarly subqueries, but I think this is near the best we can get using native queryset operations since Django refuses to implement the standard `<>` not-equals operator which I believe could simplify the generated query. 😡 Here is the raw query I produced by my queryset where the `ScanJob`'s ID is `420`:

```sql
SELECT "api_inspectresult"."id",
       "api_inspectresult"."name",
       "api_inspectresult"."status",
       "api_inspectresult"."source_id"
FROM "api_inspectresult"
INNER JOIN "api_inspectresult_tasks" ON ("api_inspectresult"."id" = "api_inspectresult_tasks"."inspectresult_id")
INNER JOIN "api_scantask" ON ("api_inspectresult_tasks"."scantask_id" = "api_scantask"."id")
WHERE ("api_scantask"."job_id" = 420
       AND NOT (EXISTS
                  (SELECT 1 AS "a"
                   FROM "api_inspectresult_tasks" V1
                   WHERE (V1."scantask_id" IN
                            (SELECT U0."id"
                             FROM "api_scantask" U0
                             WHERE NOT (U0."job_id" = 420))
                          AND V1."inspectresult_id" = ("api_inspectresult"."id"))
                   LIMIT 1)))
```

I don't _love_ it, but it seems _fine_ for now. 🤷‍♂️  By the time our tables grow large enough to cause problems with _this_ query, I'm sure we'll already have plenty of other more urgent problems _elsewhere_ in the system.

quipucords/tests/api/scanjob/test_scanjob.py
```diff
@@ -1997,3 +1984,46 @@ def test_filter_by_status(self, django_client, mocker):
...
+def test_delete_inspect_results(scanjob_with_inspect_results):
+    """Test ScanJob.delete_inspect_results method."""
+    scanjob = scanjob_with_inspect_results
+    assert InspectResult.objects.filter(tasks__job=scanjob).exists()
+    scanjob.delete_inspect_results()
+    assert not InspectResult.objects.filter(tasks__job=scanjob).exists()
+
+
+@pytest.mark.django_db
+def test_delete_inspect_results_binded_to_other_scanjobs(
```
infinitewarp, line 2009:
```suggestion
def test_delete_inspect_results_bound_to_other_scanjobs(
```

quipucords/tests/api/scanjob/test_scanjob.py
```diff
@@ -1997,3 +1984,46 @@ def test_filter_by_status(self, django_client, mocker):
...
+    """Ensure only InspectResults uniquely bound to a ScanJob are deleted."""
+    inspect_res = InspectResult.objects.first()
+    other_scanjob = ScanJobFactory()
+    task = ScanTaskFactory(job=other_scanjob)
+    inspect_res.tasks.add(task)
+    # save id in a variable so we can check later for existence
+    inspect_res_id = inspect_res.id
+
+    scanjob = scanjob_with_inspect_results
+    scanjob.delete_inspect_results()
```
infinitewarp, line 2021:
```suggestion
    scanjob = scanjob_with_inspect_results
    assert InspectResult.objects.filter(tasks__job=scanjob).count() == 3
    scanjob.delete_inspect_results()
```

As I was reworking that ugly delete queryset, I found myself doubting the state of the objects, and I even opened a text file to keep track of the ids of which objects were related to each other. If we have any doubt about the delete query, for maximum readability, we might want to assert a few extra things like this before and after the deletes so that future readers can better intuit the state of things.

Not a must-have. Just a passing thought I had as I was working through unfamiliar code.

bruno-fs: Sorry, github auto resolved this when I accepted the suggestion.

quipucords/tests/api/scanjob/test_scanjob.py
bruno-fs: Sorry, github auto resolved this when I accepted the suggestion.

infinitewarp (approved):
In local context, all of this change seems to make sense to me, but I confess I am struggling to hold all of it in my head at once for a full evaluation. I'm gonna say… let's merge it! 😅

PR #2593, opened 2024-03-01
feat: support partial deletes in bulk delete API

infinitewarp:
This also fixes a few minor issues in recent commits to this API, and this also adds a new `client_logged_in` test fixture that is much faster than our `django_client` fixture that this code was previously using extensively.

Running `poetry run pytest quipucords/tests/api/credential/test_credential.py` on my local system at commit f49787ad (**before** converting it to the new fixture):

```
129 passed, 5 warnings in 116.75s (0:01:56)
```

Running `poetry run pytest quipucords/tests/api/credential/test_credential.py` on my local system at commit fe44a58d (**after** converting it to the new fixture):

```
129 passed, 4 warnings in 8.62s
```

Gotta go fast. 🦔💨

I could split out the new fixtures into a separate PR if requested.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -131,3 +131,6 @@ var/
 
 # vagrant data
 .vagrant/
+
+# test profiling results
+prof/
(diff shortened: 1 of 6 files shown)


bruno-fs (approved):
beautiful <3

PR #2595, opened 2024-03-04
chore: requires Python >= 3.11 for setuptools.

ruda:
We are using Python 3.11 (at minimum) everywhere and we forgot to update setuptools.


infinitewarp (approved):
Thanks for spotting this!

PR #2597, opened 2024-03-04
fix(ScanJob.queue): fix a bug in partial results cleanup

bruno-fs:


infinitewarp (approved):
🔥🐶☕

this-is-fine.gif

PR #2600, opened 2024-03-05
Replace Report.sources JSONField with an actual query

bruno-fs:
This PR deprecates `Report.sources` and replaces it with a query building the same data
stored on `RawFact` model.

I strongly recommend reviewing it in a commit per commit, as there is one particular huge commit that contains just fixes in order to accommodate the new modeling.

# Summary of changes
- `Report.sources` is removed
- A new model, ~~`ResultSet`~~ `InspectGroup`, is introduced and sits between `Task` and `InspectResult`. It is required to represent data in our db in the same way it was represented on `Report.sources`.
- A data migration is added to move data to the new modeling; we might lose some data, but it is not a realistic use case (inspect scan results without sources). In its initial iteration it was taking ~20min to migrate facts for ~110k systems, but it's current iteration took less than ~3minutes. It is still a slow migration, we might decide to simply give up on data and inform users (and prob call this at least a minor release just to reflect this "breaking" change).
- a new `cached_property` is introduced in `Report` model: `sources`, which should represent data the exact same way it the deprecated field. It initially introduces a poorly optimized query that collects identical data, and is later optimized with a query including a specialized `RawSQL` annotation. See the section [Benchmarks](#Benchmarks) below for more info.

# Benchmarks
These benchmarks compare the performance of details and tarball endpoints using the same data on current `main` branch (`1.6.1`), then using the naive/traditional query and finally with the specialized `RawSQL` for raw fact aggregation. The database used for these benchmarks is PostgreSQL, as this is our production target.

## details report as json response time

Endpoint `/api/v1/reports/{report_id}/details/`

| number of systems | query # | main   | traditional query | query with RawSQL |
| ----------------- | ------- | ------ | ----------------- | ----------------- |
| 1k                | 1       | 410ms  | 1.49s             | 529ms             |
| 1k                | 2       | 66.8ms | 1.11s             | 56.4ms            |
| 1k                | 3       | 54ms   | 1.13s             | 175ms             |
| 10k               | 1       | 391ms  | 10.2s             | 455ms             |
| 10k               | 2       | 496ms  | 10.4s             | 562ms             |
| 10k               | 3       | 376ms  | 10.1s             | 611ms             |
| 100k              | 1       | 3.48s  | timed out         | 4.33s             |
| 100k              | 2       | 3.42s  | timed out         | 4.49s             |
| 100k              | 3       | 3.42s  | timed out         | 4.54s             |

## tarball response time

Endpoint `/api/v1/reports/{report_id}/`

| number of systems | attempt # | main  | "traditional" query | query with RawSQL |
| ----------------- | --------- | ----- | ------------------- | ----------------- |
| 1k                | 1         | 311ms | 1.39s               | 400ms             |
| 1k                | 2         | 164ms | 1.14s               | 266ms             |
| 1k                | 3         | 144ms | 1.11s               | 262ms             |
| 10k               | 1         | 2.77s | 11.7s               | 2.11s             |
| 10k               | 2         | 1.69s | 11.5s               | 1.3s              |
| 10k               | 3         | 1.37s | 11.3s               | 1.26s             |
| 100k              | 1         | 27.2s | time out            | 18.2s             |
| 100k              | 2         | 15.1s | time out            | 10.2s             |
| 100k              | 3         | 14.7s | time out            | 10.3s             |


quipucords/api/inspectresult/model.py
```diff
@@ -3,16 +3,134 @@
...
+    select
+        ir.id,
+        ir.result_set_id,
+        (
+        select
+            json_object_agg(raw.name, raw.value) as raw_facts
+        from
+            api_rawfact raw
+        where
+            raw.inspect_result_id = ir.id
```
infinitewarp, line 26: I wonder if we could pull this subquery into its parent query with a regular join. I may pull your branch and try some ideas.

What have you been doing to generate data for your benchmarks? I assume you're using the test factories. Would you share some specific commands that you used so I can follow the same steps for consistency?

quipucords/api/inspectresult/model.py
```diff
@@ -3,16 +3,134 @@
...
+group by
+	raw_agg.result_set_id
+"""
+
+# RAW_QUERY_POSTGRES = """
+# with raw_agg as (
+# select
+# 	ir.id,
+# 	ir.result_set_id,
+# 	json_object_agg(raw.name, raw.value) as raw_facts
```
infinitewarp, line 45: Ah, perhaps this was your attempt to do what I mentioned above. I'd still like to give it a try, though. 🙂

quipucords/api/inspectresult/model.py
```diff
@@ -3,16 +3,134 @@
...
+            raw_sql = RAW_QUERY_SQLITE
+            raw_sql_kw = {"output_field": models.JSONField()}
+        else:
+            raise NotImplementedError(
+                f"with_raw_facts not implemented for {connection.vendor=}"
+            )
+        return self.annotate(facts=RawSQL(raw_sql, params=(), **raw_sql_kw))
+
+
+class ResultSet(models.Model):
```
infinitewarp, line 104: I don't like the name `ResultSet` here because it looks like the common database term "result set" which in some frameworks is also literally called `ResultSet`, but I don't have a better name for this right now. 🤔

bruno-fs: I was calling it `ResultMeta` before, but I hated that name too...

quipucords/api/inspectresult/model.py
bruno-fs: I did try that first, but it was slightly slower (both with real queries and with `EXPLAIN ANALYSE`).

Feel free to play around with it! I shared the data with you on slack, but I realized now that with the code changes introduced to factories we don't have control of how many "systems" we want to generate. So, in order to populate the db with synthetic data we would need to use the factories available on main branch and then migrate them.

Will see if I can quickly patch `ReportFactory` to enable this for the code in this branch.

quipucords/api/inspectresult/model.py
bruno-fs: Maybe `InspectSet`? or `InspectGroup`?

bruno-fs: TBH I hate them all, lol. Naming things is hard.

quipucords/api/migrations/0051_remove_report_sources.py
```diff
@@ -0,0 +1,263 @@
...
+        # "On most database backends (all but PostgreSQL), Django will split the SQL
+        # into individual statements prior to executing them."
+        # https://docs.djangoproject.com/en/4.2/ref/migration-operations/#runsql
+        return
+    with schema_editor.connection.cursor() as cursor:
+        cursor.execute("SET CONSTRAINTS ALL IMMEDIATE")
+
+
+def defer_constraints(apps, schema_editor):
+    """Rename details_report_sequence to report_sequence."""
```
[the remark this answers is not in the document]
bruno-fs: lol, no.

thanks for spotting this

quipucords/api/scantask/model.py
```diff
@@ -371,16 +376,21 @@ def cleanup_facts(self, identity_key):
         """
         if self.scan_type != ScanTask.SCAN_TYPE_INSPECT:
             return
+        # TODO: should we even be doing this cleanup without logging?
+        # my gut feeling is just mark it as failed (and stop post processing
+        # failed data)
```
infinitewarp, line 381: Now that you mention it, yeah, I think this could benefit from at least a `logger.warning` message here. That might help us to identify this "cleanup" operation when investigating future data anomalies.

infinitewarp: And even better: If we start seeing the new warning a lot in our local development and testing, that might indicate that we should follow your gut and abort processing more aggressively.

infinitewarp (approved):

bruno-fs:
marked the PR as draft as a reminder for me not to merge before squashing the extra commit

infinitewarp (approved):
Looking good. Thanks for handling all our suggestions!

quipucords/tests/api/scantask/test_scantask_get_facts.py
```diff
@@ -1,56 +1,47 @@
...
-    systems = []
-    for _ in range(50):
-        systems.append({random_name(): random_value() for _ in range(100)})
-    return systems
+    return [
+        {
+            "name": letter,
+            "boolean": faker.pybool(),
+            "int": faker.pyint(max_value=100),
+            "str": faker.slug(),
```
infinitewarp, line 19: I would suggest giving the keys random `faker.slug()` values just to make it very obvious for future debugging developers that these are purely synthetic values, never to be confused with the `int` or `str` types.

bruno-fs:
@infinitewarp I had to add another commit to fix an issue with a flaky test https://github.com/quipucords/quipucords/pull/2600/commits/fad2c162769207bbc130270ccb0738c09e7510c5

quipucords/tests/api/scantask/test_scantask_get_facts.py
bruno-fs: ok!

bruno-fs: updated =D

quipucords/tests/api/scantask/test_scantask_get_facts.py
```diff
@@ -1,55 +1,46 @@
...
     """Fixture representing a list of random facts."""
-    systems = []
-    for _ in range(50):
-        systems.append({random_name(): random_value() for _ in range(100)})
-    return systems
+    return [
+        {
+            "name": faker.slug(),
+            "boolean": faker.pybool(),
+            "int": faker.pyint(max_value=100),
```
infinitewarp, line 18: regarding https://github.com/quipucords/quipucords/pull/2600#discussion_r1569288521

```suggestion
            faker.slug(): faker.slug(),
            faker.slug(): faker.pybool(),
            faker.slug(): faker.pyint(max_value=100),
```

quipucords/tests/api/scantask/test_scantask_get_facts.py
infinitewarp: Did I miss a change? I still see the hard-coded keys that look like types. See also https://github.com/quipucords/quipucords/pull/2600#discussion_r1569448915

quipucords/tests/api/scantask/test_scantask_get_facts.py
infinitewarp: Or do the type-looking keys have a special meaning that I've overlooked?

bruno-fs: aaaah
except for "name", they don't.

bruno-fs: "name" (or any other hardcoded string) is required for sorting later

quipucords/tests/api/scantask/test_scantask_get_facts.py
bruno-fs: sorry, I misunderstood what you were proposing. see https://github.com/quipucords/quipucords/pull/2600#discussion_r1569503580

PR #2603, opened 2024-03-11
feat(CeleryWorker): Allow for min and max concurrency specification

abellotti:
- Provide ability to deploy the celery worker to run with a minimum and maximum concurrency via the following 2 environment variables:

    QPC_CELERY_WORKER_MIN_CONCURRENCY
    QPC_CELERY_WORKER_MAX_CONCURRENCY

Defaults to 10, 10 for consistency with the current default (--concurrent=10 when not specified)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,8 @@ BINDIR  = bin
 PARALLEL_NUM ?= $(shell $(PYTHON) -c 'import multiprocessing as m;print(int(max(m.cpu_count()/2, 2)))')
 TEST_TIMEOUT ?= 15
 TEST_OPTS := -n $(PARALLEL_NUM) -ra -m 'not slow' --timeout=$(TEST_TIMEOUT) --durations=10
+QPC_CELERY_WORKER_MIN_CONCURRENCY ?= 10
+QPC_CELERY_WORKER_MAX_CONCURRENCY ?= 10
 
 QUIPUCORDS_CONTAINER_TAG ?= quipucords
 QUIPUCORDS_UI_PATH ?= ../quipucords-ui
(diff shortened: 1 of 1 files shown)


abellotti:
Tested with an updated (and simplified) Quadlet deployment of Discovery.

Makefile
```diff
@@ -142,7 +142,11 @@ server-randomize-sequences:
 	$(PYTHON) quipucords/manage.py randomize_db_sequences --settings quipucords.settings
 
 celery-worker:
-	$(PYTHON) -m celery --app quipucords --workdir quipucords worker
+	@if [ -n "${QPC_CELERY_WORKER_MIN_CONCURRENCY}" -a -n "${QPC_CELERY_WORKER_MAX_CONCURRENCY}" ]; then \
+		$(PYTHON) -m celery --app quipucords --workdir quipucords worker --autoscale=${QPC_CELERY_WORKER_MAX_CONCURRENCY},${QPC_CELERY_WORKER_MIN_CONCURRENCY}; \
+	else \
+		$(PYTHON) -m celery --app quipucords --workdir quipucords worker; \
+	fi
```
infinitewarp, line 149: If we assign defaults (maybe `1`?) to `QPC_CELERY_WORKER_MIN_CONCURRENCY` and `QPC_CELERY_WORKER_MAX_CONCURRENCY`, can we collapse this conditional fork into a single statement? Do you know what would happen if you invoke the worker with the same two values like `--autoscale=1,1`? How does that compare with not passing `--autoscale` at all? How many workers does it start be default with no `--autoscale`?

I'm thinking about how I've used `--concurrency #` elsewhere to define a static number of worker threads/processes, but I see we weren't using that here.

abellotti: default (with nothing specified), is --concurrency, its value is for a pool of 10 processes. we can specify any other value via --concurrency=#.  When the autoscaler is specified, via --autoscale=max,min, it overrides anything specified with --concurrency. If we default with 1,1 we'd be eliminating the default 10 concurrency. I guess the equivalent to the default is --autoscale=10,10. (but that defeats having the autoscaler). The nice thing about the autoscaler is that it can grow the pool as it sees fit (memory/cpu that can be configured).  Maybe we can default to the autoscaler to something like --autoscale=10,1 or --autoscale=10,4 maybe?  wdyt?

abellotti: Thinking about this a bit more, we may not want the auto-scaler for OpenShift deployments where we have the multiple celery workers scaled up via the Horizontal Pod Autoscaler.  In such scenarios, we'd want the static number of processes.  We can have that default configurable with something like:

```
QPC_CELERY_WORKER_CONCURRENCY=10
```

But for Quadlet/Systemd deployments, we can enable the celery worker process auto-scaler via the optional environment variables (no default), i.e.:

```
QPC_CELERY_WORKER_MIN_CONCURRENTY=4
QPC_CELERY_WORKER_MAX_CONCURRENCY=16
```

abellotti: as discussed on a separate thread, we'll just allow for the celery auto-scaler configuration to keep it less confusing, but we'd default the environment to the static 10 for consistency with the current default:

```
QPC_CELERY_WORKER_MIN_CONCURRENTY=10
QPC_CELERY_WORKER_MAX_CONCURRENCY=10
```

infinitewarp: Thanks for all the details! I'll leave this thread open for anyone else who wants to review, but please feel free to resolve it when you're ready to merge.

infinitewarp (approved):

abellotti:
Here's the new default _static_ but now configurable concurrency {min=10, max=10}: 

image

abellotti:
Unrelated, failed CI due to vCenter being down:

```
E                           'status_message': 'Unable to connect to VCenter source vcenter%s.\n'
E                                             'Reason for failure: [Errno -2] Name or service '
E                                             'not known',
```

Not clear when it will be available, going ahead with the merge.

PR #2605, opened 2024-03-12
feat: add support for bulk deleting sources

infinitewarp:
This change is a followup to https://github.com/quipucords/quipucords/pull/2540, https://github.com/quipucords/quipucords/pull/2586, and https://github.com/quipucords/quipucords/pull/2593 to add the ability to bulk delete sources with a single HTTP POST. The new sources bulk delete API behaves similar to the recent credentials bulk delete API.

To delete *all* sources, post a json payload like `{"ids": "all"}`:

```bash
http post \
  http://127.0.0.1:8000/api/v1/sources/bulk_delete/ \
  'Cookie:csrftoken=mytoken; sessionid=mysession' 'X-CSRFToken:mytoken' \
  ids:='"all"'
```

To delete specific sources, post a json payload like `{"ids": [1,2,3,4]}`:

```bash
http post \
  http://127.0.0.1:8000/api/v1/sources/bulk_delete/ \
  'Cookie:csrftoken=mytoken; sessionid=mysession' 'X-CSRFToken:mytoken' \
  ids:='[1,2,3,4]'
```

Sources that are still in use by other scans (or scan-related objects) will be skipped and reported as such in the response. For example:

```json
{
  "message": "Deleted 4 sources. Could not find 0 sources. Failed to delete 4 sources.",
  "deleted": [
    18090,
    18091,
    18092,
    18093
  ],
  "missing": [],
  "skipped": [
    {
      "source": 18078,
      "scans": [
        11707
      ],
      "scanjobs": [
        19200
      ],
      "scantasks": [
        15759,
        15758
      ]
    },
    {
      "source": 18079,
      "scans": [
        11708
      ],
      "scanjobs": [
        19201
      ],
      "scantasks": [
        15761,
        15762
      ]
    },
    {
      "source": 18080,
      "scans": [
        11709
      ],
      "scanjobs": [
        19202
      ],
      "scantasks": [
        15765,
        15764
      ]
    },
    {
      "source": 18081,
      "scans": [
        11710
      ],
      "scanjobs": [
        19203
      ],
      "scantasks": [
        15768,
        15767
      ]
    }
  ]
}
```

### DIFF
--- a/quipucords/api/source/view.py
+++ b/quipucords/api/source/view.py
@@ -9,7 +9,6 @@ from django.utils.translation import gettext as _
 from django_filters.rest_framework import CharFilter, DjangoFilterBackend, FilterSet
 from rest_framework import status
 from rest_framework.decorators import api_view
-from rest_framework.exceptions import ParseError
 from rest_framework.filters import OrderingFilter
 from rest_framework.response import Response
 from rest_framework.serializers import ValidationError
@@ -17,11 +16,12 @@ from rest_framework.viewsets import ModelViewSet
 
 from api import messages
 from api.common.util import (
-    DELETE_ALL_IDS_MAGIC_STRING,
+    ALL_IDS_MAGIC_STRING,
     convert_to_boolean,
     expand_scanjob_with_times,
     is_boolean,
     is_int,
+    set_of_ids_or_all_str,
 )
 from api.filters import ListFilter
 from api.models import Scan, ScanJob, ScanTask, Source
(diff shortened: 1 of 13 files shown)


quipucords/api/source/view.py
```diff
@@ -56,6 +60,105 @@ def format_source(json_source):
...
+            ],
+        }
+
+    input:      "ids" : List of ids to delete, or string ALL_IDS_MAGIC_STRING
+    returns:    200 OK - upon successfully deleting any sources.
+                400 Bad Request - ids list is missing or empty.
+    """
+    ids = set_of_ids_or_all_str(request.data.get("ids"))
+
+    with transaction.atomic():
```
abellotti, line 95: I thought we were no longer going with atomic transactions for the new bulk deletes

infinitewarp: This transaction isn't strictly necessary in the general happy path, but I don't think it's hurting, and it might help in some narrow edge cases.

In both bulk delete APIs, we now explicitly search for and filter out any objects that we believe could prevent the delete from succeeding. However, Django doesn't _actually_ issue a single `DELETE` command to the database when you tell the model to delete a lot of objects. Internally, it still iterates and triggers signals related to the models, and it's possible that another request at the same time could cause a race condition in the related objects. It's admittedly a very narrow possibility for our app, but I believe that we probably _would_ want a transaction to protect and roll back in that case. Otherwise, we'd need to implement more explicit try-except logic in both view functions to prevent the response from halting and returning a `500` error.

nicolearagao (approved):
LGTM

ruda (approved):
Looks good. Test cases are great.

PR #2607, opened 2024-03-14
Celery Scan Manager - Add support for stopping a scan

abellotti:
feature(Celery): Add support to stop celery tasks.
    
 - Installing the django-redis package
 - Enabling the Django Redis Cache
 - Defaults TTL to 60 seconds.
 - Store the celery task id in the redis cache upon successfully starting a scan.
 - Fetch the celery task id from the cache upon stopping/killing a scan job.
 - Trigger the AsyncResult revoke for canceling the celery task process
      (and related tasks via terminate=True)


abellotti:
Scans can now be stopped in the UI (and thus API) with the celery scan manager enabled. We used to get a 500 (Not Implemented error).

image

and in the Server log you'll now see the following:

```
...
[INFO 2024-03-14T15:47:14 pid=67090 tid=6190362624 scanner/manager.py:put:81] CELERY SCAN MANAGER: Started the Celery Task 0c0c8809-a913-4989-9823-bd6b8e2c151c for scan job 293
...
[INFO 2024-03-14T15:47:19 pid=67090 tid=6190362624 api/scanjob/model.py:log_message:116] Job 293 (inspect, elapsed_time: 5s) - SIGNAL CANCEL received for scan job.
[INFO 2024-03-14T15:47:19 pid=67090 tid=6190362624 scanner/manager.py:kill:57] CELERY SCAN MANAGER: Canceling the Celery Task 0c0c8809-a913-4989-9823-bd6b8e2c151c for scan job 293
...
```

quipucords/quipucords/settings.py
```diff
@@ -476,3 +476,38 @@ def app_secret_key_and_path():
...
+# Defining both local memory and Redis as Django back-end caches.
+#
+# For now, we're keeping the default as local memory cache, this way
+# the application will continue to work without the Celery scan manager enabled.
+# Once quipucords only supports the Celery Scan Manager and Redis, we can then
+# switch the default to be Redis.
+QPC_CACHE_TTL_DEFAULT = env.int("QPC_CACHE_TTL_DEFAULT", default=600)
+
+# For test environments, we only want to communicate with the local memory cache.
+QUIPUCORDS_ENVIRONMENT = env("QUIPUCORDS_ENVIRONMENT", default="production")
```
infinitewarp, line 489: We already have `PRODUCTION = env.bool("PRODUCTION", False)`. Why not use that? Do you expect to have more than just production and test/"not production" in the near future?

bruno-fs: related https://github.com/quipucords/quipucords/pull/2607/commits/5b59c720947899ddc6c6448d782172dc273606ec#r1556202047

quipucords/quipucords/settings.py
```diff
@@ -484,17 +484,29 @@ def app_secret_key_and_path():
...
+# For test environments, we only want to communicate with the local memory cache.
+if os.environ.get("QUIPUCORDS_ENVIRONMENT") == "test":
+    CACHES = {
+        "default": {
+            "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
+        },
+        "redis": {
+            "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
+        },
+    }
```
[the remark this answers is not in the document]
bruno-fs: Also, there's the `DummyCache` which seems ok for testing. Did you check that one?

(I'd link django docs here, but they are broken ATM 😢 )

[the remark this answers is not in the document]
bruno-fs: ha, docs are (partially) back
https://docs.djangoproject.com/en/4.2/topics/cache/#dummy-caching-for-development

quipucords/quipucords/settings.py
infinitewarp: Also, if we really do want a new setting for this, I'd favor something much more direct with its meaning like `QUIPUCORDS_ENABLE_REDIS_CACHE`.

abellotti: well prod/dev/test environments (dev like running from bash via make serve which we may want to run with or without redis), but yes, the main  settings are:
- thread scan manager, no redis
- celery scan manager, redis.  (prod/dev)
- celery scan manager, no-redis.  (dev/test)

I'll drop the QUIPUCORDS_ENVIRONMENT to the additional control QUIPUCORDS_ENABLE_REDIS_CACHE (which would default to True if QPC_ENABLE_CELERY_SCAN_MANAGER is True). That should give us all the controls we need (including the override settings we'd use in tests). 🤞

quipucords/quipucords/settings.py
```diff
@@ -483,30 +483,36 @@ def app_secret_key_and_path():
...
+# for development or testing purposes.
+
+defaultEnableRedisCache = False
+if QPC_ENABLE_CELERY_SCAN_MANAGER:
+    defaultEnableRedisCache = True
+
+QUIPUCORDS_ENABLE_REDIS_CACHE = env.bool(
+    "QUIPUCORDS_ENABLE_REDIS_CACHE", defaultEnableRedisCache
+)
+
```
bruno-fs, line 498: mmmm we need more ruff checks. this is violating the style we adopted.

please use snake case for this variable.

alternatively, I think you could simply use `QPC_ENABLE_CELERY_SCAN_MANAGER` as the default value for `QUIPUCORDS_ENABLE_REDIS_CACHE`

```
QUIPUCORDS_ENABLE_REDIS_CACHE = env.bool("QUIPUCORDS_ENABLE_REDIS_CACHE", QPC_ENABLE_CELERY_SCAN_MANAGER)
```

infinitewarp (approved):

PR #2609, opened 2024-03-18
feat: add support for bulk deleting scans

infinitewarp:
This change is a followup to https://github.com/quipucords/quipucords/pull/2540, https://github.com/quipucords/quipucords/pull/2586, https://github.com/quipucords/quipucords/pull/2593, https://github.com/quipucords/quipucords/pull/2605, and https://github.com/quipucords/quipucords/pull/2608 to add the ability to bulk delete scans with a single HTTP POST. The new scans bulk delete API behaves similar to the recent credentials bulk delete and sources bulk delete APIs.

To delete *all* scans, post a json payload like `{"ids": "all"}`:

```bash
http post \
  http://127.0.0.1:8000/api/v1/scans/bulk_delete/ \
  'Cookie:csrftoken=mytoken; sessionid=mysession' 'X-CSRFToken:mytoken' \
  ids:='"all"'
```

To delete specific scans, post a json payload like `{"ids": [13652,13653,13654,13655]}`:

```bash
http post \
  http://127.0.0.1:8000/api/v1/scans/bulk_delete/ \
  'Cookie:csrftoken=mytoken; sessionid=mysession' 'X-CSRFToken:mytoken' \
  ids:='[13652,13653,13654,13655]'
```

Responses generally look like this example:

```json
{
    "deleted": [
        13654,
        13655
    ],
    "message": "Deleted 2 scans. Could not find 2 scans.",
    "missing": [
        13652,
        13653
    ]
}
```

Note that if you bulk delete all scans _while_ a scan is running, the delete will succeed, and the server will (likely very soon) raise an exception and log errors with stack traces like this:

```
[ERROR 2024-03-18T18:03:38 pid=8522 tid=123145546133504 api/scantask/model.py:_log_scan_message:173] Job 17514, Task 2 of 3 (inspect, network, dev01 rhel9-with-ssh, elapsed_time: 6s) - Unexpected failure occurred. See context below.
SCAN JOB: ScanJob object (17514)
TASK: ScanTask object (16969)
SOURCE: Source object (13499)
CREDENTIALS: [['Credential object (17651)']]
```
...
```
psycopg2.errors.ForeignKeyViolation: insert or update on table "api_scantask" violates foreign key constraint "api_scantask_job_id_d5f59645_fk_api_scanjob_id"
DETAIL:  Key (job_id)=(17514) is not present in table "api_scanjob".
```
...
```
django.db.utils.IntegrityError: insert or update on table "api_scantask" violates foreign key constraint "api_scantask_job_id_d5f59645_fk_api_scanjob_id"
DETAIL:  Key (job_id)=(17514) is not present in table "api_scanjob".

[INFO 2024-03-18T18:03:40 pid=7038 tid=123145546133504 api/scanjob/model.py:log_message:116] Job 17514 (inspect, elapsed_time: 0s) - SCAN JOB MANAGER: Process successfully terminated.
```

The server itself recovers just fine after those errors, and I'm able to interact with the API and UI normally, create a new scan, etc.

Relates to JIRA: DISCOVERY-573 DISCOVERY-468

### DIFF
--- a/quipucords/api/scan/view.py
+++ b/quipucords/api/scan/view.py
@@ -17,7 +17,12 @@ from rest_framework.viewsets import ModelViewSet
 
 from api import messages
 from api.common.pagination import StandardResultsSetPagination
-from api.common.util import expand_scanjob_with_times, is_int
+from api.common.util import (
+    ALL_IDS_MAGIC_STRING,
+    expand_scanjob_with_times,
+    is_int,
+    set_of_ids_or_all_str,
+)
 from api.filters import ListFilter
 from api.models import Scan, ScanJob, ScanTask, Source
 from api.scanjob.serializer import expand_scanjob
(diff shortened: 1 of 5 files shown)


infinitewarp:
Note: This branch includes the commit from https://github.com/quipucords/quipucords/pull/2608 which really should merge before _this_ PR does.

quipucords/api/scan/view.py
```diff
@@ -133,6 +138,95 @@ def expand_scan(json_scan):
...
+    :param sender: optional sender for cancel_scan signal
+    """
+    logger.info("Deleting scan jobs associated with scan %s", scan.id)
+    if scan.jobs is not None:
+        jobs_to_cancel = scan.jobs.exclude(
+            Q(status=ScanTask.FAILED)
+            | Q(status=ScanTask.CANCELED)
+            | Q(status=ScanTask.COMPLETED)
+        )
+        if jobs_to_cancel:
```
bruno-fs, line 154: uh, never did this before!

is this equivalent to `jobs_to_cancel.exists()`? or just makes the query and is truth if it has results?

quipucords/tests/api/scan/test_scan_bulk_delete.py
```diff
@@ -0,0 +1,132 @@
...
+
+        # Deleting a scan should clean up its jobs and tasks.
+        with pytest.raises(scan2job.DoesNotExist):
+            scan2job.refresh_from_db()
+        with pytest.raises(scan2task.DoesNotExist):
+            scan2task.refresh_from_db()
+        # However, its report is orphaned and breaks its link to the job.
+        # This is a "legacy" behavior that I'm not too happy with.
+        # Maybe we should change that and clear objects better in the future.
+        # TODO Maybe deleting a scan/task/job should also delete its reports.
```
bruno-fs, line 115: Agree we definitely need to change this behavior in the future.
Soon reports will lack the json field duplicating data, making those orphans even more useless

quipucords/tests/api/credential/test_credential.py
```diff
@@ -1,4 +1,4 @@
-"""Test the API application."""
+"""Tests for quipucords.api.source.view."""
```
bruno-fs: I like that u are using `quipucords` as a main module name. I wish we weren't using default django organization and had it that way. some day!

bruno-fs (approved):

quipucords/tests/api/credential/test_credential.py
bruno-fs: s/module/package/

infinitewarp: We'll consider this an aspirational comment for now. 😉 We would love to have saner package namespaces in the future. Maybe one day this string just automatically becomes correct.

quipucords/api/scan/view.py
infinitewarp: I'll be honest here: I didn't write this code and I didn't pay much attention to it! 😅  I just cut and paste from the method `ScanViewSet.destroy` to this new function `_destroy_scan` so that my new code could perform the same operations and stay DRY. ☀️

That being said, yes, it seems like this does the latter: execute the query and be truthy if it produces any results. It's not recommended, though, if you don't also want to iterate through the results. From the [Django docs](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#when-querysets-are-evaluated):

> Note: If you only want to determine if at least one result exists (and don’t need the actual objects), it’s more efficient to use [exists()](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#django.db.models.query.QuerySet.exists).

Fortunately, in this case, we _do_ want to iterate the results in most cases, and this is fine in terms of efficiency.

PR #2614, opened 2024-03-28
test: replace django_client fixture with the much faster client_logged_in

bruno-fs:
The reality is we don't need a live server for our tests. GOTTA GO FAST.

### DIFF
--- a/quipucords/tests/api/details_report/test_details_report.py
+++ b/quipucords/tests/api/details_report/test_details_report.py
@@ -5,6 +5,7 @@ import tarfile
 from io import BytesIO
 
 import pytest
+from rest_framework import status
 from rest_framework.reverse import reverse
 
 from api.models import Report
@@ -31,11 +32,13 @@ def sources(faker):
 class TestDetailsReport:
     """Test details report view."""
 
-    def test_retrieve_json(self, sources, django_client):
+    def test_retrieve_json(self, sources, client_logged_in):
         """Test details report view in json format (the default)."""
         report = ReportFactory(sources=sources)
-        response = django_client.get(reverse("v1:reports-details", args=(report.id,)))
-        assert response.ok, response.text
+        response = client_logged_in.get(
+            reverse("v1:reports-details", args=(report.id,))
+        )
+        assert response.status_code == status.HTTP_200_OK, response.content.decode()
         assert response.json() == {
             "report_id": report.id,
             "sources": sources,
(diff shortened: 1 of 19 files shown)


quipucords/tests/integration/test_smoker.py
```diff
@@ -206,17 +219,27 @@ def test_deployments_report(  # noqa: PLR0913
...
     ):
         """Compare individual reports with the ones bundled on the tarball endpoint."""
-        details_response = django_client.get(f"reports/{report_id}/details/")
-        assert details_response.ok, details_response.json()
+        details_response = client_logged_in.get(
+            reverse("v1:reports-details", args=(report_id,))
+        )
+        assert (
+            details_response.status_code == status.HTTP_200_OK
+        ), details_response.json()
```
bruno-fs, line 230: @infinitewarp I'm almost convincing myself to hack django client response so we can keep `response.ok`. it was so elegant... look at this mess! 🤣

bruno-fs:
the file that lose coverage will be removed once @mirekdlugosz get rid of the container integration tests.
I wonder if we can simply ignore this decrease for now 🤔

quipucords/tests/api/details_report/test_details_report.py
```diff
@@ -89,6 +92,6 @@ def test_retrieve_csv(self, sources, django_client):
             "batata,tomate\r\n"
             "\r\n\r\n"
         )
-        assert response.text == expected_csv
+        assert response.content.decode() == expected_csv
```
bruno-fs, line 95: `.text` is yet another elegant property from requests we are losing

(*urge to hack django client response intensifies*)

quipucords/tests/api/reports/test_reports.py
```diff
@@ -211,25 +225,30 @@ def test_report_tarball_deployments_json(
...
+    details_csv_response = client_logged_in.get(
         reverse("v1:reports-details", args=(deployments_report.report.id,)),
         headers={"Accept": "text/csv"},
     )
-    assert details_csv_response.ok, details_csv_response.text
-    assert details_csv_response.text == tarball_details_csv
+    assert (
+        details_csv_response.status_code == status.HTTP_200_OK
+    ), details_csv_response.content.decode()
+    assert details_csv_response.content.decode() == tarball_details_csv
```
bruno-fs, line 251: ugh, my eyes... readability matters a lot 😭

bruno-fs: F readability

infinitewarp (approved):
LGTM! 👍 

We don't need to address this now, but spotting that test file with _3300+_ lines really makes me want to break it up into multiple more manageable files. Just expanding its diff here on GitHub caused my browser to hang for a while as it loaded everything in.

PR #2615, opened 2024-04-02
fix: allow /api/v1/status to be public

abellotti:
- As we use status for liveness probes and such, let's make this endpoint public, but still do the CSRF protection.
- Added tests to test without a logged in user.


abellotti:
Preferring a separate /api/v1/ping public endpoint as this endpoint returns too much information for the purpose of a liveness probe.

PR #2616, opened 2024-04-03
feature(ping): Adding a public ping endpoint

abellotti:
- adding a public /api/v1/ping/ endpoint
- text/plain pong response
- can be used for liveness probes

### DIFF
--- a/quipucords/api/status/view.py
+++ b/quipucords/api/status/view.py
@@ -1,7 +1,14 @@
 """View for server status."""
 import os
 
-from rest_framework.decorators import api_view
+from django.http import HttpResponse
+from django.views.decorators.csrf import csrf_protect
+from rest_framework.decorators import (
+    api_view,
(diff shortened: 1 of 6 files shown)


quipucords/api/status/view.py
```diff
@@ -33,3 +40,12 @@ def status(request):
             env_dict[key] = value
     server_info["environment_vars"] = env_dict
     return Response(server_info)
+
+
+@api_view(["GET"])
+@csrf_protect
```
bruno-fs, line 46: w8, why this view requires csrf protection?

bruno-fs (approved):

infinitewarp (approved):

quipucords/api/status/view.py
abellotti: overkill, not needed for a GET endpoint. just removed it

PR #2618, opened 2024-04-05
fix: restore broken fact sanitization

infinitewarp:
This previously-working behavior regressed at 86ff19d9c3b65ebe82eca5c2481acb7bc246f1d8.


bruno-fs (approved):
TY for fixing my mess

PR #2621, opened 2024-04-05
feat: check psql liveness before running migrations

infinitewarp:
If psql is the specified database but its port is not open, we now loop and emit messages like this before trying to run migrations:

```
Checking localhost:5432 status...
Ncat: Connection refused.
Checking localhost:5432 status...
Ncat: Connection refused.
```

Once the port is open, the loop exits, and migrations proceed.

I also upgraded to `postgresql:15` in our GitHub CI jobs since we recently updated to `postgresql-15` in discovery-ci.

I based these changes on the branch in https://github.com/quipucords/quipucords/pull/2622. So, that PR should merge before this one, or I need to yank its commits.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,7 +43,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       postgres:
-        image: postgres:12
+        image: postgres:15
         env:
           POSTGRES_PASSWORD: postgres
         ports:
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,6 +23,7 @@ RUN dnf install \
         jq \
         libpq \
         make \
+        nmap-ncat \
         openssh-clients \
         procps-ng \
         python3.11 \
--- a/deploy/check_related_services_liveness.sh
+++ b/deploy/check_related_services_liveness.sh
@@ -38,4 +38,19 @@ function check_db_liveness() {
     fi
 }
 
+function check_redis_liveness() {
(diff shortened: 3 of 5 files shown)


bruno-fs (approved):
Beautiful. LGTM

.github/workflows/test.yml
```diff
@@ -35,6 +35,7 @@ jobs:
       matrix:
         type:
           [
+            shellcheck,
             lint,
```
bruno-fs, line 39: I have the impression that for many linting jobs the setup on GitHub CI will take more time than the execution of the linter.

Mind checking if this is true for shellcheck? If so, I suggest modifying `make lint` to also execute this one.

bruno-fs:
@abellotti 
> Was thinking also to do the same in entrypoint_celery_worker but I don't think we need it there.

I think it might make sense. If celery worker picks up a job and tries to execute without connecting to db we will have a problem. Same for redis con ection on api container: we might need it as well, because discovery needs redis to send the jobs for celery worker.

But what we have here is an amazing start. Baby steps 👶

abellotti:
you're right @bruno-fs it wouldn't hurt. We can move the new functions/check db liveness to a common place and call from the deploy/entrypoint_celery_worker.sh script too.

infinitewarp:
@bruno-fs @abellotti I took your advice and made a few changes here:

* I moved the checking logic to a new file `check_related_services_liveness.sh`.
* Both `entrypoint_web.sh` and `entrypoint_celery_worker.sh` now call it.
* I added a check for `redis` liveness.

PR #2622, opened 2024-04-05
test: run shellcheck against deploy scripts to avoid bugs

infinitewarp:
I've been using `shellcheck` on my local development environments for a few years, and I find it helps make shell scripts more consistent in their style and catches bugs that I missed myself. I thought it might be time to add it to `quipucords` since our little shell scripts are starting to grow a bit.

This change adds a PR check to run `shellcheck ./deploy/*.sh`. I also applied several recommended `shellcheck` fixes to our existing code to make the check happy:

```
SC2006 (style): Use $(...) notation instead of legacy backticks `...`.
SC2046 (warning): Quote this to prevent word splitting.
SC2181 (style): Check exit code directly with e.g. 'if mycmd;', not indirectly with $?.
SC2248 (style): Prefer double quoting even when variables don't contain special characters.
SC2250 (style): Prefer putting braces around variable references even when not strictly required.
SC2292 (style): Prefer [[ ]] over [ ] for tests in Bash/Ksh.
SC2312 (info): Consider invoking this command separately to avoid masking its return value (or use '|| true' to ignore).
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,8 +69,11 @@ jobs:
           cache: "poetry"
           cache-dependency-path: "poetry.lock"
 
-      - name: Install dependencies
+      - name: Install Python dependencies
         run: poetry install
+
+      - name: Install additional tools
+        run: sudo apt-get install -y shellcheck
 
       - name: Run Test
         run: make ${{ matrix.type }} -k
--- a/.shellcheckrc
+++ b/.shellcheckrc
@@ -1,0 +1,9 @@
+enable=add-default-case
+enable=avoid-nullary-conditions
+enable=check-extra-masked-returns
+enable=check-set-e-suppressed
+enable=check-unassigned-uppercase
+enable=deprecate-which
+enable=quote-safe-variables
+enable=require-double-brackets
+enable=require-variable-braces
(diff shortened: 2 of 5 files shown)


.github/workflows/test.yml
```diff
@@ -35,6 +35,7 @@ jobs:
       matrix:
         type:
           [
+            shellcheck,
```
bruno-fs, line 38: I have the impression that for many linting jobs the setup on GitHub CI will take more time than the execution of the linter.

Mind checking if this is true for shellcheck? If so, I suggest modifying make lint to also execute this one.

infinitewarp: That's a good point. I'll put `shellcheck` into the `lint` target. 👍 

Regardless, all the matrix jobs should run in parallel, assuming GitHub resources are plentiful as usual, and I doubt this would be slower than any of the others.

infinitewarp: Done. There's a dedicated `lint-shell` target that is now called from `lint` like the others, and the existing `lint` GitHub job includes this now.

In the `lint` target definition, I put `lint-shell` before `lint-ruff` because `shellcheck` is faster than `ruff` because it's only checking two small files. 🦔💨

bruno-fs (approved):

PR #2624, opened 2024-04-08
chore: update ruff

bruno-fs:


pyproject.toml
```diff
@@ -57,7 +57,7 @@ pytest-profiling = "^1.7.0"
 ipython = "^8.10.0"
 django-extensions = "^3.2.1"
 notebook = "^6.5.3"
-ruff = "^0.1.3"
+ruff = "*"
```
infinitewarp, line 60: I know this is a dev-only dependency, but I don't like declaring version `*` unless I have a really good reason. What's the reason here instead of just bumping the semver?

bruno-fs: reasoning is that ruff keeps updating minors and we miss those because it didn't reach v1 yet
(the syntax with `^x.y.z` will only accept `z` updates when `x` is zero).

I'm also not a fan of a star dependency, but I don't see it as a big deal for dev dependencies.

Regardless, I'm gonna check poetry docs if there's something more flexible than `^`.

bruno-fs: maybe `^0` or `~0`, which would be equivalent to >=0.0.0<1.0.0

thoughts?


see https://python-poetry.org/docs/dependency-specification/

infinitewarp: That would work!

infinitewarp (approved):

PR #2632, opened 2024-04-22
fix: fix a bug in fingerprinting when systems have no facts

bruno-fs:
Fix the problem by reproducing the original behavior of not exposing systems with no facts on Report.sources.

### DIFF
--- a/quipucords/api/inspectresult/model.py
+++ b/quipucords/api/inspectresult/model.py
@@ -14,47 +14,47 @@ from constants import DataSources
 
 RAW_QUERY_POSTGRES = """
 with raw_agg(ir_id, inspect_group_id, raw_facts) as (
-	select
-		ir.id,
+    select
+        ir.id,
         ir.inspect_group_id,
         json_object_agg(raw.name, raw.value) as raw_facts
-	from
-		api_inspectresult ir
-	inner join
-		api_rawfact raw on ir.id = raw.inspect_result_id
+    from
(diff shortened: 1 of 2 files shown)


quipucords/api/inspectresult/model.py
```diff
@@ -13,20 +13,17 @@
 from constants import DataSources
 
 RAW_QUERY_POSTGRES = """
-with raw_agg as (
-    select
-        ir.id,
+with raw_agg(ir_id, inspect_group_id, raw_facts) as (
+	select
+		ir.id,
```
infinitewarp, line 18: I know this is a raw string being passed to SQL that doesn't follow Python coding conventions, but may we choose spaces instead of tab characters for these for sake of consistency? (I ran into this exact same line earlier today and was meaning to ask about this anyway!)

```suggestion
    select
        ir.id,
```

infinitewarp (approved):

quipucords/api/inspectresult/model.py
bruno-fs: wait, I'm using tab here? lol
fixing. thanks for spotting it

infinitewarp: Also the lines below that use raw tab characters. Using spaces makes it easier to copy-paste into a DB shell when manually debugging code too since tabs can trigger other behaviors like autocomplete.

PR #2634, opened 2024-04-25
refactor(upload-reports)!: async workflow + celery support

bruno-fs:
refactor(upload-reports): refactor upload report workflow to work asynchronously

One may argue this should be also bumping minor version as this introduces breaking changes.
Changes on CLI side are required (quipucords/qpc/pull/326).

Relates to JIRA: DISCOVERY-452

## Summary of changes
- refactor `POST /report/v1/` to play nice with celery
- deprecate `/report/v1/merge/job/*` endpoints
  - replace those with a single endpoint `POST /report/v1/merge/`. It behaves similarly to the old `PUT /report/v1/merge/job/`, the changes are internal only - it now will reuse internal raw facts instead of duplicating them.
  - the deprecated endpoints were all duplicated functionality; GET was the same as GET from `jobs` endpoint; POST was the same as `upload` endpoint.
  
I originally intended to make those changes in the v2 API namespace. However, I decided to only move these endpoints to v2 when we finally have a new format for "details report" / "raw facts".


quipucords/api/details_report/serializer.py
```diff
@@ -1,6 +1,11 @@
 """Serializer for system facts models."""
 
-from rest_framework.serializers import CharField, IntegerField, JSONField, UUIDField
+from rest_framework.serializers import (
+    CharField,
+    IntegerField,
+    JSONField,
+    UUIDField,
+)
```
infinitewarp, line 8: What touched this line? ruff seemed happy with the previous version.

quipucords/api/details_report/util.py
```diff
@@ -13,13 +13,8 @@
     sanitize_row,
 )
 from api.models import (
-    InspectGroup,
-    InspectResult,
-    RawFact,
     Report,
```
infinitewarp, line 16: Trivial nitpick but…
```suggestion
    Report
```
and reformat to lose the parens and line breaks.
(github refused to take a multiline suggestion here 🤷)

quipucords/api/details_report/serializer.py
bruno-fs: 🤔 During development I added some code here and then removed. I guess black (I was originally using black) formatted the extra dependencies like this and when removed left it in this format. Will adjust it

quipucords/tests/scanner/network/test_network_inspect.py
```diff
@@ -86,7 +85,7 @@ def setup_method(self, _test_method):
 
         self.scan_task.update_stats("TEST NETWORK INSPECT.", sys_failed=0)
 
-        self.fact_endpoint = "http://testserver" + reverse("v1:reports-list")
+        self.fact_endpoint = "http://testserver/whatever"
```
infinitewarp, line 88: What even is the point of `self.fact_endpoint`? 🤔  I see it's used once for a mock call in `test_inspect_scan_fail_no_facts`, but I don't understand what actual purpose it serves there. I commented out the line and ran the test locally, and it still passed without it…

I know question is growing beyond the scope of your diff, but it might mean deleting a couple more lines, and we like that, right? 🙂

bruno-fs: I have the same question.
No actual endpoint is being called there from what I can tell. I changed to whatever to highlight how useless at least one mock is in that test.

If I wasn't too lazy to try to understand that test, I'd have a better answer. My gut feeling was to actually axe it, but I can't do that without understanding it first 😅

infinitewarp (approved):

PR #2639, opened 2024-05-01
feat: add created_at and updated_at fields to all models, populated automatically

infinitewarp:
☝️ That's me every time I look at our objects and I'm clueless about when they were created or modified.

This PR adds `created_at` and `updated_at` fields to all our models and sets their values automatically; we should never force specific values into those fields. I believe we are long overdue for this basic model functionality, especially since we will soon expect our users to run the same server for long stretches of time potentially over many generations of credentials, sources, scans, and reports.

Fortunately, the changes to surrounding logic outside the models and migration are very minimal. 🎉

~~The only issue I ran into was one test where two returned API responses have different `updated_at` times, off only by a small fraction of a second, which makes me believe that there is some unexpected save happening between the calls. I added a `TODO FIXME` comment in the code in the hopes that we may investigate, find, and remove the presumed extra save call later. I wonder if it's related to this other older comment in `get_serialized_report_data`, but I didn't spend any more time digging into it:~~
```
    # We need to refresh from DB because something about how the factory created the
    # Report object here can cause flaky, unexpected failures in the serializer.
    # Maybe there's a bug in the factories or how we're using them?
```
🤷 

Update: I have explicitly excluded the new `created_at` and `updated_at` fields from the serializers that were just including all field. This should make this change purely internal-only for now with no externally visible API changes.


quipucords/tests/api/reports/test_reports.py
```diff
@@ -168,13 +168,22 @@ def test_report_tarball_details_json(
...
     assert api_details_response.ok, api_details_response.text
     api_details_json = api_details_response.json()
+
+    # TODO FIXME What is causing updated_at to be different in these responses?
+    # Upon initial investigation, the created_at times in the responses match,
+    # but api_details_json's updated_at is more recent than tarball_details_json's.
+    # The difference is on the order of microseconds which means me to believe that
+    # something is triggering an unexpected save between the first and second get.
+    del tarball_details_json["updated_at"]
+    del api_details_json["updated_at"]
```
bruno-fs, line 178: the csv generation is the root cause here. If the csv is not stored in the model, it will be generated on the fly.

when we ask for the details report, we ask only for json, so no csv is created (and even if we asked, there's a "bug" here, because the datetimes will be serialized before the model is updated).

finally, when we ask for the tarball, the csv will be created and the model will be updated.
or at least this is the most likely explanation that comes up here.

bruno-fs (approved):

PR #2641, opened 2024-05-03
refactor error handling in network scan

bruno-fs:
Fix a vulnerability found in network scan that could cause a whole scan to fail if a failure occurs in a single "inspection group".

This is related to [DISCOVERY-353](https://issues.redhat.com/browse/DISCOVERY-353) and could also be related to   [DISCOVERY-464](https://issues.redhat.com/browse/DISCOVERY-464) 

Relates to JIRA: DISCOVERY-353

### DIFF
--- a/quipucords/scanner/network/inspect_callback.py
+++ b/quipucords/scanner/network/inspect_callback.py
@@ -59,7 +59,7 @@ class InspectCallback:
         host = event_data.get("host", UNKNOWN_HOST)
         task = event_data.get("task")
         role = event_data.get("role")
-        logger.info("[host=%s] processing role='%s' task='%s'", host, role, task)
+        logger.debug("[host=%s] processing role='%s' task='%s'", host, role, task)
         self._process_task_facts(ansible_facts, host)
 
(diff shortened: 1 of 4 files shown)


quipucords/scanner/network/inspect_callback.py
```diff
@@ -59,7 +59,7 @@ def task_on_ok(self, event_dict: dict):
         host = event_data.get("host", UNKNOWN_HOST)
         task = event_data.get("task")
         role = event_data.get("role")
-        logger.info("[host=%s] processing role='%s' task='%s'", host, role, task)
+        logger.debug("[host=%s] processing role='%s' task='%s'", host, role, task)
```
infinitewarp, line 62: Switching to debug because this is too noisy? This will skip output in customer logs in the default server configuration, right?

infinitewarp (approved):

quipucords/scanner/network/inspect_callback.py
bruno-fs: mmhmmm. The recent logs I got from the field were the first ones with this new log and I just added noise.

PR #2642, opened 2024-05-03
feat: add aggregate report API

infinitewarp:
Relates to JIRA: DISCOVERY-611

This is a first pass at what Justin needs. On 2024-05-02, I walked with him through the top 42 rows in [his request spreadsheet](https://docs.google.com/spreadsheets/d/1YvI7KQc7ZyBtoDqj8y0kyo5PU-NerkAuoAhzLI6pDmE/edit#gid=0), explained exactly how I determine each output value, and resolved several questions about ambiguous inputs and outputs. Following that discussion, I kept some inline comments with the code to clarify a few potentially confusing parts.

The current version is still far from optimal, but it technically works. This API looks up the report fingerprints and facts and calculates the output _on demand_ every time. This obviously won't scale up to large data sets. So, a logical next step (in a future PR, not this one) would be to store the aggregate report in a real Django model, calculate it in another phase following fingerprinting, and return that stored object upon request. However, I wanted to put some eyes on the current structure and general approach _before_ taking it to that next level.

On this branch, you can call the API to generate an aggregate report like this example which was produced using data from real-world scans:

```bash
http http://127.0.0.1:8000/api/v1/reports/32770/aggregate/ "Cookie:csrftoken=$CSRFTOKEN; sessionid=$SESSIONID" "X-CSRFToken:$CSRFTOKEN"
```
```json
{
    "ansible_hosts_all": 1,
    "ansible_hosts_in_database": 1,
    "ansible_hosts_in_jobs": 0,
    "instances_hypervisor": 0,
    "instances_physical": 0,
    "instances_unknown": 91,
    "instances_virtual": 187,
    "jboss_eap_cores_physical": 0,
    "jboss_eap_cores_virtual": 0,
    "jboss_eap_instances": 0,
    "jboss_ws_cores_physical": 0,
    "jboss_ws_cores_virtual": 0,
    "jboss_ws_instances": 0,
    "missing_cpu_core_count": 89,
    "missing_cpu_socket_count": 273,
    "missing_host_done": 7,
    "missing_name": 1,
    "missing_pem_files": 0,
    "missing_system_creation_date": 267,
    "missing_system_purpose": 1,
    "openshift_cluster_instances": 1,
    "openshift_cores": 24,
    "openshift_node_instances": 6,
    "openshift_operators_by_kind": {
        "cluster-operator": 36,
        "olm-operator": 1
    },
    "os_by_name_and_version": {
        "RHEL": {
            "8.9": 18
        },
        "Red Hat Enterprise Linux": {
            "9.0 (Plow)": 2,
            "9.2 (Plow)": 3
        },
        "RedHat": {
            "7.9": 17,
            "8.1": 14,
            "9.3": 17,
            "9.4": 19,
            "9.5": 1
        },
        "unknown": {
            "4.4.4": 1,
            "unknown": 223
        }
    },
    "socket_pairs": 8,
    "system_creation_date_average": "2024-03-18",
    "vmware_hosts": 4,
    "vmware_vm_to_host_ratio": 2.75,
    "vmware_vms": 11
}
```

### DIFF
--- a/quipucords/api/aggregate_report/model.py
+++ b/quipucords/api/aggregate_report/model.py
@@ -1,0 +1,327 @@
+"""
+Model definition and calculation logic for the aggregate report.
+
+The logic here is kind of messy because we do not surface all the necessary
+details into individual fields in the normalized/de-duped SystemFingerprint model.
+To minimize the number of iterations and trips back to the database, we iterate
+through all SystemFingerprint rows once and conditionally tally various fields,
+and then we iterate through all RawFact rows (via InspectGroup) and conditionally
+tally various fields that only exist there. This iteration logic could be
+simplified to use Django query aggregates if we stored more individual fields in
+SystemFingerprint or some other even flatter data structure that DOES NOT use
+JSONField-like fields that require additional special logic. Until that exists,
+the logic here will remain somewhat long and complex.
+
+Much of this logic could be made more readable by naively breaking up individual
+fields into their own "sum(map(lambda ...))" one-liners or multiple smaller Django
+aggregate queries. However, that would mean iterating over the rows multiple times,
+and that would likely cause execution time to grow linearly with the number of hosts
+and facts, and that may get too costly too quickly. The current ugly implementation
+iterates through all the facts exactly once to minimize expected execution time.
+"""
+
+import logging
+from collections import defaultdict
+from collections.abc import Iterable
+from dataclasses import asdict, dataclass, field
+from datetime import date
+from math import ceil
+
+from api.deployments_report.model import Product, SystemFingerprint
+from api.inspectresult.model import InspectResult
+from api.reports.model import Report
+from api.scan.model import Scan
+from constants import DataSources
+from utils.datetime import average_date
+
+logger = logging.getLogger(__name__)
+
+UNKNOWN: str = "unknown"  # placeholder string for missing names/versions/kinds.
+
+
+@dataclass
+class AggregateReport:
+    """Results of the aggregate report."""
+
+    # Note the lack of special reporting needs for Satellite or RHACS here.
+    # At the time of this writing, Justin says this is correct and we have none.
+
+    # TODO Convert this to a full Django model that we can store and look up.
+
+    ansible_hosts_all: int = 0
+    ansible_hosts_in_database: int = 0
+    ansible_hosts_in_jobs: int = 0
+    inspect_result_status_failed: int = 0
+    inspect_result_status_success: int = 0
+    inspect_result_status_unknown: int = 0
+    inspect_result_status_unreachable: int = 0
+    instances_hypervisor: int = 0
+    instances_physical: int = 0
+    instances_unknown: int = 0
+    instances_virtual: int = 0
+    jboss_eap_cores_physical: int = 0
+    jboss_eap_cores_virtual: int = 0
+    jboss_eap_instances: int = 0
+    jboss_ws_cores_physical: int = 0
+    jboss_ws_cores_virtual: int = 0
+    jboss_ws_instances: int = 0
+    missing_cpu_core_count: int = 0
+    missing_cpu_socket_count: int = 0
+    missing_name: int = 0
+    missing_pem_files: int = 0
+    missing_system_creation_date: int = 0
+    missing_system_purpose: int = 0
+    openshift_cores: int = 0
+    openshift_operators_by_kind: dict = field(default_factory=dict)
+    os_by_name_and_version: dict = field(default_factory=dict)
+    socket_pairs: int = 0
+    system_creation_date_average: date | None = None
+    vmware_hosts: int = 0
+    vmware_vm_to_host_ratio: float = 0
+    vmware_vms: int = 0
+
+    # Note: The following attributes come exclusively from raw facts.
+    # For now, that's our only option, and these are required outputs.
+    # TODO Refactor how we populate these when we restructure the underlying data.
+    openshift_cluster_instances: int = 0
+    openshift_node_instances: int = 0
+
+
+def get_aggregate_report_by_report_id(report_id: int) -> dict:
+    """
+    Get the aggregate report data for the given report ID.
+
+    TODO Turn this into a database lookup after we start storing AggregateReport.
+    """
+    report = Report.objects.get(pk=report_id)
+    return asdict(build_aggregate_report(report.id))
+
+
+def _aggregate_from_system_fingerprints(  # noqa: C901,PLR0912,PLR0915
+    aggregated: AggregateReport, fingerprints: Iterable[SystemFingerprint]
+) -> None:
+    """
+    Populate the AggregateReport object with data from the given SystemFingerprints.
+
+    TODO Try to break up some to this to reduce the noqa: C901,PLR0912,PLR0915.
+    """
+    os_by_name_and_version = defaultdict(lambda: defaultdict(int))
+    system_creation_dates = []
+    vmware_vms_by_host = defaultdict(list)
+
+    for fingerprint in fingerprints:
+        # Yes, we want *all* OS name/version combos regardless of `is_redhat`.
+        # This means output may include non-RHEL distros (e.g. Ubuntu, Debian).
+        # Note: Ansible sources are fingerprinted to store the Ansible controller
+        # version (e.g. "4.4.4") in SystemFingerprint.os_version, and that can
+        # results strange in output here like {"unknown": {"4.4.4": 1}}.
+        # See also: FingerprintTaskRunner._process_ansible_fact.
+        # Those versions may look weird, but Justin says to keep them all.
+        os_by_name_and_version[fingerprint.os_name or UNKNOWN][
+            fingerprint.os_version or UNKNOWN
+        ] += 1
+
+        source_types = [source.get("source_type") for source in fingerprint.sources]
+        has_network_source = DataSources.NETWORK in source_types
+        has_satellite_source = DataSources.SATELLITE in source_types
+        has_vcenter_source = DataSources.VCENTER in source_types
+        has_openshift_source = DataSources.OPENSHIFT in source_types
+
+        if not (fingerprint.is_redhat or has_openshift_source or has_satellite_source):
+            # For all other fingerprint facts we aggregate below here, we only want
+            # to count them if we positively identified Red Hat during fingerprinting.
+            # We treat everything found by OpenShift or Satellite to be "is_redhat=True"
+            # here even if the "is_redhat" attribute was not set on the fingerprint.
+            continue
+
+        if not fingerprint.cpu_core_count and not has_vcenter_source:
+            # Do not penalize vCenter sources because they do not collect core counts.
+            aggregated.missing_cpu_core_count += 1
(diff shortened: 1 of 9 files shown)


quipucords/api/aggregate_report/view.py
```diff
@@ -0,0 +1,19 @@
...
+from api.common.util import is_int
+
+
+@api_view(["GET"])
+def aggregate_report(request, report_id=None):
+    """Lookup and return a details system report."""
+    if report_id is not None:
+        if not is_int(report_id):
+            error = {"report_id": [messages.COMMON_ID_INV]}
+            raise ValidationError(error)
```
bruno-fs, line 18: I know we have many code snippets like this, but they are useless.
Given how the URL is built, `report_id` will always be present and will always be an `int`. 
If it isn't, this code won't be reached and the user will likely get a 404 response.

IoW, I'd drop this code.

quipucords/api/aggregate_report/model.py
```diff
@@ -0,0 +1,275 @@
...
+            fingerprint.os_version or "unknown"
+        ] += 1
+
+        # TODO Is there a better way to determine the underlying source type?
+        # Is there always at least 1 fingerprint.sources? are they always the same type?
+        # Should I be doing something more clever like iterating through the list
+        # and checking for any presence of each source type?
+        # Could there be multiple different source types in the same fingerprint?
+        # If I made a bad assumption here, that might invalidate everything else below.
+        source_type = fingerprint.sources[0].get("source_type")
```
bruno-fs, line 120: you shouldn't expect a single source type from a fingerprint. 
Remember deduplication?
A fingerprint could be generated by multiple source types, so, if we are to follow with this approach, we need to have support for multiple source types - otherwise, we are just choosing one "randomly".

quipucords/api/aggregate_report/model.py
```diff
@@ -0,0 +1,275 @@
...
+            in (DataSources.NETWORK, DataSources.SATELLITE, DataSources.VCENTER)
+            and not fingerprint.is_redhat
+        ):
+            # For all other fingerprint facts we aggregate below here, we only want
+            # to count them if we positively identified RHEL during fingerprinting.
+            continue
+
+        if not fingerprint.cpu_core_count and source_type != DataSources.VCENTER:
+            # Do not penalize vCenter sources because they do not collect core counts.
+            aggregated.missing_cpu_core_count += 1
```
bruno-fs, line 133: this logic shall only be used IF we are talking about fingerprints from vcenter not deduplicated with other sources.

quipucords/api/aggregate_report/model.py
```diff
@@ -0,0 +1,275 @@
...
+
+        cpu_core_count = fingerprint.cpu_core_count or 0
+        jboss_presence = {
+            product.name: bool(product.presence == Product.PRESENT)
+            for product in fingerprint.products.all()
+            if product.name in jboss_product_names
+        }
+        # For now (forever?) we do not care to count FUSE or BRMS.
+        # If we don't care about them, then why bother collecting the data?
+        # TODO Remove *all* other API/model/etc. code related to FUSE and BRMS?
```
bruno-fs, line 154: This is a good PM question.

AFAIK, BRMS can be safely removed - we just never got the thumbs up for doing so. As for FUSE, I wish we could have a YES or NO answer for removing it. FUSE isn't EOL yet, but will soon be. Not sure if it will have extended support...

quipucords/api/aggregate_report/model.py
```diff
@@ -0,0 +1,275 @@
...
+            # Some products are sold specifically with this per-two-CPU-sockets math.
+            # Try not to think too hard about this.
+            aggregated.socket_pairs += ceil(fingerprint.cpu_socket_count / 2)
+
+        if source_type == DataSources.NETWORK:
+            if not fingerprint.system_purpose:
+                aggregated.missing_system_purpose += 1
+            if not fingerprint.redhat_certs:
+                aggregated.missing_pem_files += 1
+        elif source_type == DataSources.OPENSHIFT:
```
bruno-fs, line 186: Jeez, the way these requirements are changing is bizarre. So now we want Openshift as part of MVP?

bruno-fs: (asking because the original MVP was something in the lines of "parity with the project that shall not me mentioned)

quipucords/api/aggregate_report/model.py
```diff
@@ -0,0 +1,275 @@
...
+
+    ansible_hosts_in_database = []
+    ansible_hosts_in_jobs = []
+    openshift_operators_by_kind = defaultdict(int)
+
+    # TODO Should we query RawFacts directly instead of using report.sources?
+    # Would that be faster than going through the InspectGroupQuerySet? We would still
+    # need to decode the RawFact.value JSON, but we could filter to only get only the
+    # RawFact.name matches we actually need here, and decoding their .value JSON might
+    # be faster than InspectGroupQuerySet decoding *all* of the rows.
```
bruno-fs, line 229: my gut feeling says it will be faster. and it would be desirable for when dealing with huge reports.

Also, I implemented `Report.sources` just as a lazy way to keep compatibility with the old field. I'd avoid making it as a dependency for new code if possible.

quipucords/api/aggregate_report/model.py
```diff
@@ -0,0 +1,275 @@
...
+        # business count usage by simple hostnames.
+        vmware_vm_count = sum(len(set(items)) for items in vmware_vms_by_host.values())
+        vmware_host_count = len(vmware_vms_by_host)
+        aggregated.vmware_vms = vmware_vm_count
+        aggregated.vmware_vm_to_host_ratio = vmware_vm_count / vmware_host_count
+        aggregated.vmware_hosts = vmware_host_count
+
+    #####
+    # Aggregate more data from raw facts, not fingerprints. Hic sunt dracones!
+    # Can any of these actually be determined by fields in SystemFingerprint?
```
bruno-fs, line 218: nope

quipucords/api/aggregate_report/model.py
```diff
@@ -0,0 +1,275 @@
...
+    # Should we normalize and store some of these as fields in SystemFingerprint?
```
bruno-fs, line 219: We could. The 2yrs old problem here is: data from those sources won't fit SystemFingerprint. I say we make this part of the "shiny new normalization", and as part of the process we create the proper models to hold their data.

Openshift deserves normalization. Ansible and RHACS I'm not so sure.

I'd even say: remove normalization as a requirement for discovery. If a certain source has no need for it, don't do it. Of course we could also create specialized models just for the normalized result of those sources.

quipucords/api/aggregate_report/model.py
```diff
@@ -0,0 +1,291 @@
...
+        # Those versions may look weird, but Justin says to keep them all.
+        os_by_name_and_version[fingerprint.os_name or UNKNOWN][
+            fingerprint.os_version or UNKNOWN
+        ] += 1
+
+        source_types = [source.get("source_type") for source in fingerprint.sources]
+        has_network_source = DataSources.NETWORK in source_types
+        has_satellite_source = DataSources.SATELLITE in source_types
+        has_vcenter_source = DataSources.VCENTER in source_types
+        has_openshift_source = DataSources.OPENSHIFT in source_types
```
bruno-fs, line 124: `:thisistheway:`

quipucords/api/aggregate_report/model.py
```diff
@@ -0,0 +1,291 @@
...
+
+        if not (fingerprint.is_redhat or has_openshift_source or has_satellite_source):
+            # For all other fingerprint facts we aggregate below here, we only want
+            # to count them if we positively identified Red Hat during fingerprinting.
+            # We treat everything found by OpenShift or Satellite to be "is_redhat=True"
+            # here even if the "is_redhat" attribute was not set on the fingerprint.
+            continue
```
bruno-fs, line 131: not something we can do about it in code, but we can get lots of false positives and false negatives with vcenter here.

quipucords/api/aggregate_report/model.py
```diff
@@ -0,0 +1,291 @@
...
+    # need to decode the RawFact.value JSON, but we could filter to only get only the
+    # RawFact.name matches we actually need here, and decoding their .value JSON might
+    # be faster than InspectGroupQuerySet decoding *all* of the rows.
+    inspect_groups = report.sources.all()
+    for inspect_group in inspect_groups:
+        for raw_fact in inspect_group["facts"]:
+            source_type = inspect_group.get("source_type")
+            if source_type == DataSources.NETWORK:
+                if not raw_fact.get("host_done", False):
+                    aggregated.missing_host_done += 1
```
bruno-fs, line 242: I kinda regret revealing to @justinkreft what host done means.
While we can use this metric, I'd rather use the more generic `InspectResult.status` and count total failures or unreachable. That should be the proper way to deal with this.

Alternatively, use the counts present on `Task` with type="inspection"

quipucords/tests/api/aggregate_report/test_aggregate_report_model.py
```diff
@@ -0,0 +1,301 @@
...
+    # virtual RHEL 9.1 w/ JBoss EAP
+    fingerprint = SystemFingerprintFactory(
+        os_version="9.1",
+        infrastructure_type=SystemFingerprint.VIRTUALIZED,  # instances_virtual++
+        deployment_report=deployments_report,
+        cpu_count=None,  # cpu_count is only relevant for openshift sources
+        cpu_core_count=2,
+        cpu_socket_count=1,  # socket_pairs += 1
+        system_purpose={"These": "values"},
+        redhat_certs=["do", "not", "matter."],
```
bruno-fs, line 60: 🤣

quipucords/tests/api/aggregate_report/test_aggregate_report_model.py
```diff
@@ -0,0 +1,301 @@
...
+    expected_aggregate.openshift_cores += 32
+    expected_aggregate.missing_system_creation_date += 1
+    expected_aggregate.os_by_name_and_version[UNKNOWN]["NCC-1701"] = 1
+
+    # Not-RHEL system found by network scan
+    # Unlike the openshift one above, this one DOES NOT increment counters,
+    # but we do track its OS name and version anyway.
+    SystemFingerprintFactory(
+        os_name="LCARS",
+        os_version="NCC-1701-D",
```
bruno-fs, line 161: <3

bruno-fs (commented):
My only problem here is not with code, but with the business requirement for using `missing_host_done` metric.

@justinkreft what are you trying to solve with this? I believe engineering can give a better answer than that.

quipucords/api/aggregate_report/model.py
infinitewarp: We finally discussed this issue during backlog refinement today, and the team with Justin's blessing agreed that we _**SHOULD NOT**_ be using `host_done` in report outputs like this. 🎉  I will change this code to use `InspectResult.status` instead.

Thank you for persistently asking about this. 🙂  The `status` field will be a much more reliable and future-proof option here, and I was only using `host_done` because of the prior insistence that _specifically it_ must be included in the output.

PR #2643, opened 2024-05-06
fix: better handle some unexpected None values

infinitewarp:
Some possible "fixes" inspired by @bruno-fs's recent findings [documented here](https://docs.google.com/document/d/1wwZDHXa1dD_0UeP-fWg1lLR7MeIT5GpzUUdrmWTjrSk/edit).

These changes do not fix the underlying problems of missing data. They only improve handling of bad data so we don't crash out with unhandled exceptions.

Also, this diff grew a lot because I needed to touch `test_scantask` but I also needed to convert it from unittest style to pytest. If might help to review the commits one at a time.

### DIFF
--- a/quipucords/tests/api/scantask/test_scantask.py
+++ b/quipucords/tests/api/scantask/test_scantask.py
@@ -1,11 +1,9 @@
-"""Test the API application."""
+"""Test the ScanTask model and serializer."""
 
 from datetime import datetime
 from unittest.mock import patch
 
 import pytest
-from django.core import management
-from django.test import TestCase
 
 from api import messages
 from api.inspectresult.model import InspectResult, RawFact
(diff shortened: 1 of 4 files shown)


bruno-fs (approved):

quipucords/tests/api/scantask/test_scantask.py
```diff
@@ -47,202 +48,195 @@ def scan_job(scan: Scan) -> ScanJob:
...
+        [
+            ScanTask.status_restart,
+            ScanTask.PENDING,
+            messages.ST_STATUS_MSG_RESTARTED,
+            False,
+            False,
+        ],
+        [
+            ScanTask.status_pause,
+            ScanTask.PAUSED,
```
bruno-fs, line 96: mmmm we still need some PAUSE/RESTART cleanup.

(not that this needs to be done in this PR. just thinking out loud...)

quipucords/tests/api/scantask/test_scantask.py
```diff
@@ -47,202 +48,195 @@ def scan_job(scan: Scan) -> ScanJob:
...
-        task_instance_b = ScanTask.objects.get(id=task_instance_a.id)
+    assert 0 == task_instance_a.systems_count
+    assert 0 == task_instance_b.systems_count
 
-        assert 0 == task_instance_a.systems_count
-        assert 0 == task_instance_b.systems_count
+    with (
+        patch.object(task_instance_a, "refresh_from_db"),
+        patch.object(task_instance_b, "refresh_from_db"),
+    ):
```
bruno-fs, line 205: I can see the original code had this patch, but do you know why? seems unnecessary 🤔

bruno-fs (approved):

quipucords/tests/api/scantask/test_scantask.py
infinitewarp: I thought about that when I saw this, too, but I didn't want to scope creep here today. The old tests tested pause/restart. So, the new tests do too.

quipucords/tests/api/scantask/test_scantask.py
infinitewarp: No clue. I didn't dig into that, but let me see what happens if I yank them out…

infinitewarp: Oh. Well, there's this old docstring above it:

>    This is a special case test to help ensure that increment_stats is thread-safe.
>    We patch `refresh_from_db` with a mock object here because we don't want the
>    underlying functionality to rely on `refresh_from_db` completing immediately
>    before the update/save.

The test still passes when I commented out the patches and deindented the code it wrapped, but I'll just leave it like this. I don't feel like this is a _good_ test, but I'll dodge and say I'm trying to minimize scope creep. 🤷

infinitewarp: LOL I thought that writing style looked familiar. Guess who is on the blame for the original lines? 😅 https://github.com/quipucords/quipucords/commit/c21e53f6df4afe2bf3cde90f092abe0099d97c71

bruno-fs: lol

this caught my attention because, even though mocked, we still call the method anyways. Which is extremely counter intuitive to me. Looks like we are doing nothing when calling `refresh_from_db` later, so why do it?

infinitewarp: OH! I think I understand what I intended when I wrote that, and I think it still makes sense. `increment_stats` internally calls `refresh_from_db`. This test tries to simulate conditions of parallelism without actually running things in parallel. If you have two different instances of an object that both point to the same underlying database record, and you call `increment_stats` of each of them independently but around the same time, the underlying stats counter in the DB _should_ go from `0` to `2`. The code that existed _before_ I introduced a fix with that test was not thread-safe for cases like this, and the counter in the DB would go from `0` to only `1`, not the correct `2`.

The reason I mocked out `refresh_from_db` was to guarantee that the instances don't refresh their local copies of the underlying data until they've both completed their saves. If the second instance had refreshed its data before saving, then the simulated parallelism conditions would be invalid as the second might have "fixed" its in-memory counter to `1` when in a real parallel call it would still be `0`.

So, I believe the test should remain. I hope that clears it up a bit. 🙂

PR #2647, opened 2024-05-07
refactor: move jboss roles to the end because they seem to be problematic…

infinitewarp:
…, and we want to minimize their disruption to other roles

Relates to JIRA: DISCOVERY-676
https://issues.redhat.com/browse/DISCOVERY-676

### DIFF
--- a/quipucords/scanner/network/runner/inspect.yml
+++ b/quipucords/scanner/network/runner/inspect.yml
@@ -14,11 +14,6 @@
     - cloud_provider
     - etc_release
     - file_contents
-    - jboss_eap
-    - jboss_eap5
-    - jboss_fuse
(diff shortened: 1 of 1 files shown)


bruno-fs:
Don't need to be done in this PR, but I'd like to shuffle a few more things.

Move the following roles to the bottom as well, right before middleware:
- user_data (actually, I'd like to 🪓 this, but it can wait)
- redhat_packages
- file_contents

infinitewarp:
Closing because these changes already landed in `release/1.7` via https://github.com/quipucords/quipucords/pull/2648 and will be ported to `main` via https://github.com/quipucords/quipucords/pull/2653.

PR #2654, opened 2024-05-08
refactor: upgrade to pyvmomi "^8.0.2"

infinitewarp:
This change also shrinks some needlessly complex code in our `vcenter_connect` function, converts an old unittest test to pytest, and expands that test to cover some previously untested conditions.


quipucords/scanner/vcenter/utils.py
```diff
@@ -15,41 +15,30 @@ def vcenter_connect(scan_task):
...
-    elif ssl_protocol is None and ssl_cert_verify is None:
-        vcenter = SmartConnect(host=host, user=user, pwd=password, port=port)
-    else:
-        ssl_context = None
-        if ssl_protocol is None:
-            ssl_context = ssl.SSLContext(protocol=ssl.PROTOCOL_SSLv23)
-        else:
-            ssl_context = ssl.SSLContext(protocol=ssl_protocol)
+        smart_connect_kwargs["disableSslCertValidation"] = True
+    elif not (ssl_protocol is None and ssl_cert_verify is None):
```
abellotti, line 33: would this be the same?  `elif ssl_protocol is True or ssl_vert_verify is True:`  

just a little easier to read.

infinitewarp: That might be easier to read, yes, but it's not logically equivalent to the original logic which specifically checks `is None`, and it looks like these values really could be truthy, falsy, or `None` under different circumstances. The underlying `Source` model where these values came from has both fields defined with `null=True`, and I'm assuming there is some reason for differentiating `False` (`ssl_vert_verify` is a bool) and "empty string" (`ssl_protocol` is a string) from `None`. 😕 

Although I rearranged the code here, my intent was to preserve all of the original logic, whether I liked and agreed with it or not.

abellotti: Gotcha, ya, confusing I didn't set that value vs I set that value to False case, ugg.

Thanks for the recap, when my brains sees not(false and false and false), I always think the other way :)

abellotti (approved):

PR #2659, opened 2024-05-13
feature(openshift-scanner): include unschedulable fact

bruno-fs:
Relates to JIRA: DISCOVERY-623

### DIFF
--- a/quipucords/scanner/openshift/api.py
+++ b/quipucords/scanner/openshift/api.py
@@ -362,6 +362,10 @@ class OpenShiftApi:
         return self._route_api.get(**kwargs).items
 
     def _init_ocp_nodes(self, node) -> OCPNode:
+        # following upstream docs[1], if unschedulable is None, it is considered
+        # false.
+        # [1]: https://kubernetes.io/docs/reference/kubernetes-api/cluster-resources/node-v1/#NodeSpec
+        unschedulable = node.spec["unschedulable"] or False
         return OCPNode(
             name=node.metadata.name,
             creation_timestamp=node.metadata.creationTimestamp,
@@ -382,6 +386,7 @@ class OpenShiftApi:
             machine_id=node.status["nodeInfo"]["machineID"],
             operating_system=node.status["nodeInfo"]["operatingSystem"],
             taints=node.spec["taints"],
+            unschedulable=unschedulable,
         )
 
     def _init_cluster(self, cluster) -> OCPCluster:
--- a/quipucords/scanner/openshift/entities.py
+++ b/quipucords/scanner/openshift/entities.py
@@ -95,6 +95,7 @@ class OCPNode(OCPBaseEntity):
     operating_system: str = None
     taints: List[dict] = None
     errors: Dict[str, OCPError] = Field(default_factory=dict)
+    unschedulable: bool = None
     _kind = "node"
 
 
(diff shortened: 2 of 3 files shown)


quipucords/scanner/openshift/api.py
```diff
@@ -362,6 +362,13 @@ def _list_routes(self, **kwargs):
         return self._route_api.get(**kwargs).items
 
     def _init_ocp_nodes(self, node) -> OCPNode:
+        unschedulable = node.spec["unschedulable"]
+        if unschedulable is None:
+            # following upstream docs[1], if unschedulable is None, it is considered
+            # false.
+            # [1]: https://kubernetes.io/docs/reference/kubernetes-api/cluster-resources/node-v1/#NodeSpec
+            unschedulable = False
```
infinitewarp, line 370: How much do we trust the `spec` dict to be well-structured?
```suggestion
        # following upstream docs[1], if unschedulable is None, it is considered false.
        # [1]: https://kubernetes.io/docs/reference/kubernetes-api/cluster-resources/node-v1/#NodeSpec
        unschedulable = node.spec.get("unschedulable", False) or False
```

quipucords/scanner/openshift/entities.py
```diff
@@ -95,6 +95,7 @@ class OCPNode(OCPBaseEntity):
     operating_system: str = None
     taints: List[dict] = None
     errors: Dict[str, OCPError] = Field(default_factory=dict)
+    unschedulable: bool = None
```
infinitewarp, line 98:
```suggestion
    unschedulable: bool = False
```

Following the previous comment you cited, I think `False` is a better default than `None`.

quipucords/scanner/openshift/api.py
bruno-fs: spec isn't a dictionary. This is a magical `ResourceField`. which is a weird magical object that always returns keys, even when they don't exist.

If I ask node["foo"]["bar"] it will just return `None` 🤢 . This implies that if "unschedulable" isn't set (which is the default), `node.spec.get("unschedulable", False) ` would be equal to `None` 🤷🏽 .

TBH, I hate it. As I hate the lib we use for interacting with OCP (just adopted it because interacting directly with the API would take way more time to implement). While this kind of "smart pants" object can be neat to use, it is a huge pain to test (see the disgusting workarounds to create the object to test this).

Part of the reason I adopted vcr to test the integration with ocp was this type of thing.

quipucords/scanner/openshift/entities.py
bruno-fs: I could, but what if the object was initialized without this info. Could we infer it is indeed unschedulable?

Business logic for OCP scanner are concentrated in inspect and api modules.

quipucords/scanner/openshift/api.py
bruno-fs: I changed it anyways. See the fixup commit.
However, if we don't trust Openshift/kubernetes response, we can't properly infer `unschedulable`.

But if we go that far and don't trust the response, maybe we should not trust `spec` will be present as well.

At that point I don't see a distinction between this line of flexibility and network scan `ignore_errors: true` on all ansible tasks.

infinitewarp (approved):

infinitewarp (approved):

PR #2661, opened 2024-05-17
fix: improve creation_timestamp in _openshift_raw_facts_generator

infinitewarp:
Without `tzinfo`, the generated `creation_timestamp` will fail fingerprint processing because `fingerprinter.runner.RAW_DATE_KEYS` requires this date's string representation to match the `"%Y-%m-%dT%H:%M:%S%z"` format.

Additionally, `replace(microsecond=0)` is required because the default `faker.date_time()` output includes a random microsecond value, but then its string representation includes microseconds which also fails to match the aforementioned pattern.

### DIFF
--- a/quipucords/tests/utils/raw_facts_generator.py
+++ b/quipucords/tests/utils/raw_facts_generator.py
@@ -1,5 +1,7 @@
 """Utility functions to generate raw facts."""
 
+import zoneinfo
+
 from faker import Faker
 
 from constants import DataSources
@@ -218,7 +220,9 @@ def _openshift_raw_facts_generator(number_of_facts, as_native_types=True):
             architecture="amd64",
             kernel_version=f"{fake_semver()}-{fake_semver()}.el{_faker.random_digit()}.x86_64",
             operating_system="linux",
-            creation_timestamp=_faker.date_time(),
+            creation_timestamp=_faker.date_time(
+                tzinfo=zoneinfo.ZoneInfo("UTC")
+            ).replace(microsecond=0),
             cluster_uuid=cluster.uuid,
         )
         yield {"node": node.dict() if as_native_types else node}


quipucords/tests/utils/raw_facts_generator.py
```diff
@@ -218,7 +220,9 @@ def _openshift_raw_facts_generator(number_of_facts, as_native_types=True):
             architecture="amd64",
             kernel_version=f"{fake_semver()}-{fake_semver()}.el{_faker.random_digit()}.x86_64",
             operating_system="linux",
-            creation_timestamp=_faker.date_time(),
+            creation_timestamp=_faker.date_time(
+                tzinfo=zoneinfo.ZoneInfo("UTC")
+            ).strftime("%Y-%m-%dT%H:%M:%S%z"),
```
bruno-fs, line 225: is strftime really required? `RawFactEncoder` should be able to handle datetime objects 🤔

bruno-fs (approved):

quipucords/tests/utils/raw_facts_generator.py
infinitewarp: I believe so, because

> Without a custom strftime call, the default string representation includes microseconds which also fails to match the aforementioned pattern.

I just now checked after your comment, and `RawFactEncoder` naively calls `o.isoformat()`, and that would explain why. I considered expanding the patterns in `fingerprinter.runner.RAW_DATE_KEYS` to match `faker`'s default `datetime` outputs in the fingerprinting phase, but I wanted to limit scope since this is a testing-only change.

I suppose I could try to `.replace(microsecond=0)` the returned object from `date_time(...)` and see if that results in no microseconds in the output without forcing a `strftime` here… I'll give that a try shortly and amend a commit if that succeeds.

PR #2662, opened 2024-05-17
test: improve raw fact generators value randomness

infinitewarp:
These changes make our generated raw facts slightly more realistic or at least more varied so that they will product more interesting and less uniform fingerprints in tests.

### DIFF
--- a/quipucords/tests/utils/raw_facts_generator.py
+++ b/quipucords/tests/utils/raw_facts_generator.py
@@ -11,6 +11,12 @@ _faker = Faker()
 
 DEFAULT_RHEL_OS_NAME = "Red Hat Enterprise Linux"
 
+# Small list of cluster names to be used in all vcenter fact generation so that in a
+# large (>10) list of facts at least some VMs will appear to belong to the same host.
+VCENTER_RANDOM_CLUSTER_NAMES = [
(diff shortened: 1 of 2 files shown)


quipucords/tests/utils/raw_facts_generator.py
```diff
@@ -132,7 +138,9 @@ def _vcenter_raw_facts():
         "vm.mac_addresses": [_faker.mac_address()],
         "vm.memory_size": None,
         "vm.name": _faker.slug(),
-        "vm.os": None,
+        "vm.os": _faker.random_element(["red hat enterprise linux", "rhel"])
+        if _faker.pybool()
+        else None,
```
bruno-fs, line 143: this formatting is... weird 🤔

bruno-fs (approved):

quipucords/tests/utils/raw_facts_generator.py
bruno-fs: I think black would use parenthesis here and add indentation.

infinitewarp: Yeah… you can thank `ruff format` for that one. 😕

PR #2666, opened 2024-05-21
perf: allow bypassing cached_fingerprints build for non-production dev environments

infinitewarp:
🙈 🙈 🙈

🤫

🔥 🐶 ☕ 🔥 

This PR adds support for a new environment variable `QUIPUCORDS_BYPASS_BUILD_CACHED_FINGERPRINTS` that will bypass building and setting `DeploymentsReport.cached_fingerprints` during the fingerprinting process. This will cause problems if actual customers tried to enable that variable because it will result in empty or broken data when they try to download their report tarball. So, I set this only when `PRODUCTION` is not `True`, and that means our containers (both upstream and downstream) _will not_ enable this bypass because they both set `ENV PRODUCTION=True` in their respective `Dockerfile`s.

Then why am I adding this bypass? I'm performance testing quipucords with moderately large sets of simulated raw facts and fingerprints, but due to https://issues.redhat.com/browse/DISCOVERY-691, our service is not capable of handling more than ~60,000 fingerprints. I need far more than that, and my perf tests don't need `cached_fingerprints`. Hopefully this is a short-lived bypass until we better handle DISCOVERY-691, but we may want to keep this around until we further refactor our "cached" data.

### DIFF
--- a/quipucords/fingerprinter/runner.py
+++ b/quipucords/fingerprinter/runner.py
@@ -219,24 +219,25 @@ class FingerprintTaskRunner(ScanTaskRunner):
                     number_invalid += 1
                     self.scan_task.log_message(
                         "The fingerprint could not be saved. "
-                        f"Fingerprint: {str(error).strip()}. Error: {fingerprint_dict}",
+                        f"Fingerprint: {str(error).strip()}.",
                         log_level=logging.ERROR,
                         exception=error,
                     )
+                    logger.error(fingerprint_dict)
             else:
                 number_invalid += 1
                 self.scan_task.log_message(
-                    f"Invalid fingerprint: {fingerprint_dict}",
-                    log_level=logging.ERROR,
+                    "Invalid fingerprint.", log_level=logging.ERROR
                 )
+                logger.error(fingerprint_dict)
                 self.scan_task.log_message(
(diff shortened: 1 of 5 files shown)


quipucords/fingerprinter/runner.py
```diff
@@ -138,7 +139,7 @@ def execute_task(self, manager_interrupt):
             )
             raise error
 
-    def _process_details_report(self, manager_interrupt, report):  # noqa: PLR0915
+    def _process_details_report(self, manager_interrupt, report):  # noqa: PLR0915, C901, PLR0912
```
bruno-fs, line 142: I don't like shushing C901. That's a good check IMO. I only found ok turning it off for legacy functions that were already violating it.

bruno-fs: OTOH, I think this method was already pretty bad before. These new conditionals are not what really makes this a mess...

infinitewarp: I agree. Normally when I see C901 and related complexity alerts, that motivates me to refactor the code. In this specific case, though, I want to touch the code as little as possible. This code was just barely on the edge of hitting the complexity alert before, and my small bypass conditions pushed it over the edge. 😕

bruno-fs (approved):

PR #2672, opened 2024-05-31
refactor: rewrite deploy/setup_user.py as a Django management command…

infinitewarp:
… "create_or_update_user" so we can test and update its behavior

Relates to JIRA: DISCOVERY-683


quipucords/tests/quipucords/test_user.py
```diff
@@ -54,3 +60,82 @@ def test_reject_user_attribute_similarities(self):
...
+        password = None
+        created, updated, generated_password = create_or_update_user(
+            username, email, password
+        )
+        assert created
+        assert not updated
+        assert generated_password
+        assert User.objects.filter(username=username).exists()
+        with override_settings(AXES_ENABLED=False):
+            assert Client().login(username=username, password=generated_password)
```
bruno-fs, line 81: Is there a reason to not use [`client` fixture](https://pytest-django.readthedocs.io/en/latest/helpers.html#client-django-test-client) here?

bruno-fs: Same for other tests using `django.test.Client` directly

infinitewarp: Oh, I might have forgotten about that one! Trying it now…

infinitewarp: Good catch! It looks like that works perfectly with no other changes. I have amended and force pushed the last two commits to use the `client` fixture. 👍

bruno-fs (approved):

PR #2676, opened 2024-06-06
Bump version to 1.8.1

bruno-fs:


infinitewarp (approved):

requirements-build.txt
```diff
@@ -81,10 +80,11 @@ wheel==0.43.0
     #   tzdata
 
 # The following packages are considered to be unsafe in a requirements file:
-setuptools==70.0.0 ; python_version >= "3.11" and python_version < "3.13"
+setuptools==69.0.2 ; python_version >= "3.11" and python_version < "3.13"
```
ruda, line 83: Is this an intentional downgrade? Setup Tools `70.0.0` is the latest stable.

ruda (approved):

poetry.lock
```diff
@@ -3096,19 +3090,18 @@ win32 = ["pywin32"]
 
 [[package]]
 name = "setuptools"
-version = "69.5.1"
+version = "70.0.0"
```
bruno-fs, line 3093: downgrade?

requirements.txt
```diff
@@ -52,15 +52,15 @@ python-dateutil==2.9.0.post0 ; python_version >= "3.11" and python_version < "3.
...
 pyyaml==6.0.1 ; python_version >= "3.11" and python_version < "3.13"
-redis==5.0.4 ; python_version >= "3.11" and python_version < "3.13"
+redis==5.0.5 ; python_version >= "3.11" and python_version < "3.13"
 requests-oauthlib==2.0.0 ; python_version >= "3.11" and python_version < "3.13"
-requests==2.32.2 ; python_version >= "3.11" and python_version < "3.13"
+requests==2.32.3 ; python_version >= "3.11" and python_version < "3.13"
 resolvelib==1.0.1 ; python_version >= "3.11" and python_version < "3.13"
 rsa==4.9 ; python_version >= "3.11" and python_version < "3.13"
-setuptools==69.5.1 ; python_version >= "3.11" and python_version < "3.13"
+setuptools==70.0.0 ; python_version >= "3.11" and python_version < "3.13"
```
bruno-fs, line 60: downgrade[2]?

requirements-build.txt
bruno-fs: requirements-build "downgrades" are not proper downgrades. these are all transitive dependencies, and are the best we could solve for our current deps.

This one is the result of upgrading ansible-runner, which has a specific setuptools requirement: https://github.com/ansible/ansible-runner/blob/2.4.0/pyproject.toml#L2

bruno-fs: requirements.txt (and poetry.lock) are the true source of our dependencies, which don't have a downgrade but rather an upgrade to the latest:
- https://github.com/quipucords/quipucords/pull/2676#discussion_r1629878846
- https://github.com/quipucords/quipucords/pull/2676#discussion_r1629880191

requirements.txt
ruda: IMHO I prefer to downgrade and keep setuptools in both files to the same version.

ruda (approved):

poetry.lock
ruda: I would downgrade to keep everything in the same version, reason: to avoid surprises, but not a big thing.

requirements.txt
bruno-fs: This would mean converting setuptools from a indirect to direct dependency. Why? Just because to keep it in sync with transitive dependencies (which have no obligation to be in sync with actual runtime dependencies - they could all be in conflict and it would be fine).

If you are installing `requirements-build.txt`, you are doing something wrong. This file only exists to do dependency resolution of transitive dependencies that will be used downstream. It could have multiple versions of setuptools if required and it would be fine.

ruda (approved):
LGTM!

PR #2685, opened 2024-06-12
feat: add hostnamectl status to network fact collection

bruno-fs:
Example of a successful hostnamectl fact in details report

```sh
jq '.sources[].facts[].hostnamectl' ~/Downloads/report_id_9/details.json
```
```JSON
{
  "value": {
    "kernel": "Linux 6.9.4-200.fc40.x86_64",
    "boot_id": "12fbdbc9c4da4a44a85e529c72f7dec1",
    "chassis": "vm",
    "icon_name": "computer-vm",
    "machine_id": "73c65da657ee4ad1be417ff471f58c2f",
    "cpe_os_name": "cpe:/o:fedoraproject:fedora:40",
    "architecture": "x86-64",
    "firmware_age": "9y 4month 2w 5d",
    "firmware_date": "Fri 2015-02-06",
    "hardware_model": "Standard PC _Q35 + ICH9, 2009_",
    "os_support_end": "Tue 2025-05-13",
    "virtualization": "qemu",
    "hardware_vendor": "QEMU",
    "static_hostname": "fedora40",
    "firmware_version": "0.0.0",
    "operating_system": "Fedora Linux 40 (Server Edition)",
    "os_support_remaining": "10month 2w 1d"
  },
  "error_msg": null,
  "return_code": 0
}
```

Relates to JIRA: DISCOVERY-428
https://issues.redhat.com/browse/DISCOVERY-428


quipucords/scanner/network/processing/process.py
```diff
@@ -29,6 +31,17 @@
...
 
+@dataclass
+class ProcessedResult:
+    """Processed result including relevant metadata."""
+
+    # return_code == 0 = success
+    return_code: int
+    error_msg: str = None
+    # the actual fact value
+    value: Any = None
```
bruno-fs, line 42: This piece is more important than the feature itself, IMO.

I've been thinking how we could make the raw facts report more transparent about what happened - and help us make decisions later on how to proper use it during fingerprint/normalization and came up with this.
return code and error message are the only things that came to mind (trying to follow the same pattern that we did on openshift implementation). With this we gain granularity of what exactly happened and maybe what caused a fact to not be collected / processed.

quipucords/quipucords/settings.py
```diff
@@ -321,6 +321,9 @@ def app_secret_key_and_path():
 
 STATICFILES_DIRS = [BASE_DIR / "client"]
 
+if not PRODUCTION:
+    TESTDATA_DIR = BASE_DIR / "testdata"
```
infinitewarp, line 325: Surprise! 🎉 What do you think about this addition? If this is okay, then I hope we could expand use of this new `TESTDATA_DIR` for future unit tests that might benefit from large sets of static data, possibly captured from real-word commands like I did here for hostnamectl.

infinitewarp:
Link to (successful! ✅ ) standalone CI test run after updating camayoc (https://github.com/quipucords/camayoc/pull/522) to expect the new `hostnamectl` fact: https://stage-jenkins-csb-smqe.apps.ocp-c1.prod.psi.redhat.com/view/Discovery/job/discovery-standalone/150/

quipucords/quipucords/settings.py
bruno-fs: my first thought is this seems to be reinventing the wheel (see [tmp_path fixture](https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmp-path-fixture)), but I just started looking at the PR.

bruno-fs: ah, I see now what its purpose is. seems good to me =D

bruno-fs:
LGTM, but github won't let me approve it 😛

infinitewarp (approved):
I approve my changes. 😎

PR #2690, opened 2024-06-27
feature: allow customization of default data path

bruno-fs:
For local development, keep its default as is. But for production workloads point it the volume already documented (/var/data).

EDIT: integrations tests passed for the PR with same commit as this one: https://url.corp.redhat.com/3aec94e
closes #2691


infinitewarp (approved):

quipucords/quipucords/settings.py
```diff
@@ -28,7 +28,7 @@
 # BASE_DIR is ./quipucords/quipucords
 BASE_DIR = Path(__file__).absolute().parent.parent
 # DEFAULT_DATA_DIR is ./var, which is on .gitignore
-DEFAULT_DATA_DIR = BASE_DIR.parent / "var"
+DEFAULT_DATA_DIR = Path(env.str("QUIPUCORDS_DATA_DIR", BASE_DIR.parent / "var"))
```
mirekdlugosz, line 31: That will make https://github.com/quipucords/discovery-quadlets/pull/12 obsolete and we should revert it after merging this, right?

Or will this re-introduce problem fixed by https://github.com/quipucords/discovery-quadlets/pull/12 and we should have another PR in quadlets that will re-fix it?

bruno-fs: Thanks for pointing this out. I wasn't around when this issue was detected/fixed.

`/var/app` is the path used with current discovery deployment,  and is mapped to `%h/.local/share/discovery/data:`. I believe setting `QUIPUCORDS_DATA_DIR` to `/var/app` makes said fix obsolete. I don't think it will cause a regression, but it won't be necessary anymore.

mirekdlugosz: I _think_ reverting https://github.com/quipucords/discovery-quadlets/pull/12 will be the right choice, but let's merge this and wait for quadlet pipeline results.

mirekdlugosz (approved):

PR #2692, opened 2024-06-27
fix: fail bigger when failing to access DeploymentsReport cached_fingerprints or cached_csv

infinitewarp:
After reflecting on the initial change introduced by 8729eb4d, we decided that we would prefer to log errors and fail bigger if we fail to read data from DeploymentsReport cached_fingerprints or cached_csv. This can happen if the cached files were moved after they were written or if the process trying to access the data does not have access to the files. The latter could easily happen in a misconfigured container orchestration.

Following this change, if you perform a successful scan but then manually destroy, move, or make unreadable its cached files, the next time you try to download the report, you should encounter a 500 Internal Server Error. Returning a 500 error is appropriate here because there is nothing the client can do to fix the issue, and it requires some investigation and unspecified fixes on the server side.

In the UI, this may error manifest like this:
Screenshot 2024-06-27 at 3 18 22 PM

Unfortunately, the current UI is not very clever in its handling of 500 errors.

In the server logs, you may see errors logged (plus stack traces) like this:
```
[ERROR 2024-06-27T19:21:46 pid=84056 tid=123145384759296 api/deployments_report/model.py:cached_fingerprints:79] [Errno 2] No such file or directory: '/Users/brasmith/projects/quipucords/var/cached_reports/deployments-report-11-1719432083.0758321.json'
[ERROR 2024-06-27T19:21:46 pid=84056 tid=123145384759296 api/deployments_report/model.py:cached_fingerprints:80] Cached fingerprints file for DeploymentsReport 11 not found at '/Users/brasmith/projects/quipucords/var/cached_reports/deployments-report-11-1719432083.0758321.json'
[ERROR 2024-06-27T19:21:46 pid=84056 tid=123145384759296 /Users/brasmith/Library/Caches/pypoetry/virtualenvs/quipucords-A54Z8Zh8-py3.12/lib/python3.12/site-packages/django/utils/log.py:log_response:241] Internal Server Error: /api/v1/reports/11/
```

Relates to JIRA: DISCOVERY-703

### DIFF
--- a/quipucords/api/deployments_report/model.py
+++ b/quipucords/api/deployments_report/model.py
@@ -66,22 +66,23 @@ class DeploymentsReport(BaseModel):
         file_path = Path(self.cached_fingerprints_file_path).absolute()
         if file_path.parent != cached_files_path():
             # Check to protect against potentially malicious filesystem access.
-            logger.error(
-                "Unsupported parent path for DeploymentsReport %s file: %s",
-                self.id,
-                file_path,
+            message = (
+                "Unsupported parent path for DeploymentsReport "
+                f"{self.id} cached_fingerprints file: {file_path}"
             )
-            return None
+            logger.error(message)
+            raise PermissionError(message)
         try:
             with file_path.open("r") as f:
                 return json.load(f)
-        except FileNotFoundError:
-            logger.warning(
-                "Cached fingerprints file for DeploymentsReport %s not found at %s",
+        except FileNotFoundError as e:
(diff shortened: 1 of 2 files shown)


ruda (approved):

mirekdlugosz (approved):
Tested in environment where issue was initially observed - `/reports/` endpoint returns 500, causing test to fail with clearer hint what the problem is. It's also easier to find errors logged in container log.

PR #2693, opened 2024-06-28
feat: use hostnamectl to fingerprint infrastructure_type

infinitewarp:
This is a followup to https://github.com/quipucords/quipucords/pull/2685 which only added `hostnamectl` to fact collection but did not yet use its values in fingerprinting.

### DIFF
--- a/quipucords/fingerprinter/runner.py
+++ b/quipucords/fingerprinter/runner.py
@@ -82,6 +82,30 @@ NAME_RELATED_FACTS = ["name", "vm_dns_name", "virtual_host_name"]
 COMBINED_KEY = "combined_fingerprints"
 
 
+def fingerprint_network_infrastructure_type(fact):
+    """Determine if running on VM or bare metal."""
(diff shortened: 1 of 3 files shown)


bruno-fs (approved):

quipucords/fingerprinter/runner.py
```diff
@@ -82,6 +82,42 @@
 COMBINED_KEY = "combined_fingerprints"
 
 
+def fingerprint_network_infrastructure_type(fact: dict) -> tuple[str, str]:
+    """Determine if running on VM or bare metal."""
+    virt_what_type = fact.get("virt_what_type")
+    hostnamectl_chassis = (
+        (fact.get("hostnamectl", {}) or {}).get("value", {}).get("chassis")
+    )
```
bruno-fs, line 90: Ah, I just remembered now we have a utility for this

```suggestion
    hostnamectl_chassis = deepget(fact, "hostnamectl__value__chassis")
```

PR #2696, opened 2024-07-02
fix: improve handling of virt-what output for network scan fingerprinting

infinitewarp:
We now ignore output from virt-what if the exit code is non-zero. This also moves virt-what output processing logic from remote Ansible commands to code that we can better control, test, and debug in Python.

Of course, expect discovery-ci/camayoc tests to fail until we update camayoc to expect the updated raw fact names.

Relates to JIRA: DISCOVERY-427


quipucords/scanner/network/runner/roles/virt_what/tasks/main.yml
```diff
@@ -4,33 +4,13 @@
...
-    virt_what_type: "bare metal"
-  ignore_errors: yes
-  when: 'user_has_sudo and internal_have_virt_what and (not internal_virt_what_error|int  == 0)'
-
-- name: set virt-what.type fact if virt-what ran successfully
-  set_fact:
-    virt_what_type: '{{ ";".join(internal_virt_what_output["stdout_lines"][:-1]) }}'
-  ignore_errors: yes
-  when: 'user_has_sudo and internal_have_virt_what and (internal_virt_what_error|int == 0) and "stdout_lines" in internal_virt_what_output'
+    virt_what: "{{ internal_virt_what_output }}"
```
bruno-fs, line 16: my eyes are so happy to see this

bruno-fs (approved):
love it. :shipit:

PR #2709, opened 2024-07-22
DISCOVERY-419: use bare executable names in roles

mirekdlugosz:
The main point of the PR is:

* use `command -v` consistently inside `check_dependencies` role
* use bare executable names in all roles

A bit historical context: this fundamentally follows up to #1298 and #1309 . The main problem is that we want to run some roles conditionally (only if required tools are found), but sometimes we run roles as root and sometimes we do not. The attempt to solve it was to use hardcoded absolute paths, but this created the new issue that host might have the tool under different path. The core insight in this PR is that we should split dependencies check into root and non-root check, and then trust the path to find the right executable.

Each ansible variable is changed in separate commit. I will squash them before merging, I just thought it might be easier to review this way.

I decided to keep `internal_have_*` names for variables that require root, and use `internal_have_*_user` for variables that are run as user. The reason for that is that most of roles require root, so this way things are less noisy.

There were few variables that were only used as non-root. To keep things consistent, I have renamed them to `*_user`. Technically this is not required and I can drop these commits if people feel strongly about this.

Camayoc requires changes, introduced in https://github.com/quipucords/camayoc/pull/537 . I have run container from this PR against Camayoc branch from 537, results are in [standalone job 166](https://stage-jenkins-csb-smqe.apps.ocp-c1.prod.psi.redhat.com/view/Discovery/job/discovery-standalone/166/).

Relates to JIRA: DISCOVERY-419

### DIFF
--- a/quipucords/scanner/network/runner/roles/check_dependencies/tasks/main.yml
+++ b/quipucords/scanner/network/runner/roles/check_dependencies/tasks/main.yml
@@ -120,20 +120,16 @@
   ignore_errors: yes
 
 - name: gather internal_have_systemctl_cmd
-  raw: export LANG=C LC_ALL=C; whereis systemctl
-  register: internal_whereis_systemctl_cmd
+  raw: export LANG=C LC_ALL=C; command -v systemctl
+  register: internal_have_systemctl_cmd
+  become: yes
   ignore_errors: yes
+  when: 'user_has_sudo'
 
 - name: set internal_have_systemctl
   set_fact:
-    internal_have_systemctl: false
+    internal_have_systemctl: "{{ internal_have_systemctl_cmd.get('rc') == 0 }}"
   ignore_errors: yes
-
-- name: set internal_have_systemctl
-  set_fact:
-    internal_have_systemctl: true
-  ignore_errors: yes
-  when: (internal_whereis_systemctl_cmd.get('rc') == 0 and internal_whereis_systemctl_cmd.get('stdout_lines')[-1] != 'systemctl:')
 
 - name: gather internal_have_chkconfig_cmd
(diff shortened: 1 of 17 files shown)


bruno-fs (approved):

quipucords/scanner/network/runner/roles/check_dependencies/tasks/main.yml
```diff
@@ -149,21 +149,27 @@
     internal_have_chkconfig: "{{ internal_have_chkconfig_cmd.get('rc') == 0 }}"
   ignore_errors: yes
 
-- name: gather internal_whereis_ifconfig_cmd
-  raw: export LANG=C LC_ALL=C; whereis ifconfig
-  register:  internal_whereis_ifconfig_cmd
+- name: gather internal_have_ifconfig_cmd
+  raw: export LANG=C LC_ALL=C; command -v ifconfig
+  register:  internal_have_ifconfig_cmd
+  become: yes
```
infinitewarp, line 153: Why does this add `become: yes` and `when: 'user_has_sudo'`? Unless I misunderstand something, the way we use `ifconfig` does not actually require `sudo`, and and normal user can probably run `ifconfig -a`.

I noticed you added this sudo requirement in several other places that I don't think require them. Did I overlook an explanation for this?

infinitewarp (approved):

quipucords/scanner/network/runner/roles/check_dependencies/tasks/main.yml
mirekdlugosz: Yes, the main idea is to maintain a consistent environment between place where we call `command -v` and place that is executed conditionally on `internal_have_*` variable. There are few executables that we call as root and as normal user, so they are duplicated.

PR #2731, opened 2024-09-04
build: upgrade Django from 4.2 to 5.1, upgrade related requirements

infinitewarp:
Also, to support this upgrade, drop deprecated `User.objects.make_random_password`. `User.objects.make_random_password` was deprecated in 4.2 and removed in Django 5.1. See also: https://docs.djangoproject.com/en/5.1/releases/4.2/#id1


infinitewarp:
Or not! I forgot that we decided to wait for the next LTS release (5.2).

PR #2734, opened 2024-09-11
chore(ScanManager): remove fluff celery don't need

bruno-fs:
In another step towards getting rid of the ScanManager, remove the global manager, and the need to create a single class instance. Celery doesn't need these things.

Also, remove scan_manager fixture. From now on, tests relying on celery shall either use celery_app/celery_worker fixtures or enable CELERY_TASK_ALWAYS_EAGER.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -12,7 +12,6 @@ import pytest
 from django.contrib.auth import get_user_model
 from django.core import management
 from django.test import Client as DjangoClient
-from django.test import override_settings
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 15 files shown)


quipucords/tests/api/management/commands/test_create_or_update_user.py
```diff
@@ -45,7 +45,12 @@ def test_create_or_update_user_create_with_valid_password(faker):
     """Test creation of a user with a valid password."""
     out = StringIO()
     username = faker.user_name()
-    password = faker.password(length=settings.QUIPUCORDS_MINIMUM_PASSWORD_LENGTH)
+    password = faker.password(
+        length=settings.QUIPUCORDS_MINIMUM_PASSWORD_LENGTH,
+        digits=True,
+        upper_case=True,
+        lower_case=True,
+    )
```
infinitewarp, line 53: Why change this? It looks like the defaults are already `True`.

https://faker.readthedocs.io/en/master/providers/faker.providers.misc.html#faker.providers.misc.Provider.password
```
password(length: int = 10, special_chars: bool = True, digits: bool = True, upper_case: bool = True, lower_case: bool = True) → str
```

bruno-fs: funny. I got an local error I assumed was due to these values being false.

bruno-fs: It's a shame I didn't record the error. It was something along the lines of "password has no numeric characters". 

I'm gonna remove this commit and, shall this error happen again, I'll record it.

PR #2735, opened 2024-09-16
feat: add "auth_type" to credentials api

infinitewarp:
Example API responses showing the new `auth_type` field:

```sh
http 127.0.0.1:8000/api/v1/credentials/ Authorization:"Token $API_TOKEN" page_size==2 page==8
```

```json
{
    "count": 23,
    "next": "http://127.0.0.1:8000/api/v1/credentials/?page=9&page_size=2",
    "previous": "http://127.0.0.1:8000/api/v1/credentials/?page=7&page_size=2",
    "results": [
        {
            "auth_type": "password",
            "created_at": "2024-09-16T18:39:23.329554Z",
            "cred_type": "vcenter",
            "id": 27823,
            "name": "b2ebca84-f592-43ef-8f90-98c3e2910046",
            "password": "********",
            "updated_at": "2024-09-16T18:39:24.941053Z",
            "username": "87a46b00-da1e-43bc-b8fc-87371861c611"
        },
        {
            "auth_type": "ssh_keyfile",
            "become_method": "sudo",
            "become_user": "root",
            "created_at": "2024-09-16T18:39:53.259351Z",
            "cred_type": "network",
            "id": 27829,
            "name": "b3624fb3-8a89-49fd-82c2-06e8e1912069",
            "ssh_keyfile": "/Users/brasmith/.ssh/discovery-keys/brad-dev01-20230221.pem",
            "updated_at": "2024-09-16T18:39:53.259368Z",
            "username": "ba82489f-d722-4d70-8676-433a8e35b2aa"
        }
    ]
}
```

```sh
http 127.0.0.1:8000/api/v1/credentials/27791/ Authorization:"Token $API_TOKEN"
```

```json
{
    "auth_token": "********",
    "auth_type": "auth_token",
    "cred_type": "openshift",
    "id": 27791,
    "name": "openshift-20240708"
}
```

### DIFF
--- a/quipucords/api/credential/serializer.py
+++ b/quipucords/api/credential/serializer.py
@@ -5,6 +5,7 @@ from collections import defaultdict
 from pathlib import Path
 
 from django.utils.translation import gettext as _
+from rest_framework.fields import SerializerMethodField
 from rest_framework.serializers import ValidationError, empty
 
 from api import messages
@@ -42,6 +43,19 @@ class CredentialSerializer(NotEmptySerializer):
     """Base Serializer for the Credential model."""
 
     sources = RelatedSourceSerializer(many=True, read_only=True)
+    auth_type = SerializerMethodField()
+
+    def get_auth_type(self, credential: Credential) -> str:
+        """Determine a credential's authentication type."""
+        if credential.auth_token:
+            return "auth_token"
+        elif credential.password:
+            return "password"
+        elif credential.ssh_key:
+            return "ssh_key"
+        elif credential.ssh_keyfile:
+            return "ssh_keyfile"
+        return "unknown"
 
     class Meta:
         """Metadata for the serializer."""
@@ -132,7 +146,7 @@ class AuthTokenSerializer(CredentialSerializer):
         """Serializer configuration."""
 
         model = Credential
-        fields = ["id", "name", "cred_type", "auth_token", "sources"]
+        fields = ["auth_type", "id", "name", "cred_type", "auth_token", "sources"]
         extra_kwargs = {
             "auth_token": {"required": True, **ENCRYPTED_FIELD_KWARGS},
         }
@@ -146,6 +160,7 @@ class UsernamePasswordSerializer(CredentialSerializer):
 
         model = Credential
         fields = [
+            "auth_type",
             "cred_type",
             "id",
             "name",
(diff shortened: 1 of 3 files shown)


quipucords/api/credential/serializer.py
```diff
@@ -42,6 +43,19 @@ class CredentialSerializer(NotEmptySerializer):
...
+        """Determine a credential's authentication type."""
+        if credential.auth_token:
+            return "auth_token"
+        elif credential.password:
+            return "password"
+        elif credential.ssh_key:
+            return "ssh_key"
+        elif credential.ssh_keyfile:
+            return "ssh_keyfile"
+        return "unknown"
```
infinitewarp, line 58: I support these could reference an enum or constants or something, but… effort? 🤷  This may be a short-lived piece of code anyway since we've started working on the v2 credentials api. This just needs to help cover the gap since the new UI code is suffering without it.

nicolearagao: I am ok with this.

nicolearagao (approved):

ruda:
Needs to update the REST API documentation. https://github.com/quipucords/quipucords/blob/main/docs/swagger.yml

bruno-fs:
@ruda, Instead of updating the swagger doc can't we just 🪓 it instead?

IMO We should put our effort for documenting soon-to-be deprecated routes, we should instead document the new ones (v2).

DRF has its built in way of doing it, however their [documentation suggests using a third party lib instead](https://www.django-rest-framework.org/topics/documenting-your-api/#built-in-openapi-schema-generation-deprecated).
For now I'd follow their advice and use [drf-spetacular](https://drf-spectacular.readthedocs.io/en/latest/).

bruno-fs:
> It looks like "ssh_keyfile" value is not present in credential unit test. Could / should we add a test with that value?

I noticed that too. However, I expect `ssh_keyfile` to be completely wiped away from this codebase VERY SOON (1.11?). With that in mind, I particularly would look the other way if we are missing something for it.

Unless you strongly believe we need it, in that case I wouldn't argue, @mirekdlugosz.

mirekdlugosz:
No, I don't consider that a blocker. If adding a test is easy enough, then better to add it, but I'm generally ok with that merged with a bit of missing coverage.

PR #2737, opened 2024-09-16
chore: get rid of unsupported installation methods

bruno-fs:
Yet another push in the movement of "THE ONLY SUPPORTED INSTALLATION METHOD IS QUIPUCORDS-INSTALLER"


infinitewarp (approved):

mirekdlugosz (commented):
I think you can also remove the word "docker" from `clean-db` description in help (line 40).

And is this time to remove the monstrosity in lines 24-32 (`ifndef DOCKER_HOST`)? That was added to help people on macOS, who still used Docker at the time, and it _might_ have been useful for tests that started container. We have since removed these tests, and from what I can see Makefile now calls `podman` explicitly and never `docker`.

bruno-fs:
> And is this time to remove the monstrosity in lines 24-32 (ifndef DOCKER_HOST)? 

Good point. Axing it as well.

mirekdlugosz (approved):

PR #2722, opened 2024-08-14
feat: Optionally save a list of skipped Ansible tasks

mirekdlugosz:
A lot of Ansible tasks used during network scans are conditional. Some of them might never get executed - these are tasks that might require additional testing or extra attention during changes. If things are not executed during standard PR / nightly run, then they might be broken without us knowing.

This commit adds a new feature flag, enabled through `QPC_FEATURE_REPORT_SKIPPED_TASKS` environment variable. When active, it will track skipped tasks during scan run and create a list of all tasks that were skipped on all hosts. This effectively creates a list of tasks that are not executed.

Following command may be used to create a list of tasks that are skipped on every host in every scan from last Discovery session:

    cd ~/.local/share/discovery/data  # or other QUIPUCORDS_DATA_DIR value
    sort skipped-tasks-* |uniq -d

When running against our standard testlab environment, this provided following list:

```
/app/quipucords/scanner/network/runner/roles/check_dependencies/tasks/main.yml:115
/app/quipucords/scanner/network/runner/roles/date/tasks/main.yml:19
/app/quipucords/scanner/network/runner/roles/dmi/tasks/main.yml:22
/app/quipucords/scanner/network/runner/roles/dmi/tasks/main.yml:43
/app/quipucords/scanner/network/runner/roles/dmi/tasks/main.yml:64
/app/quipucords/scanner/network/runner/roles/etc_release/tasks/main.yml:105
/app/quipucords/scanner/network/runner/roles/etc_release/tasks/main.yml:118
/app/quipucords/scanner/network/runner/roles/etc_release/tasks/main.yml:124
/app/quipucords/scanner/network/runner/roles/etc_release/tasks/main.yml:60
/app/quipucords/scanner/network/runner/roles/etc_release/tasks/main.yml:89
/app/quipucords/scanner/network/runner/roles/etc_release/tasks/main.yml:97
/app/quipucords/scanner/network/runner/roles/ifconfig/tasks/main.yml:20
/app/quipucords/scanner/network/runner/roles/ifconfig/tasks/main.yml:32
/app/quipucords/scanner/network/runner/roles/ifconfig/tasks/main.yml:45
/app/quipucords/scanner/network/runner/roles/ifconfig/tasks/main.yml:7
/app/quipucords/scanner/network/runner/roles/ip/tasks/main.yml:20
/app/quipucords/scanner/network/runner/roles/ip/tasks/main.yml:7
[...]

PR #2740, opened 2024-09-19
refactor(fingerprinter): make all type coercions EXPLICIT

bruno-fs:
Explicit is better than implicit. This should fix a RIDICULOUS bug that caused os_version "8.10" to become "8.1".

For the record, type coercions were all based on SystemFingerprint fields as follows:

Integer fields:
- cpu_count
- cpu_socket_count
- cpu_core_per_socket
- system_user_count
- vm_host_socket_count
- vm_host_core_count
- redhat_package_count
- system_memory_bytes

Float fields:
- cpu_core_count

Boolean fields:
- cpu_hyperthreading
- is_redhat

Relates to JIRA: DISCOVERY-783

### DIFF
--- a/quipucords/api/common/util.py
+++ b/quipucords/api/common/util.py
@@ -48,37 +48,6 @@ def convert_to_int(value):
     return int(value)
 
 
-def is_float(value):
-    """Check if a value is convertable to float.
-
-    :param value: The value to convert
-    :returns: bool indicating if it can be converted
-    """
-    if isinstance(value, float):
-        return True
-    if isinstance(value, int):
-        return False
-    if not isinstance(value, str):
-        return False
-
-    try:
-        float(value)
-        return True
-    except ValueError:
-        return False
-
-
-def convert_to_float(value):
-    """Convert value to float if possible.
-
-    :param value: The value to convert
-    :returns: The int or None if not convertable
-    """
-    if not is_float(value):
-        return None
-    return float(value)
-
-
 def is_boolean(value):
     """Check if a value is a bool cast as string.
 
@@ -103,6 +72,19 @@ def convert_to_boolean(value):
     if is_boolean(value):
         return value.lower() == "true"
     return False
+
+
+def convert_to_bool_or_none(value):
+    """Convert a string "True" or "False" to boolean.
+
+    If value already is a boolean, returns it untouched.
+    Otherwise, return None.
+    """
+    if isinstance(value, bool):
+        return value
+    if is_boolean(value):
+        return value.lower() == "true"
+    return None
 
 
 def check_for_existing_name(queryset, name, error_message, search_id=None):
--- a/quipucords/fingerprinter/formatters.py
+++ b/quipucords/fingerprinter/formatters.py
@@ -31,3 +31,12 @@ def convert_architecture(architecture):
     if not architecture_map.get(architecture):
         return architecture
     return architecture_map.get(architecture)
+
+
+def str_or_none(value):
+    """Normalize value to be a string (or None when "empty")."""
+    if value is None:
+        return None
+    if isinstance(value, str) and value.strip() == "":
+        return None
+    return str(value).strip()
--- a/quipucords/fingerprinter/formatters.py
+++ b/quipucords/fingerprinter/formatters.py
@@ -1,8 +1,15 @@
 """Fingerprint formatters."""
 
+import datetime
+from contextlib import suppress
+from logging import getLogger
+from typing import Any
+
+logger = getLogger(__name__)
+
 
 def format_mac_addresses(mac_addresses):
-    """Format mac addresess."""
+    """Format mac addresses."""
     if isinstance(mac_addresses, list):
         mac_addresses = list(map(lambda x: x.lower(), mac_addresses))
     return mac_addresses
@@ -33,10 +40,62 @@ def convert_architecture(architecture):
     return architecture_map.get(architecture)
 
 
-def str_or_none(value):
+def str_or_none(value) -> str | None:
     """Normalize value to be a string (or None when "empty")."""
     if value is None:
         return None
     if isinstance(value, str) and value.strip() == "":
         return None
     return str(value).strip()
+
+
+def float_or_none(value) -> float | None:
+    """Convert value to float."""
+    with suppress(ValueError, TypeError):
+        return float(value)
+    return None
+
+
+def int_or_none(value) -> int | None:
+    """Convert value to int."""
+    try:
+        return int(value)
+    except TypeError:
+        return None
+    except ValueError:
+        ...
+
+    with suppress(ValueError, TypeError):
+        return int(float_or_none(value))
+
+    return None
+
+
+def list_or_none(value: Any) -> list | None:
+    """Normalize value as list or return None."""
+    if isinstance(value, list):
+        return value
+
+
+def dict_or_none(value: Any) -> dict | None:
+    """Normalize value as dict or return None."""
+    if isinstance(value, dict):
+        return value
+
+
+def list_of_dicts(value: Any) -> list[dict]:
+    """Normalize value as a list of dicts."""
+    if isinstance(value, list) and all(
+        isinstance(inner_val, dict) for inner_val in value
+    ):
+        return value
+    return []
+
+
+def date_or_none(value: Any) -> datetime.date | None:
+    """Normalize value as date or return None."""
+    if isinstance(value, datetime.datetime):
+        return value.date()
+    if isinstance(value, datetime.date):
+        return value
+    return None
--- a/quipucords/fingerprinter/runner.py
+++ b/quipucords/fingerprinter/runner.py
@@ -10,14 +10,7 @@ from django.db import DataError
 from rest_framework.serializers import DateField
 
 from api.common.common_report import create_report_version
-from api.common.util import (
-    convert_to_boolean,
-    convert_to_float,
-    convert_to_int,
-    is_boolean,
-    is_float,
-    is_int,
-)
+from api.common.util import convert_to_bool_or_none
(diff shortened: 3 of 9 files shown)


quipucords/fingerprinter/runner.py
```diff
@@ -938,12 +931,6 @@ def _add_fact_to_fingerprint(  # noqa: PLR0913
         # Remove empty string values
         if isinstance(actual_fact_value, str) and not actual_fact_value:
             actual_fact_value = None
-        if is_boolean(actual_fact_value):
-            actual_fact_value = convert_to_boolean(actual_fact_value)
```
bruno-fs, line 942: this particular logic was ported to a function outside of "formatters" module

quipucords/api/common/util.py
```diff
@@ -105,6 +74,19 @@ def convert_to_boolean(value):
...
+    """Convert a string "True" or "False" to boolean.
+
+    If value already is a boolean, returns it untouched.
+    Otherwise, return None.
+    """
+    if isinstance(value, bool):
+        return value
+    if is_boolean(value):
+        return value.lower() == "true"
+    return None
```
bruno-fs, line 87: This reproduces the original logic that was implicit [here](https://github.com/quipucords/quipucords/pull/2740/files#r1767553142)

quipucords/fingerprinter/runner.py
bruno-fs: https://github.com/quipucords/quipucords/pull/2740/files#r1767554875

mirekdlugosz (commented):
Why do tests in CI all pass, when tests like `test_process_satellite_source` [expect Satellite details report to store `os_version` as a float](https://github.com/quipucords/quipucords/blob/main/quipucords/tests/fingerprinter/test_fingerprint_task.py#L341)? Shouldn't that test fail here? And be changed in the same PR?

-------------------

I've ran standard Camayoc scans against testlab from current `main` and again using container from this PR. Then I compared `details.json` and `deployments.json` from all reports. I saw that in Satellite scans `os_version` is a string. One machine that used to report `8.1` now report `"8.10"`; another used to report `10` but now reports `"10.0"`.

Other than that, I have not noticed any suspicious changes in these reports.

I also compared a list of facts from this PR with a list of roles skipped during standard testlab run (https://github.com/quipucords/quipucords/pull/2722), to have a basic sanity check that our testlab environment could find issues with affected facts. Looks like all facts called out explicitly are covered by testlab.

bruno-fs:
>Why do tests in CI all pass, when tests like test_process_satellite_source [expect Satellite details report to store os_version as a float](https://github.com/quipucords/quipucords/blob/main/quipucords/tests/fingerprinter/test_fingerprint_task.py#L341)? Shouldn't that test fail here?

It definitely should have failed 🤔 
What's going on here? 🤔 🤔 🤔 🤔

infinitewarp (changes requested):
In isolation, I think these changes look good, but I believe we need unit test coverage at least for the use case that revealed the problematic code that we do not want to cast (e.g. assert version value `"8.10"` in raw fact is preserved as `"8.10"`), but also actively check at least one that _should continue_ to be type cast is still cast correctly.

I'm not sure if or how much of that type-checking logic exists in old unit tests, but something probably needs to be updated if no tests failed when changing this logic.

quipucords/tests/fingerprinter/test_fingerprint_task.py
```diff
@@ -338,7 +339,7 @@ def _create_satellite_details_report_json(  # noqa: PLR0913, PLR0912, C901
     hostname="9.8.7.6",
     os_name="RHEL",
     os_release="RHEL 7.3",
-    os_version=7.3,
+    os_version="7.3",
```
bruno-fs, line 342: without making this change, tests relying on this helper fail. Would you consider this enough, coverage-wise @infinitewarp ?

quipucords/fingerprinter/formatters.py
```diff
@@ -1,5 +1,9 @@
 """Fingerprint formatters."""
 
+import datetime
+from contextlib import suppress
+from venv import logger
```
infinitewarp, line 5: `from venv`? 🤔

quipucords/fingerprinter/formatters.py
```diff
@@ -31,3 +35,57 @@ def convert_architecture(architecture):
...
+    except ValueError:
+        ...
+
+    with suppress(ValueError):
+        return int(float_or_none(value))
+
+    return None
+
+
+def list_or_none(value):
```
infinitewarp, line 71: Maybe we should add some type hints so our editors might yell at us if we are using these wrongly.

```suggestion
def list_or_none(value: list) -> list | None:
```

quipucords/fingerprinter/formatters.py
```diff
@@ -31,3 +35,57 @@ def convert_architecture(architecture):
...
+    """Normalize value as list or return None."""
+    if isinstance(value, list):
+        return value
+    logger.error("value '%s' is NOT a list")
+
+
+def dict_or_none(value):
```
infinitewarp, line 78:
```suggestion
def dict_or_none(value: dict) -> dict | None:
```

quipucords/fingerprinter/formatters.py
```diff
@@ -31,3 +35,57 @@ def convert_architecture(architecture):
...
+    """Normalize value as dict or return None."""
+    if isinstance(value, dict):
+        return value
+    logger.error("value '%s' is NOT a dict")
+
+
+def date_or_none(value):
```
infinitewarp, line 85: This one gets a little long, but if we add any, we probably should add them all. Could define custom types, I guess, but _effort_.

```suggestion
def date_or_none(value: datetime.date | datetime.datetime) -> datetime.date | datetime.datetime | None:
```

quipucords/fingerprinter/formatters.py
```diff
@@ -31,3 +35,57 @@ def convert_architecture(architecture):
...
+
+
+def float_or_none(value):
+    """Convert value to float."""
+    with suppress(ValueError, TypeError):
+        return float(value)
+    return None
+
+
+def int_or_none(value):
```
infinitewarp, line 56:
```suggestion
def int_or_none(value) -> int | None:
```

quipucords/fingerprinter/formatters.py
```diff
@@ -31,3 +35,57 @@ def convert_architecture(architecture):
...
+def str_or_none(value):
+    """Normalize value to be a string (or None when "empty")."""
+    if value is None:
+        return None
+    if isinstance(value, str) and value.strip() == "":
+        return None
+    return str(value).strip()
+
+
+def float_or_none(value):
```
infinitewarp, line 49:
```suggestion
def float_or_none(value) -> float | None:
```

infinitewarp (changes requested):
I found one import that seems wrong, and I made some type hint suggestions for the new functions. The former seems important. The latter is more an opinionated nice to have. LMK what you think.

quipucords/fingerprinter/formatters.py
bruno-fs: WTF 🤣 

definitely wrong, lol

thanks for spotting this one

quipucords/fingerprinter/formatters.py
bruno-fs: datetime should not be a possible output, so I only included date and None

quipucords/fingerprinter/runner.py
```diff
@@ -1015,21 +1005,25 @@ def _process_network_fact(self, source: dict, fact: dict) -> dict:
...
             # Set OS information
-            ("etc_release_name", "os_name"),
-            ("etc_release_version", "os_version"),
-            ("etc_release_release", "os_release"),
+            ("etc_release_name", "os_name", formatters.str_or_none),
+            ("etc_release_version", "os_version", formatters.str_or_none),
+            ("etc_release_release", "os_release", formatters.str_or_none),
             # Installed products (name + product/eng ID pairs)
-            ("installed_products", "installed_products"),
+            ("installed_products", "installed_products", formatters.str_or_none),
```
mirekdlugosz, line 1026: Since this is `JSONField` in DB, should we have a formatter that also verifies that string can be parsed as JSON?

quipucords/fingerprinter/runner.py
```diff
@@ -1038,43 +1032,59 @@ def _process_network_fact(self, source: dict, fact: dict) -> dict:
                     else "ip_address_show_ipv4"
                 ),
                 "ip_addresses",
+                formatters.list_or_none,
```
mirekdlugosz, line 1035: That one also is a `JSONField`... so perhaps `installed_products` should be formatted to list or dict instead of string?

mirekdlugosz (commented):
Left two questions. They probably come from my lack of understanding how this entire thing works. Once I have a confirmation there are no more changes in this PR, I will do second round of checking reports for differences.

quipucords/fingerprinter/runner.py
bruno-fs: TY. Camayoc is correct! installed products is indeed a list of dicts. Will fix this and add a unit test covering this one

infinitewarp (approved):
🚀 🌕

bruno-fs:
This is dev-complete, but I decided to mark it as DRAFT again so we don't accidentally merge it before QE's review (cc @ruda / @mirekdlugosz )

mirekdlugosz:
As in previous comment, I've ran standard Camayoc scans against testlab from current `main` and then again using container from this PR (built locally from branch). Then I compared `details.json` and `deployments.json` from all reports. I saw a change around `8.1` / `"8.10"` (as expected).

Other than that, I have not noticed any suspicious changes in these reports.

ruda (approved):

PR #2746, opened 2024-10-21
feat: improve json response for failed token login (Axes "too many failed login attempts")

infinitewarp:
Add a signal handler that is triggered when Axes blocks login auth for the token API due to the user being locked out.

Before this change, if a user attempted login via `/api/v1/token/` but Axes was blocking them, they would receive a 400 status response with a message like the following:

    {"non_field_errors":["Unable to log in with provided credentials."]}

However, that message is not very useful, and we want to make the cause of the auth failure more clear to the user. After this change, a user making the same type of request would receive a 403 status response with a messages like the following:

    {"non_field_errors":["Too many failed login attempts."]}

Relates to JIRA: DISCOVERY-805


bruno-fs (approved):
LGTM

nicolearagao (approved):
LGTM

ruda (approved):

PR #2747, opened 2024-10-29
test: add an integration test for network scan

bruno-fs:
Add a much-needed smoke test for the network scan.

### DIFF
--- a/quipucords/tests/integration/test_network_scan.py
+++ b/quipucords/tests/integration/test_network_scan.py
@@ -1,11 +1,17 @@
 """Integration test for network scan."""
 
 from logging import getLogger
+from unittest.mock import Mock
 
 import pytest
 from django.conf import settings
 
 from api.models import SystemFingerprint
+from constants import DataSources
+from scanner.network.inspect_callback import HOST_DONE, InspectCallback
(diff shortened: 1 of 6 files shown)


quipucords/tests/integration/test_network_scan.py
```diff
@@ -255,3 +261,114 @@ def test_sanity_check_raw_fact_matches(
...
+
+
+@pytest.fixture
+def patched_scan_data(mocker, target_ip, raw_facts):
+    """Patched data for network scan.
+
+    This fixture aims to mock very surgical points on network scan, allowing to exercise
+    most of the code.
+    """
+    global _runner_counter  # noqa: PLW0603
```
infinitewarp, line 292: 🤨 

I think I understand your intent here, but seeing `global` _immediately_ sets off my spidey senses. Maybe we can find a slightly less offensive alternative? If you made this a class-based test method, I think you could use a member variable for the counter instead.

bruno-fs: I can try! Yeah, a class could be a way to get rid of it.

infinitewarp (approved):

PR #2750, opened 2024-11-07
feature: raw fact report

bruno-fs:
Relates to JIRA: DISCOVERY-646


bruno-fs:
@sourcery-ai review

infinitewarp (approved):
I'm sure I've said the same thing in previous conversations, but it always takes me a moment to remember what django-json-agg is doing with our raw facts/inpsect result model whenever I look at it. Certainly not a reason to block this change, more like a sign that our underlying schema is kinda ugly, but we already knew that. Considering what we have now, this should be a good step forward at least to make those raw facts accessible.

PR #2751, opened 2024-11-07
fix: connect scan should fail if no hosts are reachable

infinitewarp:
This resolves a longstanding issue where the UI would present an apparent success status for a network-type connection scan even if all of the hosts in the source were unreachable. Previously, we would only fail if an individual ansible run (hosts are split into groups, and run is invoked once for each group) returned an edge-case final status like "timeout".

Relates to JIRA: DISCOVERY-823

### DIFF
--- a/quipucords/scanner/network/connect.py
+++ b/quipucords/scanner/network/connect.py
@@ -230,6 +230,7 @@ def _connect(  # noqa: PLR0913, PLR0912, PLR0915
         f" with use_paramiko: {use_paramiko} and {forks:d} forks"
     )
     scan_task.log_message(log_message)
+    any_successful_connection = False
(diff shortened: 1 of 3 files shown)


quipucords/scanner/network/connect.py
```diff
@@ -292,6 +293,7 @@ def _connect(  # noqa: PLR0913, PLR0912, PLR0915
         delete_ssh_keyfiles(inventory)
 
         final_status = runner_obj.status
+        any_successful_connection |= scan_task.systems_scanned >= 1
```
bruno-fs, line 296: Oh, I didn't knew | operator worked for booleans! TIL

bruno-fs (approved):
:shipit:

PR #2759, opened 2024-12-02
chore: get rid of the deprecated legacy UI views

bruno-fs:
With the new and standalone UI released we can finally get rid of UI views, templates, js, css and other static files from that UI.

With this change, Django development server can FINALLY run in true standalone mode, without the argument --nostatic.

Please remember to remove these unneeded files from local storage:
```
rm -rf quipucords/client
rm -rf quipucords/quipucords/templates
rm -rf quipucords/staticfiles
```

~~EDIT: This PR removes Django admin and adds an autogenerated schema view (mostly for v2 endpoints).~~

Relates to JIRA: DISCOVERY-777
~~Relates to JIRA: DISCOVERY-726~~

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,15 +48,6 @@ RUN pip install -r requirements.txt
 RUN dnf remove ${BUILD_PACKAGES} -y && \
     dnf clean all
 
-# Fetch UI code
-# TODO FIXME REMOVE THE UI CODE!
-# This bundles an old version of the UI while we are release a new version separately.
-# After we have released the "new" UI code and proven it to be stable, we must remove
-# these commands that bundle the old UI code with the quipucords image.
-COPY Makefile .
-ARG UI_RELEASE="1.8.0"
-RUN --mount=type=secret,id=gh_api_token make fetch-ui -e QUIPUCORDS_UI_RELEASE=${UI_RELEASE}
-
 # Create /deploy
 COPY deploy  /deploy
 
--- a/quipucords/quipucords/urls.py
+++ b/quipucords/quipucords/urls.py
@@ -17,38 +17,9 @@ Including another URLconf
 
 from django.conf.urls import include
 from django.contrib import admin
-from django.contrib.auth import views as auth_views
-from django.urls import path, re_path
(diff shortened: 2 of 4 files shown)


quipucords/quipucords/urls.py
```diff
@@ -17,38 +17,11 @@
...
-        RedirectView.as_view(url="/client/docs/install.html", permanent=False),
-        name="docs",
-    ),
-    # static files (*.css, *.js, *.jpg etc.)
-    re_path(
-        r"^(?!/?client/)(?P<path>.*\..*)$",
-        RedirectView.as_view(url="/client/%(path)s", permanent=False),
-        name="client",
-    ),
+    path("", RedirectView.as_view(url="/admin/login", permanent=False), name="home"),
```
bruno-fs, line 26: now that we don't have the template for login, the auth view for login won't work.

I decided to point home to admin just so we can use it to login for local development.
I know we don't actually have the intention of keeping admin (having it is an accident AFAIK), but not having a session login makes manually test the api a bit harder.

Quick alternatives I could think right now:
- I could enable django admin and views conditionally, only on debug mode
- Could add a simple replacement for login view (and have that view only available on debug mode)

infinitewarp: My ranked votes are 1) enable admin/views and this `path` line only if debug is True, or 2) just axe the admin and this path entirely (does anybody use admin?)

Dockerfile
```diff
@@ -48,15 +48,6 @@ RUN pip install -r requirements.txt
 RUN dnf remove ${BUILD_PACKAGES} -y && \
     dnf clean all
 
-# Fetch UI code
-# TODO FIXME REMOVE THE UI CODE!
-# This bundles an old version of the UI while we are release a new version separately.
-# After we have released the "new" UI code and proven it to be stable, we must remove
```
infinitewarp, line 54: I'm just leaving this comment here as an unresolved thread so we don't accidentally merge this _before_ we have finished shipping 1.11. 😉

infinitewarp (approved):

PR #2761, opened 2024-12-04
fix: forbid credential delete if assigned to a source (v2 credentials API)

infinitewarp:
This PR effectively restores the behavior of the v1 credentials API that we missed when we added the v2 credentials API. In addition to returning a 400 error when the requested credential is related to a source, the API also now returns a 404 error when the requested credential ID does not exist.

Thanks for finding this issue, @mirekdlugosz !


nicolearagao (approved):
LGTM

PR #2768, opened 2024-12-11
fix: update scan task/job status to failed when unhandled exception occurs in celery tasks

infinitewarp:
Relates to JIRA: DISCOVERY-800
Relates to JIRA: DISCOVERY-838

### DIFF
--- a/quipucords/scanner/tasks.py
+++ b/quipucords/scanner/tasks.py
@@ -85,7 +85,7 @@ def set_scan_task_failure_on_exception(func):
     @functools.wraps(func)
     def wrapper(*args, **kwargs):
         if not (scan_task_id := kwargs.get("scan_task_id", None)):
-            logger.warning(
+            raise NotImplementedError(
                 f"Missing scan_task_id kwarg for decorated function {func.__name__}"
             )
         try:
@@ -121,7 +121,7 @@ def set_scan_job_failure_on_exception(func):
     @functools.wraps(func)
     def wrapper(*args, **kwargs):
         if not (scan_job_id := kwargs.get("scan_job_id", None)):
-            logger.warning(
+            raise NotImplementedError(
                 f"Missing scan_job_id kwarg for decorated function {func.__name__}"
             )
         try:
--- a/quipucords/scanner/tasks.py
+++ b/quipucords/scanner/tasks.py
@@ -73,8 +73,28 @@ def get_task_runner_class(source_type, scan_type) -> type[ScanTaskRunner]:
 def celery_run_task_runner(
     self: celery.Task, scan_task_id: int, source_type: str, scan_type: str
 ) -> tuple[bool, int, str]:
-    """Wrap _celery_run_task_runner to call it as an async Celery task."""
-    return _celery_run_task_runner(self, scan_task_id, source_type, scan_type)
+    """
+    Wrap _celery_run_task_runner to call it as an async Celery task.
+
+    Also wrap the call with try-except to handle any unexpected exceptions
+    and update the task's status to "failed" if any exception was raised.
+    """
+    try:
+        return _celery_run_task_runner(self, scan_task_id, source_type, scan_type)
+    except Exception:  # noqa: BLE001
+        logger.exception(
+            f"Unexpected exception in celery_run_task_runner (ScanTask {scan_task_id})"
+        )
+
+    try:
+        scan_task = ScanTask.objects.get(id=scan_task_id)
+        scan_task.status_fail(
+            f"Unexpected exception in celery_run_task_runner (ScanTask {scan_task_id})"
+        )
+    except Exception:  # noqa: BLE001
+        logger.exception(
+            f"Failed to record celery_run_task_runner failure (ScanTask {scan_task_id})"
+        )
 
 
 def _celery_run_task_runner(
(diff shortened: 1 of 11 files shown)


ruda (approved):

quipucords/scanner/tasks.py
```diff
@@ -69,32 +70,79 @@ def get_task_runner_class(source_type, scan_type) -> type[ScanTaskRunner]:
...
-        scan_task.status_fail(
-            f"Unexpected exception in celery_run_task_runner (ScanTask {scan_task_id})"
-        )
-    except Exception:  # noqa: BLE001
-        logger.exception(
-            f"Failed to record celery_run_task_runner failure (ScanTask {scan_task_id})"
-        )
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        scan_task_id = args[1] if len(args) > 1 else kwargs.get("scan_task_id", None)
```
bruno-fs, line 87: I don't love the idea of inferring the argument by position as default. I'd consume it as a kwarg and force decorated functions to use it as a kwarg
you probably know how to do this, but just in case you don't, here's how to:

```python
def some_func(*, keyword_only_arg):
    ...
    ````

Comment (bruno-fs) [reply quipucords/scanner/tasks.py:87]: and the reason I don't like the positional argument is because someone might change it in the future and it could become a bit hard to debug why things weren't working as expected.

Comment (infinitewarp) [reply quipucords/scanner/tasks.py:87]: I like the suggestion, but I'm not sure if the `*` keyword enforcer works safely with the way Celery task are defined and invoked, though, or at least when we set `bind=True` in `@celery.shared_task`. Using `bind` defers some calling logic to Celery internals to set an _additional_ argument at the start of the function calls, and that's why we have `self` on some task functions like this:

```python
@celery.shared_task(bind=True)
@set_scan_task_failure_on_exception
def fingerprint(self: celery.Task, scan_task_id: int) -> tuple[bool, int, str]:
    ...
```
We use that `self` to interrogate things about the task itself, such as its ID so we can check if it's been canceled.

Does that `self` work with `*`? I cannot tell yet. I am trying to parse through the Celery and Kombu code to try to understand how its `bind` argument actually works, but it quickly gets complicated.  😕 

I have at least locally (and maybe only temporarily) replaced that ugly one-liner with this as a potential safety check if `*` doesn't work as expected:

```python
if not (scan_task_id := kwargs.get("scan_task_id", None)):
    logger.warning(
        f"Missing scan_task_id kwarg for decorated function {func.__name__}"
    )
```

infinitewarp: Further testing with `*` reveals things to be… still complicated. For example, having this task defined:

```python
@celery.shared_task(bind=True)
@set_scan_job_failure_on_exception
def finalize_scan(self: celery.Task, *, scan_job_id: int):
    """Wrap _finalize_scan to call it as an async Celery task."""
    return _finalize_scan(self=self, scan_job_id=scan_job_id)
```

One would expect `scan_job_id` to be a required keyword argument, right? However, we never invoke this function directly. Because it's a Celery task, we indirectly invoke it by requesting it to be run through `finalize_scan.delay(...)`, and that `delay` function does not enforce the `*` magic, as far as I can tell. So, the caller can "successfully" call `finalize_scan` without `scan_job_id` as a keyword argument, and it only fails later (I think) when handled by the Celery worker. This means we need to be _very careful_ in how we write unit tests where we don't fully execute the Celery tasks (or use `CELERY_TASK_ALWAYS_EAGER`). Totally workable, but something to keep in mind.

bruno-fs: I see. Dammit, celery 😓 

Thanks for trying, tho!

quipucords/scanner/tasks.py
```diff
@@ -69,16 +70,94 @@ def get_task_runner_class(source_type, scan_type) -> type[ScanTaskRunner]:
...
+    If we merge these two decorators, we probably need a clever way to determine
+    whether the relevant object to "fail" is a ScanJob or a ScanTask.
+    """
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        if not (scan_task_id := kwargs.get("scan_task_id", None)):
+            logger.warning(
+                f"Missing scan_task_id kwarg for decorated function {func.__name__}"
+            )
```
bruno-fs, line 90: same observation I made for scan_job decorator: https://github.com/quipucords/quipucords/pull/2768/commits/9907a930aa4414e9e5cb5547b6ec89037ca489e6#r1904196237

quipucords/scanner/tasks.py
```diff
@@ -117,7 +120,10 @@ def set_scan_job_failure_on_exception(func):
 
     @functools.wraps(func)
     def wrapper(*args, **kwargs):
-        scan_job_id = args[1] if len(args) > 1 else kwargs.get("scan_job_id", None)
+        if not (scan_job_id := kwargs.get("scan_job_id", None)):
+            logger.warning(
+                f"Missing scan_job_id kwarg for decorated function {func.__name__}"
+            )
```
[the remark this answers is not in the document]
infinitewarp: Ah, this is a good point. I originally wrote the conditional check with this relaxed approach because I was thinking that I would have _one_ decorator that could be made generic and not all uses would necessarily provide all arguments, but I abandoned that path and kept the two decorators purpose-specific. I will take another close look at all the current use cases, and I will probably make this check more aggressive like you suggested.

bruno-fs (approved):

PR #2772, opened 2025-01-06
DONT MERGE THIS

bruno-fs:

PR #2773, opened 2025-01-06
build: update pyproject.toml to work with poetry 2

infinitewarp:


.github/workflows/test.yml
```diff
@@ -29,6 +29,7 @@ jobs:
 
       - name: Check dependencies
         run: |
+          poetry self add poetry-plugin-export
```
mirekdlugosz, line 32: poetry 2 claims to support automatic per-project plugin installation: https://github.com/python-poetry/poetry/pull/9547

wonder if it makes sense to try that instead of adding plugin manually?

mirekdlugosz:
Poetry 2.0 also supports `[project.dependencies]`, with some non-PEP metadata to be stored in `[tool.poetry.dependencies]`. `[project.dependencies]` is a list of strings, and Python package spec does not support caret, so that would be larger change. But maybe we should do everything in one fell swoop?

See https://github.com/python-poetry/poetry/pull/9135 and https://python-poetry.org/docs/dependency-specification/ (upstream PR is explicit what parts of documentation are changed).

bruno-fs:
about `[dependencies]`, feel free to steal the code already formated on my WIP/abandoned branch

https://github.com/quipucords/quipucords/pull/2772/files#r1905465435

(I abandoned the branch when I realized poetry 2 still requires its own section for dev dependencies 😢 )

.github/workflows/test.yml
bruno-fs: it works as they claim indeed

so instead of adding this command on the workflow, we can just add the relevant section on pyproject toml (the part requiring poetry 2 on poetry.tool might be required as well)
https://github.com/quipucords/quipucords/pull/2772/files#r1905468347

infinitewarp: Copying here for the public record [what I said in Slack](https://redhat-internal.slack.com/archives/C02QSNF1UKE/p1736271234940059?thread_ts=1736264611.225519&cid=C02QSNF1UKE):

I think I'm going to apply a hybrid approach for the export plugin. I want to _keep_ the explicit `poetry self add poetry-plugin-export` command in the GitHub job, but I'm _also_ going to declare it as a dependency in the `pyproject toml`. Why? If you only declare a plugin it as a dependency, you need to do a full `poetry install` which included all of the project dependencies to get the plugin working. That's overkill and completely unnecessary for the GitHub check-dependencies job, it slows it down, and it causes some other errors.

On your local system you're probably `poetry install`ing everything anyway. So, declaring as a dependency will be fine there. I tried other options like `poetry self install` but that doesn't get the plugin even though you'd think it would. I've also experimented with the new `sync` command (which is just the old `install --sync`) and a few other flags, but I have found no path aside from the direct command in the original PR to install only the `tool.poetry.requires-plugins` and not the full set of project dependencies. :shrug:

bruno-fs:
We might want to wait before merging this PR.

Renovate/minkmaker (which is used by konflux to manage lockfiles) don't have support for poetry 2

https://github.com/renovatebot/renovate/issues/33509

(or maybe we do merge it and disable renovate support for python dependencies)

infinitewarp:
Poetry 2 is dead (to us). Long live uv. 🫡

PR #2775, opened 2025-01-07
migrate from poetry to uv

bruno-fs:
poetry 2 was recently released and it can use the now standardized
"project" section to add package metadata and declare dependencies
(PEP-621), while, alternatively, still supports its old
tool.poetry.dependency section. This prompted us to adapt pyproject.toml
to this change. During the update process we noticed poetry 2 still uses
its old tool.poetry.dev.dependencies to handle dev dependencies instead
of the now standardized dependency-group.dev (PEP-735).

uv, however, already adopted both standards. 

Besides that, uv is extremely FAST when it comes to dependency resolution and installation.
This, paired with the recent update of pybuild-deps, greatly accelerate `make lock-requirements` and `make update-requirements`.

To give you a taste of the performance, this is how long `make update-requirements` took on my system **with a warm cache**:
- before this PR: 86 seconds
- with this PR: 10 seconds

Its sibling target, `make lock-requirements`, is even faster (6 seconds in my machine 💻)!

### MIGRATING TO UV
Before checking out to this branch, destroy your poetry environment
```
poetry env remove --all
```
Then...
1. Install uv (see https://docs.astral.sh/uv/getting-started/installation/#installation-methods).

2. Optionally install python 3.13 with uv
```
uv python install 3.13
```
3. Finally, create the uv virtual environment
```
uv sync
```
4. Configure your IDE python interpreter to point to the output of `uv run which python`.


nicolearagao (approved):
went through the steps, everything works as intended.

mirekdlugosz (approved):

abellotti:
just tried it with a brew installed uv on macos 15.3.2 👍

abellotti (approved):

PR #2793, opened 2025-01-21
feature: regenerate cached fingerprint and cached csv if missing when requested

infinitewarp:
This change automatically recreates files in the `cached_reports` directory if they are missing when any report download that needs them is requested. The initial download request may immediately return an error message (we should consider making the UI handle this more gracefully), but if you wait a few seconds for a newly spawned async task to complete and then try the download again, you should get the completed report. This works for operations in the UI and the CLI.

Additionally, this change adds a management command `regenerate_missing_report_cache_files` that attempts to identify and fix all missing `cached_reports` files. Running this command still uses Celery tasks. So, be sure you have Celery workers running when you call it or else it will hang forever while it waits for Celery task results.

Please note: Normally nothing in our system should delete `cached_reports` files while they might still be needed. The use case for this change includes installations that were recovered from a previous setup using _only_ a database backup and the requisite `secret.txt` file.

Also note: quipucords server ensures that the `cached_reports` directory exists at startup and only at startup. These instructions tell you to delete the contents of that directory, not the directory itself. If you delete the whole directory, please restart the server and Celery workers.

What follows are documented steps that I performed manually to exercise the new behaviors.

Breaking and fixing typical reports made through the UI:

- starting from new database
- add a network credential and source for known working address
- create a scan using that credential and source
- wait for scan to complete
- immediately delete contents of `cached_reports` directory (should be single JSON file)
- try to download the report
- observe error
- wait a few seconds, and try to download the report again
- verify that it downloads successfully
- repeat last 5 steps (directory should have 1 JSON file and 1 CSV file)

Breaking and fixing uploaded reports:

- downloaded the report from the previous section
- extract the details.json file
- use the CLI to upload it
    ```bash
    qpc report upload --json-file ~/Desktop/report_id_2/details.json
    ```
- check the job given in the response, expecting `"status": "completed",`
    ```bash
    qpc scan job --id 8
    ```
- delete the contents of `cached_reports` directory
- attempt to download the report
    ```bash
    qpc report download --report 3 --output-file ~/Desktop/report-3.tar.gz
    ```
- observe error
- wait a few seconds, and try to download the report again
- verify that it downloads successfully

Breaking and fixing merged reports:

- get the report IDs generated by the previous steps (e.g. 1 and 2)
- delete the contents of `cached_reports` directory
- use the CLI to merge the previous reports
    ```bash
    qpc report merge --report-ids 2 3
    ```
- check the job given in the reponse, expecting `"status": "failed"`
    ```bash
    qpc scan job --id 9
    ```
- wait a few seconds, and try the merge command again
    ```bash
    qpc report merge --report-ids 2 3
    ```
- check the job given in the reponse, expecting `"status": "completed"`
    ```bash
    qpc scan job --id 10
    ```
- download the report
    ```bash
    qpc report download --report 6 --output-file ~/Desktop/report-6.tar.gz
    ```
- verify that it downloads successfully

Using the management command to fix all broken but fixable reports:

- delete the contents of `cached_reports` directory
- use the management command
    ```bash
    python quipucords/manage.py regenerate_missing_report_cache_files
    ```
- download any of the reports as described in the previous steps
- verify that they download successfully

Relates to JIRA: DISCOVERY-887

### DIFF
--- a/quipucords/api/deployments_report/model.py
+++ b/quipucords/api/deployments_report/model.py
@@ -19,6 +19,7 @@ from fingerprinter.constants import (
     PRODUCTS_KEY,
     SOURCES_KEY,
 )
+from utils.misc import is_valid_cache_file
 
 logger = logging.getLogger(__name__)
 
@@ -61,6 +62,11 @@ class DeploymentsReport(BaseModel):
     )
 
     @property
+    def cached_fingerprints_file_exists(self) -> bool:
+        """Return True if cached fingerprints file exists."""
+        return is_valid_cache_file(self.cached_fingerprints_file_path)
+
+    @property
     def cached_fingerprints(self) -> dict | None:
         """Return cached fingerprints data if it exists."""
         if not self.cached_fingerprints_file_path:
@@ -84,6 +90,7 @@ class DeploymentsReport(BaseModel):
                 self.id,
                 file_path,
             )
+            self._rerun_latest_fingerprint()
             raise
 
     @cached_fingerprints.setter
@@ -97,6 +104,11 @@ class DeploymentsReport(BaseModel):
         with file_path.open("w") as f:
             json.dump(data, f)
         self.cached_fingerprints_file_path = file_path
+
+    @property
+    def cached_csv_file_exists(self) -> bool:
+        """Return True if cached csv file exists."""
+        return is_valid_cache_file(self.cached_csv_file_path)
 
     @property
     def cached_csv(self):
@@ -126,6 +138,7 @@ class DeploymentsReport(BaseModel):
                 self.id,
                 self.cached_csv_file_path,
             )
+            self._rerun_latest_fingerprint()
             raise
 
     @cached_csv.setter
@@ -139,6 +152,41 @@ class DeploymentsReport(BaseModel):
         with file_path.open("w") as f:
             f.write(data)
         self.cached_csv_file_path = file_path
+
+    def _rerun_latest_fingerprint(self, wait=False) -> None:
+        """
+        Trigger the latest fingerprint task to rerun.
+
+        Why would you want to do this? If you expect the cached_fingerprints or
+        cached_csv data to exist but either is missing, you may recreate them by
+        calling this method.
+
+        :param wait: if true, wait for task to complete
+        """
+        logger.info("Rerunning the last fingerprint task for %s", self)
+
+        # local imports to avoid possible import loop
+        from api.scantask.model import ScanTask
+        from scanner import tasks
+
+        scan_task = (
+            ScanTask.objects.filter(
+                job__report__deployment_report_id=self.id,
+                scan_type=ScanTask.SCAN_TYPE_FINGERPRINT,
+            )
+            .order_by("-id")
+            .first()
+        )
+        if not scan_task:
+            # It is unclear how this is possible, but just in case, raise an
+            # exception because we cannot proceed here without a ScanTask,
+            # and we may not want to (or be able to) create a new one on demand.
+            raise ScanTask.DoesNotExist(f"No fingerprint ScanTask is related to {self}")
+
+        logger.info("Found %s for %s", scan_task, self)
+        fingerprint_task = tasks.fingerprint.delay(scan_task_id=scan_task.id)
+        if wait:
+            fingerprint_task.get()
 
 
 @receiver(post_delete, sender=DeploymentsReport)
--- a/quipucords/api/management/commands/regenerate_missing_report_cache_files.py
+++ b/quipucords/api/management/commands/regenerate_missing_report_cache_files.py
@@ -1,0 +1,130 @@
+"""Regenerate missing DeploymentsReport cache files."""
+
+from django.core.management.base import BaseCommand, CommandError
+
+from api.deployments_report.model import DeploymentsReport
+from api.deployments_report.tasks import generate_and_save_cached_csv
+from utils.misc import is_valid_cache_file
+
+
+class Command(BaseCommand):
+    """Django management command to regenerate missing DeploymentsReport cache files."""
+
+    help = "Regenerate missing DeploymentsReport cache files"
+
+    def write_info(self, message):
+        """Write info message to stdout."""
+        self.stdout.write(message)
+
+    def write_warning(self, message):
+        """Write warning message to stdout."""
+        self.stdout.write(self.style.WARNING(message))
+
+    def write_success(self, message):
+        """Write success message to stdout."""
+        self.stdout.write(self.style.SUCCESS(message))
+
+    def find_reports_with_missing_cached_data(self) -> tuple[set, set]:
+        """
+        Find reports with missing cached data and return their IDs.
+
+        Returned tuple contains 1) the set of IDs with missing fingerprints
+        and 2) the set of IDs with missing CSV data.
+        """
+        missing_fingerprints_ids = set()
+        missing_csv_ids = set()
+        deployments_reports = DeploymentsReport.objects.filter(
+            status=DeploymentsReport.STATUS_COMPLETE
+        ).values("id", "cached_fingerprints_file_path", "cached_csv_file_path")
+        self.write_info(
+            f"Found {len(deployments_reports)} completed DeploymentsReports to check."
+        )
+        for deployments_report in deployments_reports:
+            # Note: We check for presence of value before checking is_valid_cache_file.
+            # If the value is None, then the file was never written, and we do not care.
+
+            if deployments_report[
+                "cached_fingerprints_file_path"
+            ] and not is_valid_cache_file(
+                deployments_report["cached_fingerprints_file_path"]
+            ):
+                self.write_warning(
+                    f"{deployments_report['cached_fingerprints_file_path']} is missing "
+                    f"for DeploymentsReport {deployments_report['id']}."
+                )
+                missing_fingerprints_ids.add(deployments_report["id"])
+
+            if deployments_report["cached_csv_file_path"] and not is_valid_cache_file(
+                deployments_report["cached_csv_file_path"]
+            ):
+                self.write_warning(
+                    f"{deployments_report['cached_csv_file_path']} is missing "
+                    f"for DeploymentsReport {deployments_report['id']}."
+                )
+                missing_csv_ids.add(deployments_report["id"])
+
+        return missing_fingerprints_ids, missing_csv_ids
+
+    def generate_fingerprints(
+        self, deployments_report_ids: set[int]
+    ) -> tuple[list, list]:
+        """Generate cached fingerprints for the given deployments report IDs."""
+        success_ids = []
+        failure_ids = []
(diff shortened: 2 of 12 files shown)


quipucords/utils/misc.py
```diff
@@ -34,3 +36,16 @@ def sanitize_for_utf8_compatibility(value):
     if isinstance(value, tuple):
         return tuple(sanitize_for_utf8_compatibility(value) for value in value)
     return value
+
+
+def file_exists(file_path: str | None) -> bool:
```
mirekdlugosz, line 41: The name suggests it's a wrapper around `path.exists()`, but it checks more than just that.

What about `file_readable` (I guess non-existing file is not readable by definition)? Or `file_accessible`? Or `valid_cache_file` (that in turn implies more than this function can in fact deliver)?

quipucords/utils/misc.py
```diff
@@ -34,3 +36,16 @@ def sanitize_for_utf8_compatibility(value):
...
+    """Check if a file exists, has non-zero size, and is readable."""
+    if not file_path:
+        return False
+    path = Path(file_path).absolute()
+    return (
+        path.exists()
+        and path.is_file()
+        and path.stat().st_size > 0
```
mirekdlugosz, line 49: I'm not _really_ worried, but what if file is 0 size and not writable?

That check will return False, and I assume we will try to write to file. But with writable bit turned off, we will fail. Not sure if cache generating code handles that.

quipucords/api/management/commands/regenerate_missing_report_cache_files.py
```diff
@@ -0,0 +1,93 @@
...
+    def find_reports_with_problematic_paths(self) -> set:
+        """Find reports with problematic paths and return their IDs."""
+        report_ids = set()
+        reports_paths = DeploymentsReport.objects.filter(
+            status=DeploymentsReport.STATUS_COMPLETE
+        ).values("id", "cached_fingerprints_file_path", "cached_csv_file_path")
+        self.write_info(
+            f"Found {len(reports_paths)} complete DeploymentsReports to check."
+        )
+        for report_paths in reports_paths:
```
mirekdlugosz, line 35: Just a suggestion: `report_paths` and `reports_paths` is super confusing. What if instead of `report_paths`, we call it `deployment_report`? I know, technically this is not a deployment report but a dict with a subset of deployment report data, but that would express the intent, wouldn't it?

quipucords/api/management/commands/regenerate_missing_report_cache_files.py
```diff
@@ -0,0 +1,93 @@
...
+            if not report_paths["cached_fingerprints_file_path"]:
+                self.write_warning(
+                    "cached_fingerprints_file_path not set for DeploymentsReport "
+                    f"{report_paths["id"]}"
+                )
+            elif not file_exists(report_paths["cached_fingerprints_file_path"]):
+                self.write_warning(
+                    f"{report_paths["cached_fingerprints_file_path"]} is missing."
+                )
+                report_ids.add(report_paths["id"])
```
mirekdlugosz, line 45: Since that `if - elif` is pretty much the same, they just work on different keys, would it make sense to call them in a loop or extract to function?

quipucords/api/management/commands/regenerate_missing_report_cache_files.py
```diff
@@ -0,0 +1,93 @@
...
+                    self.stdout.write(self.style.ERROR(str(e)))
+                    failed_report_ids.append(report_id)
+        else:
+            self.write_success(
+                "All DeploymentReport cached_fingerprints_file_path and "
+                "cached_csv_file_path files exist."
+            )
+
+        if set(report_ids) - set(failed_report_ids):
+            successes = set(report_ids) - set(failed_report_ids)
```
mirekdlugosz, line 84:
```suggestion
        successes = set(report_ids) - set(failed_report_ids)
        if successes:
```

quipucords/api/management/commands/regenerate_missing_report_cache_files.py
```diff
@@ -0,0 +1,93 @@
...
+            self.write_success(
+                f"Successfully regnerated files for "
+                f"{len(successes)} DeploymentsReports."
+            )
+        if failed_report_ids:
+            self.write_warning(
+                f"Failed to regenerate files for "
+                f"{len(failed_report_ids)} DeploymentsReports."
```
mirekdlugosz, line 92: I don't know if that's useful, but we could also print ids of reports where we failed to regenerate cache. id is not the easiest thing to handle, but it does feel better than blank "some failed, good luck".

quipucords/api/management/commands/regenerate_missing_report_cache_files.py
```diff
@@ -0,0 +1,93 @@
...
+                "cached_csv_file_path files exist."
+            )
+
+        if set(report_ids) - set(failed_report_ids):
+            successes = set(report_ids) - set(failed_report_ids)
+            self.write_success(
+                f"Successfully regnerated files for "
+                f"{len(successes)} DeploymentsReports."
+            )
+        if failed_report_ids:
```
mirekdlugosz, line 89: If there are failures, please make sure that command returns non-zero status. This will help discovery-ci to decide if we should proceed (and likely encounter issues later), or stop right here.

I guess if we want to be fancy, we can exit 1 when there are any failures, and exit 2 when there are **only** failures. Not sure if there's a point in differentiating these two situations.

quipucords/api/management/commands/regenerate_missing_report_cache_files.py
```diff
@@ -0,0 +1,93 @@
...
+                try:
+                    report._rerun_latest_fingerprint(wait=True)
+                except Exception as e:  # noqa: BLE001
+                    self.stdout.write(self.style.ERROR(str(e)))
+                    failed_report_ids.append(report_id)
+        else:
+            self.write_success(
+                "All DeploymentReport cached_fingerprints_file_path and "
+                "cached_csv_file_path files exist."
+            )
```
mirekdlugosz, line 81: Can't we return here? No further conditions are going to be met, are they?

Personally, I would reverse that construct. If there are no `report_ids`, call `write_success()` and return. Remove `else` and one level of indentation from the rest.

mirekdlugosz (changes requested):
I let my mind wander a bit, hence my comments might seem a bit chaotic and nitpicky. Sorry about that.

The one really important thing is ensuring that if we couldn't regenerate the cache, then command returns non-zero status code. I don't really know, but I think in such case discovery-ci should bail out and fail the entire pipeline. Without status code, we can't easily catch that case.

quipucords/api/deployments_report/model.py
```diff
@@ -140,6 +153,37 @@ def cached_csv(self, data):
...
+            )
+            .order_by("-id")
+            .first()
+        )
+        logger.info("Found %s for %s", scan_task, self)
+        if not scan_task:
+            raise ScanTask.DoesNotExist(f"No fingerprint ScanTask is related to {self}")
+        fingerprint_task = tasks.fingerprint.delay(scan_task_id=scan_task.id)
+        if wait:
+            fingerprint_task.get()
```
bruno-fs, line 185: What if instead of invoking celery when `wait` is true we just run the function synchronously, w/o celery?

quipucords/utils/misc.py
infinitewarp: Good point on the name. This helper function started simpler, thus the simple name, and grew in complexity as I worked through the issues. I have renamed this function to `is_valid_cache_file`.

quipucords/utils/misc.py
infinitewarp: That's a good point about the writable bit. I've added that to the check here, but ultimately I consider handling "failure to write" to be outside the scope of this issue. Failing to write to the file will certainly cause Celery to blow up exceptions elsewhere that should be very obvious to the reader.

quipucords/api/management/commands/regenerate_missing_report_cache_files.py
infinitewarp: I believe the current version of the last line looks like
```
for deployments_report in deployments_reports:
```
which more accurately expresses the intent as you described.

The preceding `self.write_info` lines were removed.

quipucords/api/management/commands/regenerate_missing_report_cache_files.py
infinitewarp: I do not remember what the else/elif here was, but it's gone now.

quipucords/api/management/commands/regenerate_missing_report_cache_files.py
infinitewarp: These set assignments and related conditional messaging are gone or rewritten.

quipucords/api/management/commands/regenerate_missing_report_cache_files.py
infinitewarp: Text output for this command is a lot more useful now. Here's a screenshot of the ending output from a recent run containing some successes and failures.

command-output-example

quipucords/api/management/commands/regenerate_missing_report_cache_files.py
infinitewarp: See also the screenshot in the previous comment. The final red line is the result of invoking Django's `CommandError` which is the prescribed method to automatically cause the command to exit `1`, which I only call if there are nonzero failures. Otherwise, it automatically exits `0`.

I didn't add support for different numbers in different situations, though, as that would require a custom `CommandError` subclass or invoking `sys.exit` directly, but I can add one if you feel strongly about it.

quipucords/api/management/commands/regenerate_missing_report_cache_files.py
infinitewarp: I removed those "all" message in favor of more detailed individual messages. I don't remember what was in the related else code, but I suspect it's gone now too.

quipucords/api/deployments_report/model.py
infinitewarp: Unfortunately, we can't invoke `tasks.fingerprint` directly because we defined it as a task with `bind=True` which means it _requires_ a `self` argument pointing to a `celery.Task` which we don't have when calling it synchronously. That `self` is passed into lower layers that check if the `ScanJob` has been cancelled/revoked. Even though that check is not relevant to us in this context, it is not trivial to refactor it out.

quipucords/api/management/commands/regenerate_missing_report_cache_files.py
mirekdlugosz: Returning non-zero is the one that truly matters, thanks.

I don't feel strongly about differentiating failures. It felt like something that could be useful, so maybe could be done if that's easy enough, but I don't have any immediate use-case in mind.

mirekdlugosz (approved):
Looks good, thanks :+1:

bruno-fs (approved):

ruda (approved):

PR #2805, opened 2025-02-13
feat: simplify host validation & add IPv6 support

nicolearagao:
Refactored `validate_ipaddr_list` to streamline host validation by breaking down logic into smaller methods, improving readability and maintainability while preserving functionality.

- Replaced regex-based CIDR and IP validation with Python's built-in `ipaddress` library for more accurate and reliable checks
- Normalized Ansible ranges using Ansible's private function `expand_hostname_range()`, reducing complexity
- Improved FQDN validation for hostnames by leveraging the `FQDN` library

By relying on well-tested libraries instead of custom regex patterns, this change enhances correctness and ensures better long-term maintainability.

Relates to JIRA: DISCOVERY-394


quipucords/tests/api/source/test_source.py
```diff
@@ -2090,13 +2091,8 @@ def test_create_invalid_hosts(self, client_logged_in, net_cred):
         """Test invalid host patterns."""
         hosts = [
             "192.1..2",
-            "192.01.5.10",
             "192.1.5.1/",
```
nicolearagao, line 2094: Some hosts that were considered invalid in the previous implementation are now technically valid. Even if they are not valid IPs, they might be considered valid according to the FQDN hostname validation rules.

quipucords/tests/api/source/test_source.py
```diff
@@ -1250,7 +1251,7 @@ def test_negative_create_non_network_with_excluded_hosts(
     @pytest.mark.parametrize(
         "source_type", (ds for ds in DataSources if ds != DataSources.NETWORK)
     )
-    @pytest.mark.parametrize("hosts_range", ("1.2.3.4/5", "1.2.3.[0:255]"))
+    @pytest.mark.parametrize("hosts_range", ("1.2.3.4/30", "1.2.3.[0:255]"))
```
nicolearagao, line 1254: The CIDR range "1.2.3.4/30" was causing these tests to time out. In the previous implementation, we converted CIDR ranges to Ansible ranges, but now we expand both. The given range includes 8,388,608 IP addresses, so I replaced it with a smaller one. As a team, we agreed to address potential performance issues caused by CIDR expansion in a separate story. Given this, I considered the change acceptable, as it does not impact the test’s evaluation.

quipucords/api/source/serializer.py
```diff
@@ -273,228 +276,130 @@ def validate_name(name):
...
                 )
-        if not host_errors:
-            return normalized_hosts
-        error_message = [error.detail.pop() for error in host_errors]
-        raise ValidationError(error_message)
+            except ValidationError as error:
+                host_errors.append(error)
+
+        if host_errors:
+            error_message = [msg for error in host_errors for msg in error.detail]
```
nicolearagao, line 310: I'm not entirely satisfied with how the UI displays invalid hosts and their corresponding messages. While it's not a regression—since the previous implementation handled it in a similar way—I see room for improvement. Initially, I attempted to adjust this on the server side to control the display format, but that approach resulted in concatenating multiple invalid hosts into a single long string. I believe this change should instead be handled on the client side.

Example with two invalid hosts:
Screenshot from 2025-03-06 11-53-33

infinitewarp: I think this example output is fine. Yes it is ugly and difficult to read, but it is no worse than most other current UI error messages. IMO, _all_ the UI error handling needs major attention, but that is outside of our scope here.

lockfiles/requirements-build.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with Python 3.13
+# This file is autogenerated by pip-compile with Python 3.12
```
infinitewarp, line 2: You may want to check and rebuild your local dev environment since it looks like your poetry may be using the older Python version.

infinitewarp: That being said, I believe 3.12 technically should still work, but I think the rest of us have moved to 3.13, and we started explicitly changing support for 3.13 a couple months ago (https://github.com/quipucords/quipucords/pull/2769).

bruno-fs: one could say this is a good moment to install uv and then...

```
uv python install 3.13 && uv python pin 3.13
uv tool install 'poetry<2'
poetry env remove --all && poetry env use $(uv python find) && poetry install
```

😉

quipucords/api/source/serializer.py
bruno-fs: definitely should be handled on client side, but I wouldn't sweat on "fixing" if we keep using the toast. The proper fix should be done in the context of the proper error handling in forms, rendering inline errors as PF documentation suggests:

https://v5-archive.patternfly.org/components/forms/form#invalid-with-form-alert

image

the "alert" should be mapped to `non_field_errors`, while each form element should be mapped to its respective error.

bruno-fs (approved):
LGTM, but please don't forget to use a new venv with python 3.13 and run make lock-requirements

quipucords/api/source/serializer.py
bruno-fs: added a link to above comment to DISCOVERY-874

quipucords/api/source/serializer.py
```diff
@@ -273,228 +276,130 @@ def validate_name(name):
         return name
 
     @staticmethod
-    def validate_ipaddr_list(hosts):  # noqa: PLR0912, PLR0915, C901
-        """Make sure the hosts list is present and has valid IP addresses."""
-        ipaddr_list = hosts
-        if isinstance(ipaddr_list, list):
-            ipaddr_list = [item for item in ipaddr_list if item]
+    def validate_ipaddr_list(hosts):
```
infinitewarp, line 279: This is maybe a minor nitpick, but to make this more readable, I would rename this function to `validate_addresses` because the current title implies that only IP addresses are allowed or expected, and the `_list` suffix is redundant with a plural name. If you want to be extra clear to the reader, we could add type hints like:

```suggestion
    def validate_addresses(addresses: list) -> list:
```

…and I would rename `hosts` (and related variables like `normalized_hosts`) to something like `addresses` (and `normalized_addresses`) since the original name may also be slightly confusing or _contradict_ the original function name. Thinking as a naive reader might: "Are these IP addresses? Or hostnames? Or actual hosts (systems)? Or both or neither?"

quipucords/api/source/serializer.py
```diff
@@ -273,228 +276,130 @@ def validate_name(name):
...
+        Validate a list of IP addresses, CIDRs, hostnames, and Ansible ranges.
 
-        if not isinstance(ipaddr_list, list):
+        Returns a list of normalized hosts if valid, otherwise raises ValidationError.
+        """
+        # Initial List Processing
+        if not isinstance(hosts, list):
             raise ValidationError(_(messages.SOURCE_HOST_MUST_BE_JSON_ARRAY))
 
+        ipaddr_list = [host for host in hosts if host]
```
infinitewarp, line 289: It looks like `ipaddr_list` is never used again after we use this line to filter out falsy values. I would simply reuse the original input variable name instead of creating a new one:

```suggestion
        hosts = [host for host in hosts if host]
```

or if you accept my earlier suggestion about renaming the function and its input variable name:

```suggestion
        addresses = [address for address in addresses if address]
```

(plus all the references below)

quipucords/api/source/serializer.py
infinitewarp: I do realize that a lot of legacy code here and outside here uses the word "host", which I'm not terribly fond of, and I also realize there is some subjective judgment here. Part of my thought process is that we should clean up the language while we are in the neighborhood, especially if renaming things might make them more immediately comprehensible for future readers (or future us), because we likely will not touch the code in this area for a long time after this PR completes.

infinitewarp (approved):
Overall, I love this change! The code and logical flow is _much_ simpler and easier to understand than the old version, and the many varied parametrize'd unit tests appear to have excellent coverage. ❤️ 

I am giving this PR my initial approval because it's already so much better than came before it, but I did post a few inline nitpick comments below. I suggested some improved naming/working changes while we are touching this code, but they are admittedly subjective and open for discussion or even rejection if I am the only one who feels this way.

quipucords/api/source/serializer.py
nicolearagao: Agreed! My intention wasn’t to propose a quick fix but to highlight a concern that needs to be properly addressed. Hopefully, soon.

lockfiles/requirements-build.txt
nicolearagao: Definetely!

nicolearagao: @bruno-fs so fast 😮

bruno-fs: >so fast

image

infinitewarp (approved):
This PR just gets better every time I look at it. Have another approval! 😀

PR #2822, opened 2025-02-25
fix: unexpected report count inflation caused by regenerating missing cache files

infinitewarp:
This change significantly simplifies the feature originally added in https://github.com/quipucords/quipucords/pull/2793 that (re)generates missing cached fingerprint files. Instead of running a whole fingerprint-type `ScanTask` as in the original implementation, now we simply look for the related `SystemFingerprint` objects and save their serialized contents. I previously just missed the fact that `SystemFingerprint` objects exist with _effectively_ the full contents of the `cached_fingerprint` field.

Quoting an explanation I [originally posted here in Slack](https://redhat-internal.slack.com/archives/C02QSNF1UKE/p1740426777485639?thread_ts=1740411522.844759&cid=C02QSNF1UKE):

> …the idea for my manual verification so far has been:
>
> * create a new network source, scan it, and download the tarball (1)
> * delete cached files, try to download, see error, try again, download the tarball (2)
> * delete cached files, run the management command, delete cached files again, run the management command twice more, download the tarball (3)
> * compare the three exploded files
>
> Contents of 2 and 3 are identical. Their `SHA256SUM` files match.
>
> Comparing 1 and 2 (or 1 and 3), there are very slight differences, but I think they are acceptable.  Most files are truly identical (`aggregate.json`, `deployments.csv`, `details.csv`, `details.json`). However, the regenerated `deployments.json` is slightly more verbose and includes some fields with `null` values that were absent from the original version of the download. This is because the original file is written using the source data during the original fingerprinting operation that is the input to the `SystemProfile` model. When I regenerate the cached files, the original version of the data is gone, but the transformed version still lives in `SystemProfile` objects. My regenerated files use the `SystemProfileSerializer` output which is nearly identical to its input, but that serializer's output includes some optional fields that may not have been present in the original input (and thus the original "cached" data). Since they are just extra optional fields, I believe the difference is negligible. Just a few extra key-value pairs like `"virtual_host_name": null`.

Relates to JIRA: DISCOVERY-899

### DIFF
--- a/quipucords/api/deployments_report/model.py
+++ b/quipucords/api/deployments_report/model.py
@@ -13,6 +13,7 @@ from django.dispatch import receiver
 
 from api.common.common_report import REPORT_TYPE_CHOICES, REPORT_TYPE_DEPLOYMENT
 from api.common.models import BaseModel
+from api.deployments_report import tasks
 from fingerprinter.constants import (
     ENTITLEMENTS_KEY,
     META_DATA_KEY,
@@ -90,7 +91,7 @@ class DeploymentsReport(BaseModel):
                 self.id,
                 file_path,
             )
-            self._rerun_latest_fingerprint()
+            tasks.generate_cached_fingerprints.delay(self.id)
             raise
 
     @cached_fingerprints.setter
@@ -138,7 +139,7 @@ class DeploymentsReport(BaseModel):
                 self.id,
                 self.cached_csv_file_path,
             )
-            self._rerun_latest_fingerprint()
+            tasks.generate_and_save_cached_csv.delay(self.id)
             raise
 
     @cached_csv.setter
@@ -152,41 +153,6 @@ class DeploymentsReport(BaseModel):
         with file_path.open("w") as f:
             f.write(data)
         self.cached_csv_file_path = file_path
-
-    def _rerun_latest_fingerprint(self, wait=False) -> None:
-        """
-        Trigger the latest fingerprint task to rerun.
-
-        Why would you want to do this? If you expect the cached_fingerprints or
-        cached_csv data to exist but either is missing, you may recreate them by
-        calling this method.
-
-        :param wait: if true, wait for task to complete
-        """
-        logger.info("Rerunning the last fingerprint task for %s", self)
-
-        # local imports to avoid possible import loop
-        from api.scantask.model import ScanTask
-        from scanner import tasks
-
-        scan_task = (
-            ScanTask.objects.filter(
-                job__report__deployment_report_id=self.id,
-                scan_type=ScanTask.SCAN_TYPE_FINGERPRINT,
-            )
-            .order_by("-id")
-            .first()
-        )
-        if not scan_task:
-            # It is unclear how this is possible, but just in case, raise an
-            # exception because we cannot proceed here without a ScanTask,
-            # and we may not want to (or be able to) create a new one on demand.
-            raise ScanTask.DoesNotExist(f"No fingerprint ScanTask is related to {self}")
-
-        logger.info("Found %s for %s", scan_task, self)
-        fingerprint_task = tasks.fingerprint.delay(scan_task_id=scan_task.id)
-        if wait:
-            fingerprint_task.get()
 
 
 @receiver(post_delete, sender=DeploymentsReport)
--- a/quipucords/api/deployments_report/tasks.py
+++ b/quipucords/api/deployments_report/tasks.py
@@ -3,17 +3,49 @@
 import logging
 
 import celery
-
-from api.deployments_report.model import DeploymentsReport
-from api.deployments_report.util import create_deployments_csv
-from api.deployments_report.view import build_cached_json_report
+from django.db import transaction
 
 logger = logging.getLogger(__name__)
 
 
 @celery.shared_task()
+@transaction.atomic
+def generate_cached_fingerprints(deployments_report_id: int):
+    """Generate cached fingerprints based on saved SystemFingerprint objects."""
+    from api.deployments_report.model import DeploymentsReport, SystemFingerprint
+
+    try:
+        from api.deployments_report.serializer import SystemFingerprintSerializer
+
+        deployments_report = DeploymentsReport.objects.get(id=deployments_report_id)
+        deployments_report.cached_csv_file_path = None  # cached CSV may now be invalid
+        deployments_report.cached_fingerprints = SystemFingerprintSerializer(
+            SystemFingerprint.objects.filter(deployment_report=deployments_report),
+            many=True,
+        ).data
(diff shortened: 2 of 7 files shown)


quipucords/api/deployments_report/tasks.py
```diff
@@ -3,17 +3,49 @@
...
+    try:
+        from api.deployments_report.serializer import SystemFingerprintSerializer
+
+        deployments_report = DeploymentsReport.objects.get(id=deployments_report_id)
+        deployments_report.cached_csv_file_path = None  # cached CSV may now be invalid
+        deployments_report.cached_fingerprints = SystemFingerprintSerializer(
+            SystemFingerprint.objects.filter(deployment_report=deployments_report),
+            many=True,
+        ).data
+        deployments_report.save()
```
bruno-fs, line 26: Huh, my memory played me here. I thought we were using a json file stored on the filesystem.

infinitewarp: From current version in `main`:

https://github.com/quipucords/quipucords/blob/main/quipucords/fingerprinter/runner.py#L248
https://github.com/quipucords/quipucords/blob/main/quipucords/fingerprinter/runner.py#L290

Note that we've historically just assigned a `list` of `dict`s directly to `cached_fingerprints`, and we have not converted or sanitized it in some way for future JSON output. I think we've just been lucky that it's already JSON-safe with no troublesome `datetime` or other objects. 🤷

infinitewarp: We do write it as JSON when it finally goes to the filesystem, but not anywhere before then:

https://github.com/quipucords/quipucords/blob/main/quipucords/api/deployments_report/model.py#L99-L105

quipucords/api/deployments_report/model.py
```diff
@@ -153,41 +154,6 @@ def cached_csv(self, data):
...
-                scan_type=ScanTask.SCAN_TYPE_FINGERPRINT,
-            )
-            .order_by("-id")
-            .first()
-        )
-        if not scan_task:
-            # It is unclear how this is possible, but just in case, raise an
-            # exception because we cannot proceed here without a ScanTask,
-            # and we may not want to (or be able to) create a new one on demand.
-            raise ScanTask.DoesNotExist(f"No fingerprint ScanTask is related to {self}")
```
mirekdlugosz, line 184: Double checking: with this new version, fingerprints that are not related to any scan (usually because scan was removed after it was run) are not causing any troubles, right?

I've re-run this PR against database dump that failed in previous version, and was a bit surprised that now all DeploymentsReports are reporting success.

This will also simplify things on generating fresh database dumps front using Camayoc, although I am likely to forget these details before we get to implementing it.

quipucords/api/management/commands/regenerate_missing_report_cache_files.py
```diff
@@ -71,16 +74,10 @@ def generate_fingerprints(
...
-        for _id in deployments_report_ids:
-            deployments_report = DeploymentsReport.objects.get(pk=_id)
-            self.write_info(f"Rerunning fingerprint task for {deployments_report}")
-            try:
-                deployments_report._rerun_latest_fingerprint(wait=True)
-                success_ids.append(_id)
-            except Exception as e:  # noqa: BLE001
-                self.stdout.write(self.style.ERROR(str(e)))
-                failure_ids.append(_id)
+        for _id in sorted(list(deployments_report_ids)):
```
mirekdlugosz, line 77:
```suggestion
        for _id in sorted(deployments_report_ids):
```

Calling `sorted()` on a set already returns a list, no need for `list()` call.
(But not like this call is expensive)

Same applies to line 88.

quipucords/api/management/commands/regenerate_missing_report_cache_files.py
```diff
@@ -71,16 +74,10 @@ def generate_fingerprints(
...
+            success_ids.append(_id) if generate_cached_fingerprints(
+                _id
+            ) else failure_ids.append(_id)
```
mirekdlugosz, line 80: This formatting turned out ugly. Can we follow the same pattern as in `generate_csvs`?
```suggestion
            success = generate_cached_fingerprints(_id)
            success_ids.append(_id) if success else failure_ids.append(_id)
```

quipucords/api/management/commands/regenerate_missing_report_cache_files.py
infinitewarp: Good catch! Simplified in latest push.

quipucords/api/management/commands/regenerate_missing_report_cache_files.py
infinitewarp: Agreed! Make consistent in latest push.

quipucords/api/deployments_report/model.py
infinitewarp: I'm pretty sure this is fine, but I'll run through a couple more manual tests locally to include before and after rerunning and deleting some scans.

infinitewarp: I created three scans. Let's call them Scan A, Scan B, and Scan C.

* Scan A: scan only once, download report, delete cache, regenerate files, download report, compare
* Scan B: scan once then rescan, download report, delete cache, regenerate files, download report, compare
* Scan C: scan once then rescan, download report, delete scan, delete cache, regenerate files, can't download to compare

Scan A and Scan B first and last report downloads match as expected. `SHA256SUM` confirms all files match except `deployments.json`, and the only differences in `deployments.json` are the extra optional `null` values as mentioned in the description for https://github.com/quipucords/quipucords/pull/2793 that introduced this change.

Scan C is a different story. As far as I can tell, we don't offer the ability in the UI or CLI to delete individual jobs or reports, only scans. So, deleting Scan C nuked the Scan and ScanJob rows from the DB, but the related DeploymentsReport rows and their SystemFingerprint rows remain in an orphaned state. Deleting their related cached files and running the management command to regenerate them seems to work fine despite the missing Scans and ScanJobs. However, you cannot download the reports from the UI, and when you *try* to download from the CLI, it fails with a 500 error because `Report has no scanjob.` which I think is preexisting behavior/issue/design flaw. So, there's no way to verify directly through the service that the files regenerated correctly for a deleted scan are correct. :shrug:

However, I think the reports *would* match if you could generate them because the underlying cache files appear to have matching contents before and after regeneration. (I actually just moved the directory without deleting the files so I could compare or restore later.)

mirekdlugosz: I was thinking mostly about failures in regenerating files. Thanks for thorough testing and explanation :) .

mirekdlugosz (approved):

ruda (approved):

PR #2832, opened 2025-03-05
ci(konflux): dynamically set version labels

bruno-fs:
Dynamically create "version" and "version_minor" labels based on pyproject.toml
version. These labels are expected on downstream `ReleasePlanAdmission`.

Relates to JIRA: DISCOVERY-708


abellotti:
👍 nice definition and use of shared task: .tekton/task/generate-version-labels.yaml

infinitewarp (approved):

PR #2845, opened 2025-03-11
refactor: remove connect jobs and tasks from satellite scanner

infinitewarp:
This commit introduces many TODO comments because it is one step in a much larger refactor operation to completely remove the connect-type scan jobs and scan tasks. Some code cannot be removed until we complete the work for scanners of all types.


quipucords/api/scanjob/model.py
```diff
@@ -340,7 +362,31 @@ def _create_inspection_tasks(self, conn_tasks):
...
+                inspect_task = ScanTask.objects.create(
+                    job=self,
+                    source=source,
+                    scan_type=ScanTask.SCAN_TYPE_INSPECT,
+                    status=ScanTask.PENDING,
+                    status_message=_(messages.ST_STATUS_MSG_PENDING),
+                    sequence_number=index,
+                    connection_result=TaskConnectionResult.objects.create(
+                        job_connection_result=self.connection_results
+                    ),
```
bruno-fs, line 383: where are these results coming from? at this point do we expect these results to be empty?

EDIT: reading the rest of the code, I think this starts empty and is populated later when `api.host_count() ` and/or `api.hosts()` are invoked.

quipucords/scanner/satellite/inspect.py
```diff
@@ -21,16 +20,8 @@ class InspectTaskRunner(SatelliteTaskRunner):
...
-            error_message = (
-                f"Prerequisites scan task {conn_task.sequence_number} failed."
-            )
-            return error_message, ScanTask.FAILED
-        return super().execute_task()
-
     def handle_api_calls(self, api):
         """Handle api calls for inspection phase."""
+        api.host_count()
+        api.hosts()
```
bruno-fs, line 26: perfect. cleaner then I anticipated! :clap:

bruno-fs (approved):

ruda (approved):

quipucords/scanner/satellite/six.py
```diff
@@ -482,9 +482,10 @@ def _requests_hosts_unique(self):
 
     def hosts_facts(self):
         """Obtain the managed hosts detail raw facts."""
-        systems_count = len(self.connect_scan_task.connection_result.systems.all())
+        systems_count = len(self.inspect_scan_task.connection_result.systems.all())
```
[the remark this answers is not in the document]
infinitewarp: Yes, this preserves any quirks from the old implementation that used connect-type scan tasks, as I was trying to minimize any other possible side effects from this refactoring. In practice, there _should always_  be an instance assigned by the time execution reaches this point, and if there's not, we probably _do_ want it to blow up with an unhandled exception because that is a sign of something having gone wrong.

PR #2846, opened 2025-03-12
Update project to use Poetry 2.1.1

ruda:
Poetry 2.1.1 _seems_ to work fine now for us,
upgrade to this version and update the `pyproject.toml` file.

On Makefile, remove reference to `poetry lock --no-update` because it is now the default behavior. See: https://python-poetry.org/blog/announcing-poetry-2.0.0/#poetry-lock-defaults-to---no-update

Updated the project requirements with `make update-requirements`.


bruno-fs (changes requested):
please don't update any dependency while we are not safe from "new rust dependencies". See https://url.corp.redhat.com/18ba23c

PR #2847, opened 2025-03-12
build: DRY rpm dependencies

bruno-fs:
The necessity of locking RPM dependencies forced us to repeat them both on `Containerfile` and `rpms.in.yaml`.

This PR changes that.

- DRY: parse rpm dependencies from the now obligatory "rpms.in.yaml"
      file; since we are now forced to track our dependencies on
    rpms.in.yaml, no need to actually repeat it on the Containerfile
- Move `rpms.in.yaml` from `lockfiles` folder to make it clear it is NOT
    an automatically managed file.
- Create a dedicated builder stage on our Containerfile. This makes
      cleaning up build dependencies easier. Right now this might seem
    overkill, but in a subsequent PR rust transitive dependencies will
    be included and this will help on cleanup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,24 +114,16 @@ make server-init QUIPUCORDS_SERVER_PASSWORD="SuperAdmin1"
 Both of the above commands create a superuser with name `admin` and password of `SuperAdmin1`.
 
 ## Running the Server
-Currently, quipucords needs quipucords-ui to run. In order to get it's latest version, run
-
 ```
-make fetch-ui
 make server-static
-```
-
-If you prefer to build it from source, then `make build-ui` rule will be used instead.
-See [quipucords-ui installation instructions](https://github.com/quipucords/quipucords-ui) for further information.
-
-To run the development server, run the following command:
-```
 make serve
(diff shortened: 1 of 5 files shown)


infinitewarp (approved):
Pulled and built locally using raw `podman build` command. Took 6 min 57 sec. To compare, local build from the latest `main` took 5 min 47 sec.

The `lock-rpms` make target took 1 min 15 sec, and `update-lockfiles` took 3 min 53 sec.

Not great, not terrible. ☢️ 🤷 

I do not _love_ the extra complexity and time, but I understand the need in order to satisfy recently added external requirements. I approve this change. `:rubber-stamp:`

README.md
```diff
@@ -106,7 +121,7 @@ make fetch-ui
 make server-static
 ```
 
-If you prefer to build it from source, then `make build-ui` rule will be used instead. 
+If you prefer to build it from source, then `make build-ui` rule will be used instead.
```
bruno-fs, line 124: whoops, this one doens't exist anymore

PR #2861, opened 2025-03-19
fix: restore connect task for satellite to fix broken multi-type scans

infinitewarp:
Before this change, standalone scans for each type (Satellite, network, etc.) all worked fine in isolation. However, if you created a scan with multiple types of sources, and one of those types included Satellite which recently stopped using connect scan tasks, the overall scan would fail because the task models had an inconsistent API across different source types. That problem was introduced in b4999fde72c35b1d0712b529d2f6b92ac0ff12bf via <https://github.com/quipucords/quipucords/pull/2845>.

This change restores the use of connect scan tasks for Satellite to preserve the old internal interface as still used by other source types. The restored use of the connect scan task is only superficial, though, and all "connect" logic for Satellite has still been moved out of its old connect task runner and into its inspect task runner. After we have performed similar refactoring work to the other source types, we should be able to fully remove the connect type scan task from *all* of the different source type task runners.

### DIFF
--- a/quipucords/api/scanjob/model.py
+++ b/quipucords/api/scanjob/model.py
@@ -20,7 +20,6 @@ from api.scan.model import Scan
 from api.scanjob.queryset import ScanJobQuerySet
 from api.scantask.model import ScanTask
 from api.source.model import Source
-from constants import DataSources
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


ruda (approved):

infinitewarp:
I ran [this standalone CI test job](https://jenkins-csb-smqe-stage.dno.corp.redhat.com/view/Discovery/job/discovery-standalone/260/) that _effectively_ succeeded. That job failed on the `test_rerun_scanjob` tests, but [as noted in this Slack conversation](https://redhat-internal.slack.com/archives/C02QSNF1UKE/p1742408290902059?thread_ts=1742407781.073669&cid=C02QSNF1UKE):

> the test that will probably fail is `camayoc/tests/qpc/cli/test_scanjobs.py::test_rerun_scanjob`. if you see it failing in your standalone, you can ignore it and consider pipeline passing.

Because everything else in the job passed, I believe we can consider that to be a success.

bruno-fs (approved):

PR #2863, opened 2025-03-21
Add per-source HTTP proxy support for OpenShift, vCenter, RHACS, Satellite and Ansible

nicolearagao:
What was done:
- Added proxy_url field to the Source model and exposed it in SourceSerializerV2.
- Implemented proxy support for the sources mentioned above. 
- Included validation to ensure the proxy URL follows the host:port format.
- Added tests for serializer validation and proxy integration in the respective scanners

For consistency with the current UI hostname/port format, it may be worth splitting proxy_url into proxy_host and proxy_port right now/in the future.
When setting proxy configuration for requests, it’s important to avoid confusion by following [requests' proxy documentation](https://requests.readthedocs.io/en/latest/user/advanced/#proxies), which recommends:
```
proxies = {
    "http": "http://10.10.1.10:3128",
    "https": "http://10.10.1.10:1080",
}
```
Note that both http and https values use the http:// schema.


quipucords/api/source/model.py
```diff
@@ -53,6 +53,8 @@ class Source(BaseModel):
     most_recent_connect_scan = models.ForeignKey(
         "api.ScanJob", null=True, on_delete=models.SET_NULL, related_name="+"
     )
+    http_proxy = models.CharField(max_length=255, null=True, blank=True)
+    https_proxy = models.CharField(max_length=255, null=True, blank=True)
```
bruno-fs, line 57: `:interestingly-intrigued:` I couldn't help but notice this. Do we need this distinction?

AFAIK, we don't support singular sources with multiple protocols in use. If the host is using https, then proxy HAS to be https. if its HTTP, then http.

quipucords/scanner/openshift/api.py
```diff
@@ -106,18 +106,35 @@ def _dynamic_client(self):
...
+        https_proxy: str = None,
+        **kwargs,
     ):
         """Initialize OpenShiftApi without providing a KubeConfig object."""
         host_uri = f"{protocol}://{host}:{port}"
+        proxy_url = None
+        if http_proxy:
+            proxy_url = "http://" + http_proxy
+        elif https_proxy:
+            proxy_url = "http://" + https_proxy
```
bruno-fs, line 125: aaaaaaaaand this is deeply related to my question here: https://github.com/quipucords/quipucords/pull/2863/files#r2060281078

this gets worse at this point, as a incorrect proxy config can cause a override and the user will scratch their head trying to figure out why their proxy config is not working.

With this in mind, I'd change the proxy related column on `Source` model and simply call it `proxy_url` (taking inspiration on the nomenclature used here). As a validation, I'd check if the protocol is the same used on the source host url.

quipucords/api/source/serializer.py
```diff
@@ -529,10 +530,26 @@ class Meta:
...
+        if not value:
+            return value
+
+        SourceSerializerV2.is_valid_proxy_url_format(value)
+        return value
+
+    @staticmethod
+    def is_valid_proxy_url_format(proxy_url):
+        """Check that the proxy URL is in 'host:port' format."""
+        match = re.match(r"^([\w\.-]+):(\d+)$", proxy_url)
```
infinitewarp, line 549: Would we want to allow users not to specify the port number? Can we assume port 80/443 for plain/secure HTTP, or is that a false assumption?

i.e. Can a user just say "my proxy is https://my-great-proxy.local" without appending ":443"?

(Honest question; I don't know the typical user behavior here.)

quipucords/scanner/ansible/api.py
```diff
@@ -28,10 +36,19 @@ def from_connection_info(  # noqa: PLR0913
...
+        :param proxy_url: proxy URL in the format 'host:port'.
         """
         base_uri = f"{protocol}://{host}:{port}"
         auth = HTTPBasicAuth(username=username, password=password)
-        return cls(base_url=base_uri, verify=ssl_verify, auth=auth)
+        session = cls(base_url=base_uri, verify=ssl_verify, auth=auth)
+
+        proxies = {}
+        if proxy_url:
+            proxies[protocol] = "http://" + proxy_url
```
infinitewarp, line 47: This will never have `https` protocol in the string?

quipucords/api/source/serializer.py
bruno-fs: do we want only host:port to be specified? what about the protocol?

quipucords/scanner/ansible/api.py
bruno-fs, line 47: same as in the serializer comment: is there a reason so support exclusively HTTP protocol? what if we are dealing with a https proxy?

if that's not the intention, I'd also make the serializer expect the protocol.

quipucords/scanner/openshift/api.py
```diff
@@ -106,10 +106,20 @@ def _dynamic_client(self):
...
+        port,
+        ssl_verify: bool = True,
+        proxy_url: str = None,
+        **kwargs,
     ):
         """Initialize OpenShiftApi without providing a KubeConfig object."""
         host_uri = f"{protocol}://{host}:{port}"
+        proxies = {}
+        if proxy_url:
+            proxies[protocol] = "http://" + proxy_url
```
infinitewarp, line 122: Like with Ansible, this will never have `https` protocol in the string? I see this again in OpenShift and won't repeat a third time. Potentially mismatching on `proxies[protocol]` feels wrong to me. Could this instead be:

```suggestion
            proxies[protocol] =  f"{protocol}://{proxy_url}"
```

Or are there cases where `protocol` is neither `'http` nor `https`?

quipucords/scanner/ansible/api.py
infinitewarp: Also, micro optimization (assuming the http-only logic is valid:
 
```suggestion
            proxies[protocol] = f"http://{proxy_url}"
```

bruno-fs:
>For consistency with the current UI hostname/port format, it may be worth splitting proxy_url into proxy_host and proxy_port right now/in the future.

It all depends on how we will support this feature on UI end. Considering there's no explicit ask, go for the simplest approach: a single optional proxy_url field that expects everything. Or maybe consider this story blocked until we hear from a UI/UX expert.

quipucords/scanner/rhacs/api.py
```diff
@@ -36,7 +43,16 @@ def from_connection_info(  # noqa: PLR0913
...
+        :param proxy_url: proxy URL in the format 'host:port'.
         """
         base_uri = f"{protocol}://{host}:{port}"
         auth = HTTPBearerAuth(auth_token=auth_token)
-        return cls(base_url=base_uri, verify=ssl_verify, auth=auth)
+        session = cls(base_url=base_uri, verify=ssl_verify, auth=auth)
+
+        proxies = {}
+        if proxy_url:
+            proxies[protocol] = f"http://{proxy_url}"
```
infinitewarp, line 54: See earlier comments. 🙂 Just adding another comment in case we forget.

quipucords/scanner/openshift/api.py
bruno-fs: forcing the protocol to match host protocol will not always work. one can setup an https proxy using using HTTP (and I think vice versa works as well, but I'm not sure).

I'd suggest going expecting the proxy_url to explicitly include the proper protocol (of the proxy, not of the source - that would be the protocol used to communicate with the source).

quipucords/scanner/ansible/api.py
nicolearagao: see comment above

infinitewarp: Thank you for the explanation referencing the `requests` docs!

quipucords/api/source/serializer.py
infinitewarp: Discussed elsewhere but recording it here for posterity: for the current iteration of this feature, we agreed simply to accept and require exactly `host:port` as the input for this single text field.

bruno-fs (commented):
Looks pretty good! Just asking for small readability improvements.

bruno-fs (approved):

PR #2872, opened 2025-03-26
build: update container image to include ARM64

abellotti:
- Include the build_arm64 run to the main build workflow
- Remove the build-arm64.yml workflow
- Add a task that pushes the multi-arch image that includes both amd64 and arm64 architectures.


abellotti:
Looks like 2:41 for amd64 vs 3:47 for arm64, worth the extra minute for having a multi-arch container image.

abellotti:
image

.github/workflows/build.yml
```diff
@@ -26,31 +26,103 @@ jobs:
...
+      - name: Push Versioned Manifest
+        id: push-versioned-manifest
+        if: contains( env.RELEASE_TAG, '.' )
+        uses: pixelfederation/gh-action-manifest-tool@v0.1.7
+        with:
+          username: ${{ secrets.QUAYIO_USERNAME }}
+          password: ${{ secrets.QUAYIO_PASSWORD }}
+          platforms: linux/amd64,linux/arm64
+          template: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ env.STABLE_TAG }}-ARCH
+          target: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ env.STABLE_TAG }}
```
bruno-fs, line 128: do we need this extra step? asking because we also had it on UI side, but it was later removed

bruno-fs: and if we do need it, wouldn't the push-manifest step need a conditional as well?

abellotti: yeah, this is for the releases, i.e. 1.14.0 image, only triggered with the conditional in line 121. We don't do this yet on the UI side (I should not have pasted this over there).

.github/workflows/build.yml
```diff
@@ -26,31 +26,103 @@ jobs:
...
+
+      - name: Define Image Tag
+        id: define-image-tag
+        shell: bash
+        run: echo "IMAGE_TAG=${{ env.STABLE_TAG == 'main' && 'latest' || contains( env.RELEASE_TAG , '.' ) && env.RELEASE_TAG || env.STABLE_TAG }}" >> "${GITHUB_ENV}"
+
+      - name: Push Manifest
+        id: push-manifest
+        uses: pixelfederation/gh-action-manifest-tool@v0.1.7
+        with:
```
bruno-fs, line 112: Don't we need a conditional here? otherwise when `push-versioned-manifest` is invoked, this will run as well 🤔

.github/workflows/build.yml
bruno-fs: I see? but in this case don't we need a conditional on `push-manifest`?

comment with code context here: https://github.com/quipucords/quipucords/pull/2872/files#r2017417106

.github/workflows/build.yml
abellotti: 🤔 I guess if we only want the x.y.z tag we can do an `if: !contains( env.RELEASE_TAG, '.' )` here

bruno-fs (commented):
@infinitewarp suggested trying to use RedHat's tooling to achieve this ("eat your own dogfood", right? XD) and I agree it might be worth a try.

Looks like buildah action does support multiarch builds, it even has an example [here](https://github.com/redhat-actions/buildah-build/blob/main/.github/workflows/multiarch.yml). What do you think?

.github/workflows/build.yml
abellotti: had to remove that back, this puppy handled x.y 
In latest changes, I've added comments as to which tags are handles in which task.

PR #2878, opened 2025-03-31
build: Update build requirements

abellotti:


lockfiles/requirements-build.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with Python 3.13
+# This file is autogenerated by pip-compile with Python 3.12
```
infinitewarp, line 2: Looks like you need to update your Python. 😉

.github/workflows/test.yml
```diff
@@ -25,6 +25,12 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      - name: Install MacOS dependencies
+        if: matrix.os == 'macos-latest'
```
bruno-fs, line 29: I think this needs to be...

```suggestion
        if: ${{ matrix.os }} == 'macos-latest'
```

bruno-fs (approved):

PR #2879, opened 2025-03-31
refactor: move network connect task runner logic to inspect task runner

infinitewarp:
This includes cleaning up, deleting, rewriting, or moving many existing unit tests.


quipucords/scanner/network/inspect.py
```diff
@@ -65,7 +69,33 @@ def execute_task(self):
...
+
+        This is redundant because we could just scan immediately and handle
+        its failure as needed, but this exists due to legacy design decision
+        that requires an existing list of connection results to be referenced
+        later during the inspection. This could (should) be flattened into a
+        single operation.
+
+        TODO Remove this function when we remove connect scan tasks.
+        """
+        result_store = ConnectResultStore(self.connect_scan_task)
```
bruno-fs, line 91: I hate with so much prejudice `ConnectResultStore`

so convoluted and intertwined with django models. ugh. and this is the reason I still did not write a smoke test for network scans yet - patching this would be a nightmare.

However, I understand it is not the time to refactor this, so I just wanted to point out my hatred.

quipucords/tests/scanner/network/test_utils.py
```diff
@@ -1,11 +1,102 @@
...
+        ]["ansible_ssh_private_key_file"]
+        assert Path(ssh_keyfile).exists()
+    # check context manager cleanup
+    assert not Path(ssh_keyfile).exists()
+
+
+@pytest.mark.django_db
+def test_construct_inventory_from_source_and_credential_models(faker):
+    """Test construct_inventory using data from full Source and Credential objects."""
+    # TODO Simplify this test. Does this really need to use full model objects?
```
bruno-fs, line 45: I think we could get rid of the database requirement here by using the factories to create the objects but not saving it to db (`SomeFactory.build(*args, **kwargs)`)

It won't solve the design issue here (mainly because we don't even need the models, just a dict representations of it)

bruno-fs (approved):

quipucords/scanner/network/inspect.py
infinitewarp: I completely 💯💯💯% agree. `:sadnod:`

quipucords/tests/scanner/network/conftest.py
```diff
@@ -0,0 +1,49 @@
...
+    """Return a Credential with an ssh_key."""
+    return Credential.objects.create(
+        name="network-credential-name",
+        username="network-credential-username",
+        become_method=faker.random_element(Credential.BECOME_METHOD_CHOICES)[0],
+        become_password=faker.password(),
+        become_user=faker.slug(),
+        cred_type=DataSources.NETWORK,
+        password=faker.password(),
+        ssh_key=openssh_key,
```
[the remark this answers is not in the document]
infinitewarp: This is new, and I vaguely remember having a specific reason to create the Credential "manually" instead of using the factory, but that reason currently escapes my memory. I'll take a closer look.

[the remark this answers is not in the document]
infinitewarp: Well, whatever my rationale _was_, it seems either I was wrong or surrounding circumstances changed. I switched this to `CredentialFactory` as you suggested, and all the tests still pass. 🤷 So, I am updating this in an additional commit.

quipucords/tests/scanner/network/test_utils.py
infinitewarp: Simply swapping in `.build` calls does not completely solve the problem here. So, I will look away and let refactoring this be a problem for _future us_ because this is _mostly_ just the result of transplanting (and slightly updating) an old test. 🙈

quipucords/tests/scanner/network/conftest.py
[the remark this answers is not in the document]
infinitewarp: Changed in 777f69f6d26349e28bc4d25c8850097d5aba4464

PR #2883, opened 2025-04-02
PoC: Opt out from automatic fingerprinting on scans + v2 reports

bruno-fs:
Add the feature flag `QUIPUCORDS_AUTOMATIC_REPORT_GENERATION`,  which is enabled by default. When turned off, it disables the fingerprint task in "inspection type scan jobs".

The new endpoint `/api/v2/reports` was added to support this workflow:
- the POST verb accepts a list of scanjob ids or a list of report ids, which shall be used to create a "fingerprint type scan job"  that will fingerprint and duplicate the data associated with the original scanjobs / reports.
- the GET verb (and the detail `/api/v2/reports/<report-id>`) have extra information from the scanjob that originated the report, so this info can be easily used on upcoming changes to UI's "report tab". Example response below:

```json
{
  "count": 3,
  "next": null,
  "previous": null,
  "results": [
    {
      "id": 9,
      "report_platform_id": "f4eaee4c-8988-48ff-a9cd-ac01dd8d8496",
      "status": "running",
      "start_time": "2025-04-02T11:10:28.600008Z",
      "end_time": null,
      "sources": [
        {
          "id": 1,
          "name": "vagrant",
          "source_type": "network"
        }
      ]
    },
    {
      "id": 8,
      "report_platform_id": "83c6eae0-6039-4f70-aa3f-7cba06087e6a",
      "status": "running",
      "start_time": "2025-04-02T11:09:42.959889Z",
      "end_time": null,
      "sources": [
        {
          "id": 1,
          "name": "vagrant",
          "source_type": "network"
        }
      ]
    },
    {
      "id": 4,
      "report_platform_id": "fddf173d-47ad-4c0d-a924-38d95c2cd60f",
      "status": "completed",
      "start_time": "2025-04-02T03:08:57.156170Z",
      "end_time": "2025-04-02T03:09:40.837940Z",
      "sources": [
        {
          "id": 1,
          "name": "vagrant",
          "source_type": "network"
        }
      ]
    }
  ]
}
```

This workflow was manually tested with `QUIPUCORDS_AUTOMATIC_REPORT_GENERATION` set to `True` and `False`.

The workflow with uploaded reports was not tested.

### DIFF
--- a/quipucords/api/reports/serializer.py
+++ b/quipucords/api/reports/serializer.py
@@ -1,71 +1,54 @@
-"""Serializers for reports."""
-
-import re
-
-from django.db import transaction
 from rest_framework.serializers import (
     CharField,
-    ChoiceField,
-    DictField,
+    DateTimeField,
     ListField,
+    ModelSerializer,
+    PrimaryKeyRelatedField,
     Serializer,
-    UUIDField,
     ValidationError,
 )
 
-from api.models import (
-    Report,
-    ScanJob,
-    ScanTask,
+from api.models import ScanJob, ScanTask
+from api.reports.model import Report
+from api.source.serializer import InternalSourceSerializer
+
+SCANJOB_QUERYSET = ScanJob.objects.filter(
+    # scan_type=ScanTask.SCAN_TYPE_INSPECT,
+    status=ScanTask.COMPLETED,
 )
-from constants import MINIMUM_REPORT_VERSION, DataSources
 
 
-class SourceSerializer(Serializer):
-    """Serializer for 'details report' sources."""
+class CreateReportSerializer(Serializer):
+    from_jobs = ListField(
+        child=PrimaryKeyRelatedField(queryset=SCANJOB_QUERYSET),
+        required=False,
+    )
+    from_reports = ListField(
+        child=PrimaryKeyRelatedField(
+            queryset=Report.objects.filter(scanjob__in=SCANJOB_QUERYSET)
+        ),
+        required=False,
+    )
 
-    server_id = UUIDField()
-    report_version = CharField()
-    source_name = CharField()
-    source_type = ChoiceField(choices=DataSources.choices)
-    facts = ListField(child=DictField(), min_length=1)
-
-    def validate_report_version(self, value):
-        """Validate report version."""
-        regex_match = re.match(r"(\d+)\.(\d+)\.(\d+)\+\S", value)
-        if not regex_match:
-            raise ValidationError("Invalid report_version.")
-        source_version = tuple(map(int, regex_match.groups()))
-        if source_version < MINIMUM_REPORT_VERSION:
-            source_version_str = ".".join(map(str, source_version))
-            minimum_version_str = ".".join(map(str, MINIMUM_REPORT_VERSION))
-            raise ValidationError(
-                f"Source version ({source_version_str}) is below the minimum "
-                f"supported report version ({minimum_version_str})"
-            )
-        return value
+    def validate(self, attrs):
+        if not attrs:
+            raise ValidationError("Either from_jobs or from_reports is required.")
+        return super().validate(attrs)
 
 
-class ReportUploadSerializer(Serializer):
-    """Serializer for uploaded details reports."""
+class ReportOverviewSerializer(ModelSerializer):
+    status = CharField(source="scanjob.status", read_only=True)
+    start_time = DateTimeField(source="scanjob.start_time", read_only=True)
+    end_time = DateTimeField(source="scanjob.end_time", read_only=True)
+    sources = InternalSourceSerializer(source="scanjob.sources", many=True)
 
-    report_type = ChoiceField(choices=["details"])
-    sources = SourceSerializer(many=True)
-
-    def validate_sources(self, value):
-        """Validate sources."""
-        if len(value) == 0:
-            raise ValidationError("minimum length required for 'sources' is 1.")
-        return value
-
-    @transaction.atomic
-    def create(self, validated_data):
-        """Create a report filled with raw data and a fingerprint-type ScanJob."""
-        report = Report.objects.create()
-        # create a scan job and required scan tasks for a "upload" job
-        scan_job = ScanJob.objects.create(
-            scan_type=ScanTask.SCAN_TYPE_FINGERPRINT,
-            report=report,
-        )
-        scan_job.ingest_sources(validated_data["sources"])
-        return report
+    class Meta:
+        model = Report
+        fields = [
+            "id",
+            "report_platform_id",
+            "status",
+            "start_time",
+            "end_time",
+            "sources",
+        ]
--- a/quipucords/api/scanjob/model.py
+++ b/quipucords/api/scanjob/model.py
@@ -3,6 +3,8 @@
 These models are used in the REST definitions.
 """
 
+from __future__ import annotations
+
 import logging
 from datetime import UTC, datetime
 
@@ -368,7 +370,9 @@ class ScanJob(BaseModel):
         :param conn_tasks: list of connection tasks
         :param inspect_tasks: list of inspection tasks
         """
-        if self.scan_type == ScanTask.SCAN_TYPE_FINGERPRINT or inspect_tasks:
+        if self.scan_type == ScanTask.SCAN_TYPE_FINGERPRINT or (
+            inspect_tasks and settings.AUTOMATIC_REPORT_GENERATION
+        ):
             prerequisites = conn_tasks + inspect_tasks
             count = len(prerequisites) + 1
             # Create a single fingerprint task with dependencies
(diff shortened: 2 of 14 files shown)


quipucords/quipucords/settings.py
```diff
@@ -430,14 +430,14 @@ def app_secret_key_and_path():
...
+# def warn_on_deprecation(message, category, filename, lineno, file=None, line=None):  # noqa: PLR0913
+#     """Redirect deprecation warnings to our logger."""
+#     logger = logging.getLogger("quipucords.warnings")
+#     logger.warning(f"{message} (from {filename}:{lineno})")
 
 
-warnings.showwarning = warn_on_deprecation
-warnings.simplefilter("always", DeprecationWarning)
+# warnings.showwarning = warn_on_deprecation
+# warnings.simplefilter("always", DeprecationWarning)
```
bruno-fs, line 440: I commented these out only because **somehow** this was causing a recursion error when importing quipucords in a jupyter notebook

infinitewarp (changes requested):
Reorganizing the code for v1 and adding a v2 API with the ability to request a report to be generated all seems reasonable sense to me in general.

My chief concern, though, is that this defines an API called "report" but reading from that API doesn't appear to return Reports; it instead returns a truncated mix of Report and its related ScanJob's fields? It's not terrible, but it certainly feels weird seeing and thinking about this for the first time. As a naive client, I would expect a RESTful report API to return Reports, and some _other_ API (maybe even a child path of /report/) could return this mixed multi-model object. And yes I know our APIs are definitely not _strictly_ RESTful.

I am trying to imagine how I might do this starting from a blank canvas. Following classic RESTful conventions, POST on the top-level nouns are _generally_ responsible for defining the whole content of an object, not defining parameters to request that something else goes and generates the object, or if they do the latter, they still return the actual base noun's object. I have seen and used child paths be used for not-typically-RESTful operations. Maybe something like…

* `POST /report/` creates a Report with the exact given parameters and returns the generated Report. Typical REST, but we might _not_ actually implement or use this.
* `GET /report/` gets a paginated list of Report objects. Typical REST.
* `GET /report/{id}/` gets a particular Report object. Typical REST.
* `POST /report/generate/` could be like your `POST /v2/report/` and accept these non-model "from_X" parameters and return the new `ReportOverviewSerializer` data.
* `GET /report/generate/{id}/` could be like your `POST /v2/report/{id}`.

Or maybe calling this the "`report`" API in general is the wrong approach?

I am not in love with the list I described here either, and "generate" does not feel like the right word, but neither does "overview". I am just trying to think up some alternate API schemas, and that's the first that comes to mind. Maybe it can at least serve to spark other ideas and conversations? 🤷  Or maybe I'm completely off-track here and need you to explain some factors I may have missed. 🙂 

Other questions…

* Do you see this as one step in a larger refactoring of Report-related objects? Thinking about this in context of how different these v1 and v2 report APIs are.
* Do you have any thoughts about the future of the "old" `ReportsGzipRenderer`? Would v2's report API start to offer tarball downloads somewhere, or would that live on the "legacy" v1 API for the foreseeable future?

bruno-fs:
> My chief concern, though, is that this defines an API called "report" but reading from that API doesn't appear to return Reports; it instead returns a truncated mix of Report and its related ScanJob's fields?

I most definitely agree that it is weird. I proposed this when I considered what info should appear in a hypothetical reports list view. Not any of the current reports, aggregate aside, as we will even paginate their json responses. What is being proposed here most definitely works on scanjob API, though. All we would need is to add optional fields for report platform id and a filter for only showing scanjobs with valid reports.

This also makes me think... Should we make the report view about aggregate report? That neatly summarises all content of a report and would fit in a paginated view.

> POST /report/ creates a Report with the exact given parameters and returns the generated Report. Typical REST, but we might not actually implement or use this.

I beg to differ. The report generation is an async process. IMO all we should be giving as an answer is a scanjob id so user can follow its status.

> GET /report/ gets a paginated list of Report objects. Typical REST.

Unless we are talking about Aggregate Report (which, quite honestly I like the idea), I don't see this working. in v2 API the json response for a report will be paginated. How would 2 levels of pagination inside a view like this? Well, we could just show some top level info, then but that's what this PoC is proposing too.

> GET /report/{id}/ gets a particular Report object. Typical REST.

This I agree kinda agree but then comes the question: which report. I think the best candidate for the `/report/<id>` is the normalized report itself. However, technically that would not be a typical REST as the "detail" view would be a paginated one 😅 

> - POST /report/generate/ could be like your POST /v2/report/ and accept these non-model "from_X" parameters and return the new ReportOverviewSerializer data.
> - GET /report/generate/{id}/ could be like your POST /v2/report/{id}.

This works for sure. OTOH, I've been thinking after I said that ScanJob API could easily work for this as well if we add report info there. I guess we can stick to what you are suggesting IF we have plans for moving away from ScanJobs (maybe we should, but probably only in API v3 haha).


> Do you see this as one step in a larger refactoring of Report-related objects? Thinking about this in context of how different these v1 and v2 report APIs are.

Yes. In a broad stroke, my plans for report API would be
- `GET /report/<id>/raw-facts` that returns a paginated json response of raw facts. Besides pagination, the big change when compared to details report is that this view would consider the "root entity" being listed the actual `InspectResult` instead of `Source`. `Source` information would still be present, but nested under `InspectResult` (on details report this is reversed, which makes it very hard to paginate it).
- `GET /report/<id>/` json response is the proper normalised report (or other endpoint we decide). It is a paginated view as well.

>Do you have any thoughts about the future of the "old" `ReportsGzipRenderer`? Would v2's report API start to offer tarball downloads somewhere, or would that live on the "legacy" v1 API for the foreseeable future?

I did not discussed this point with Justin yet, but I assume that yes, we do want it the tarball. I didn't think of the endpoint yet, though. But I do not believe the current ReportsGzipRenderer will be used on v2. Instead, the tarball would be generated during the async report generation and would just be served when requested. We need to discuss its format as well. @mirekdlugosz suggested using the same structure used for insights reports, which is basically a tarball of multilple report pages. I like it.

Also, you did not ask, but I will comment on CSV as well. The proper report (aka normalised report) should support it. As for raw facts I got the blessing to axe it. I need to confirm if this is still the case.

quipucords/api/reports/serializer.py
```diff
@@ -1,71 +1,54 @@
...
+    )
+
+    def validate(self, attrs):
+        if not attrs:
+            raise ValidationError("Either from_jobs or from_reports is required.")
+        return super().validate(attrs)
+
+
+class ReportOverviewSerializer(ModelSerializer):
+    status = CharField(source="scanjob.status", read_only=True)
```
mirekdlugosz, line 40: How will that work when report does not have scanjob associated? That's most common when I run a scan and then delete it, but you could also import report using CLI or merge reports from 2 scans (I'm not super confident about these last two cases).

Does fingerprinting create a new scanjob? I thought it does not. So I think we can't put `scanjob.status` value into `status` field here, because report status should be in {fingerprint_not_run, fingerprint_running, fingerprint_complete, ...errors}.

quipucords/api/reports/serializer.py
```diff
@@ -1,71 +1,54 @@
...
+    start_time = DateTimeField(source="scanjob.start_time", read_only=True)
+    end_time = DateTimeField(source="scanjob.end_time", read_only=True)
+    sources = InternalSourceSerializer(source="scanjob.sources", many=True)
```
mirekdlugosz, line 43: What if I generate report and then modify the source? You can't change source type, but you can change name, hosts, credentials...

I imagine we already have that problem to some extent.

quipucords/api/reports/serializer.py
bruno-fs: It won't, but reports always have a scanjob. Even uploaded or merged reports have a scanjob that contains a single ScanTask with type set to fingerprint. What they lack is a Scan instance.

I've been thinking about what Brad said and maybe we should simply use the scanjob API the part that concern reports API in this PoC. I don't see us moving away from having scanjobs in the near future...

quipucords/api/reports/serializer.py
bruno-fs: Yup, that's a problem, and it will be worse for sourceless Reports (like the ones uploaded). This is part of why I'm now moving [`InspectGroup`](https://github.com/quipucords/quipucords/blob/3851f522b5c2be504ad153ba9d953aacc7bd78eb/quipucords/api/inspectresult/model.py#L84-L107) to be easily accessible to reports (#2895). `InspectGroup` has a snapshot of metadata from a source at the moment the scan was performed.

quipucords/api/scanjob/model.py
```diff
@@ -368,7 +370,9 @@ def _create_fingerprint_task(self, conn_tasks, inspect_tasks):
         :param conn_tasks: list of connection tasks
         :param inspect_tasks: list of inspection tasks
         """
-        if self.scan_type == ScanTask.SCAN_TYPE_FINGERPRINT or inspect_tasks:
+        if self.scan_type == ScanTask.SCAN_TYPE_FINGERPRINT or (
+            inspect_tasks and settings.AUTOMATIC_REPORT_GENERATION
+        ):
```
bruno-fs, line 375: This is the main change required to support optional report generation.

quipucords/api/reports/serializer.py
```diff
@@ -1,71 +1,54 @@
...
+        child=PrimaryKeyRelatedField(
+            queryset=Report.objects.filter(scanjob__in=SCANJOB_QUERYSET)
+        ),
+        required=False,
+    )
+
+    def validate(self, attrs):
+        if not attrs:
+            raise ValidationError("Either from_jobs or from_reports is required.")
+        return super().validate(attrs)
```
bruno-fs, line 36: Input for creating new reports

PR #2894, opened 2025-04-08
Persist aggregate reports in the database

abellotti:
- Updated Database schema to include AggregateReports
- Create new AggregateReports upon query
- Recreate if primary record is newer
- Added Unit tests to check DB record creation/deletion
- Update the /api/schema OpenAPI spec for /reports/<report_id>/aggregate/

Solves: https://issues.redhat.com/browse/DISCOVERY-916


abellotti:
Updated drf-spectacular to include the aggregate response structure:

Screenshot 2025-04-07 at 7 23 14 PM

abellotti:
Requests originally taking around:

- For network resources ~14 - 20 msec
- For vcenter resources ~140 -224 msec
- For ocp resources ~17 - 32 msec

Are now down to a couple of milliseconds consistent across the board once the AggregateRecords are stored in the database.

infinitewarp (changes requested):
I have some mixed feelings about this PR.

On one hand, this is a straightforward conversion of the existing `dataclass` to a `BaseModel`, and it does that just fine. No notes. Ten outta ten. 👍 

On the other hand, part of my thinking for _not_ creating a "proper" model originally was driven by uncertainty about the shape of the aggregate report in the near future, expecting future changes that would require messy data migrations or potentially losing past reports with new versions. If we were expecting changes to the shape of the aggregate report, and because the content of the report is inherently _non-relational_/de-normalized/derivative, I would prefer using the "dumb" document storage technique of shoving all fields into a single unstructured JSON field. That would grant us much greater flexibility _if_ the report structure were to change over time, and that would allow users to continue to access old reports as they were originally generated, unaffected by future developments.

That being said, looking back over the last year since I originally implemented the aggregate report, we have had no changes or additions. Also, despite grand plans a year ago, we still aren't in the "disco pop" world with users keeping long-running instances of quipucords/Discovery. So, maybe those concerns are no longer relevant. 🤷  Still, I'd like to sanity-check with the rest of the team here before we commit to this schema update.

If the answer is "yes, the aggregate report is stable for the foreseeable future" or "we don't care about losing/rebuilding old reports if/when there are changes", then I will return and give this PR my stamp of approval.

I am marking this review as "request changes" just so we don't forget to have that conversation.

ruda (approved):
From my top view, LGTM.

quipucords/api/aggregate_report/model.py
```diff
@@ -37,62 +39,98 @@
...
+    )
+    inspect_result_status_unreachable = models.PositiveIntegerField(
+        default=0, blank=True, null=True
+    )
+    instances_hypervisor = models.PositiveIntegerField(default=0, blank=True, null=True)
+    instances_not_redhat = models.PositiveIntegerField(default=0, blank=True, null=True)
+    instances_physical = models.PositiveIntegerField(default=0, blank=True, null=True)
+    instances_unknown = models.PositiveIntegerField(default=0, blank=True, null=True)
+    instances_virtual = models.PositiveIntegerField(default=0, blank=True, null=True)
+    jboss_eap_cores_physical = models.PositiveIntegerField(
```
mirekdlugosz, line 79: I wonder if that should be int, and not float?

That value is derived from `cpu_core_count`, and `cpu_core_count` was [explicitly changed from int to float 7 years ago](https://github.com/quipucords/quipucords/commit/989dccd7677be6309ad1ecf9c59be5da368af2c0). We have a thread about it on Slack (September 19, 2024). Nobody in current team ever saw that field being anything else than integer, but we also can't rule out some random cloud or ancient RHEL patch revision returning float. Personally, I would err on the side of past team knowing something that we don't.

Same comment applies to `jboss_eap_cores_virtual`, `jboss_ws_cores_physical` and `jboss_ws_cores_virtual`.

quipucords/api/aggregate_report/model.py
```diff
@@ -37,62 +39,98 @@
...
+AGGREGATE_REPORT_SKIP_ATTRS: list = ["id", "report"]
 
 
-@dataclass
-class AggregateReport:
+class AggregateReport(BaseModel):
     """Results of the aggregate report."""
 
+    report = models.OneToOneField(
+        "Report", models.CASCADE, related_name="aggregate_report", null=True
```
bruno-fs, line 49: in which scenario would we expect an aggregate report to exist without a report?

(looking at this `null=True`)

quipucords/api/aggregate_report/view.py
```diff
@@ -1,12 +1,67 @@
...
+from rest_framework import serializers
 from rest_framework.decorators import api_view
 from rest_framework.exceptions import NotFound
 from rest_framework.response import Response
 
 from api.aggregate_report.model import get_aggregate_report_by_report_id
 
 
+class ResultsSerializer(serializers.Serializer):
+    """Serializer for the results portion of the Aggregate Report."""
```
bruno-fs, line 13: question for both serializers: why not use model serializer and configure Meta.fields to simply list the fields we are going to use ?

quipucords/api/aggregate_report/model.py
abellotti: good point. 👍

quipucords/api/aggregate_report/view.py
abellotti: I could not use the model serializer as it is a flat representation structure, no results or diagnostics section, that's only generated in the view via the helper methods, thus I took this approach.

abellotti:
Updated drf schema view with the Decimal represenation (xxxx.xxx) for the cpu cores (virt & phys).

Screenshot 2025-04-09 at 6 12 16 PM

quipucords/api/aggregate_report/model.py
```diff
@@ -37,62 +39,98 @@
...
+        default=0, blank=True, null=True
+    )
+    instances_hypervisor = models.PositiveIntegerField(default=0, blank=True, null=True)
+    instances_not_redhat = models.PositiveIntegerField(default=0, blank=True, null=True)
+    instances_physical = models.PositiveIntegerField(default=0, blank=True, null=True)
+    instances_unknown = models.PositiveIntegerField(default=0, blank=True, null=True)
+    instances_virtual = models.PositiveIntegerField(default=0, blank=True, null=True)
+    jboss_eap_cores_physical = models.DecimalField(
+        default=0.0, max_digits=7, decimal_places=3, blank=True, null=True
+    )
```
mirekdlugosz, line 81: This won't work. Causes:
```
  File "/app/quipucords/api/aggregate_report/model.py", line 227, in _aggregate_from_system_fingerprints
    aggregated.jboss_eap_cores_virtual += jboss_eap_cpu_core_count
TypeError: unsupported operand type(s) for +=: 'decimal.Decimal' and 'float'
```

This is caught by Camayoc UI tests, causing one of them to fail. A bit strange that only UI test fails - I would expect this to fail also unit test and some CLI test.

abellotti: Thanks @mirekdlugosz for finding this, strange indeed, both float during Unit tests. switching to Float for now.

mirekdlugosz:
I give my :+1: 

I am keeping my "request changes" to hold merge of this PR until upgrade testing pipeline data is prepared. I can prepare the data, but only once that PR achieves the state we intend to merge. Once other people confirm they are happy with current state of changes, I will prepare data and switch my review to ACK.

infinitewarp (approved):

quipucords/api/aggregate_report/view.py
bruno-fs: Actually you can use it.

You won't be able to use for the "final" serializer that combines both in the non-flat representation structure,
but you would definitely be able to.

I would prefer that approach to avoid unnecessarily repeating types.

Let me write a quick example here. Hold on a minute...

quipucords/api/aggregate_report/model.py
bruno-fs: If there's no particular reason for setting this relationship with `null=True`, could you set it to `False`?
```suggestion
        "Report", models.CASCADE, related_name="aggregate_report"
```
(I think it is false by default)

quipucords/api/aggregate_report/view.py
bruno-fs:
```python
class ResultsSerializer(serializers.ModelSerializer):
    """Serializer for the results portion of the Aggregate Report."""

    class Meta:
        """Serializer configuration."""

        model = AggregateReport
        fields = [
            "ansible_hosts_all",
            "ansible_hosts_in_database",
            "ansible_hosts_in_jobs",
            "instances_hypervisor",
            "instances_not_redhat",
            "instances_physical",
            "instances_unknown",
            "instances_virtual",
            "jboss_eap_cores_physical",
            "jboss_eap_cores_virtual",
            "jboss_eap_instances",
            "jboss_ws_cores_physical",
            "jboss_ws_cores_virtual",
            "jboss_ws_instances",
            "openshift_cores",
            "openshift_operators_by_name",
            "openshift_operators_by_kind",
            "os_by_name_and_version",
            "socket_pairs",
            "system_creation_date_average",
            "vmware_hosts",
            "vmware_vm_to_host_ratio",
            "vmware_vms",
            "openshift_cluster_instances",
            "openshift_node_instances",
        ]


class DiagnosticsSerializer(serializers.ModelSerializer):
    """Serializer for the diagnostics portion of the Aggregate Report."""

    class Meta:
        """Serializer configuration."""

        model = AggregateReport
        fields = [
            "inspect_result_status_failed",
            "inspect_result_status_success",
            "inspect_result_status_unknown",
            "inspect_result_status_unreachable",
            "missing_cpu_core_count",
            "missing_cpu_socket_count",
            "missing_name",
            "missing_pem_files",
            "missing_system_creation_date",
            "missing_system_purpose",
        ]


class AggregateReportSerializer(serializers.ModelSerializer):
    """Aggregate Report Serializer."""

    results = ResultsSerializer(many=False, source="*")
    diagnostics = DiagnosticsSerializer(many=False, source="*")

    class Meta:
        """Serializer configuration."""

        model = AggregateReport
        fields = ["results", "diagnostics"]

```

This allow us to change the view to properly use the serializer and obsoletes some helper function.

```python
@extend_schema(responses=AggregateReportSerializer)
@api_view(["GET"])
def aggregate_report(request, report_id: int) -> Response:
    """Lookup and return a details system report."""
    aggregate_report = get_object_or_404(AggregateReport, report_id=report_id)
    serializer = AggregateReportSerializer(instance=aggregate_report)
    return Response(serializer.data)
```

quipucords/api/aggregate_report/view.py
```diff
@@ -1,12 +1,67 @@
...
+from drf_spectacular.utils import extend_schema
+from rest_framework import serializers
 from rest_framework.decorators import api_view
 from rest_framework.exceptions import NotFound
 from rest_framework.response import Response
 
 from api.aggregate_report.model import get_aggregate_report_by_report_id
 
 
+class ResultsSerializer(serializers.Serializer):
```
bruno-fs, line 12: please move all these serializers to a dedicated `serializer` module to conform with the pattern used across other models.

quipucords/tests/api/aggregate_report/test_aggregate_report_model.py
```diff
@@ -330,9 +330,11 @@ def test_build_aggregate_report_system_fingerprint(
 ):
     """Test build_aggregate_report using generated Report, SystemFingerprints, etc."""
     report, expected_aggregate_report = report_and_expected_aggregate
-    aggregated: AggregateReport = build_aggregate_report(report.id)
+    aggregated = build_aggregate_report(report.id)
     assert aggregated is not None
-    assert aggregated == expected_aggregate_report
+    assert reformat_aggregate_report_to_dict(
+        aggregated
+    ) == reformat_aggregate_report_to_dict(expected_aggregate_report)
```
bruno-fs, line 337: Now that AggregateReport is its own model I don't see value on the current implementation of the fixture `report_and_expected_aggregate`.

The problem here is that `build_aggregate_report` won't produce a new aggregate report instance. It will simply get the one instance created in that fixture. So in this test if feels we are testing if `1 == 1`.

I would replace the AggregateReport instance returned by that fixture with a dictionary containing all the expected data and in the exact format we expect it to be used.

After that, change this assertion to something like this

```suggestion
    assert reformat_aggregate_report_to_dict(
        aggregated
    ) == expected_aggregate_report
```


Also, if you follow my suggestion about the serializer, this would change to 

```suggestion
    assert AggregateReportSerializer(instance=aggregated).data == expected_aggregate_report
```

quipucords/fingerprinter/jboss_eap.py
```diff
@@ -3,7 +3,7 @@
 import bisect
 import logging
 
-from api.models import Product
+from api.deployments_report.model import Product
```
bruno-fs, line 6: outside of `api` package we should be importing all models from `api.models`. Please change it back.

quipucords/fingerprinter/jboss_web_server.py
```diff
@@ -1,6 +1,6 @@
 """Ingests raw facts to determine the status of JBoss Web Server on system."""
 
-from api.models import Product
+from api.deployments_report.model import Product
```
bruno-fs, line 3: outside of api package we should be importing all models from api.models. Please change it back.

bruno-fs (changes requested):
Please review and adapt `quipucords/tests/api/aggregate_report/test_aggregate_report_model.py` accordingly.

Some tests seem unnecessary now but, most importantly, the fixture deserves some love.

quipucords/api/aggregate_report/model.py
```diff
@@ -104,11 +134,7 @@ def reformat_aggregate_report_to_dict(aggregated: AggregateReport) -> dict:
 
 
 def get_aggregate_report_by_report_id(report_id: int) -> dict | None:
```
bruno-fs, line 136: I think this helper made sense when AggregateReport wasn't a model. Now this function should be deprecated and `get_object_or_404` from `from `rest_framework.generics` should be used instead.

quipucords/api/aggregate_report/model.py
```diff
@@ -37,62 +39,90 @@
...
-    openshift_node_instances: int = 0
+    openshift_cluster_instances = models.PositiveIntegerField(
+        default=0, blank=True, null=True
+    )
+    openshift_node_instances = models.PositiveIntegerField(
+        default=0, blank=True, null=True
+    )
 
 
 def reformat_aggregate_report_to_dict(aggregated: AggregateReport) -> dict:
```
bruno-fs, line 120: Shouldn't we use proper DRF serializers? I get having this when we are not using proper Django models, but now this seems completely unnecessary.

They can handle nested fields. We did that for the insights report, for instance. I made a suggestion on another comment.

quipucords/tests/api/aggregate_report/test_aggregate_report_model.py
```diff
@@ -38,9 +88,13 @@ def report_and_expected_aggregate() -> tuple[Report, AggregateReport]:  # noqa:
     Please read the expected_aggregate updates between objects carefully
     to understand how each new object should affect the end result.
     """
-    expected_aggregate = AggregateReport()
+    expected_aggregate = expected_aggregate_template().copy()
```
bruno-fs, line 91: while this works, the "pytest recommended" approach would be "expect_aggregate" to be its own fixture

quipucords/api/aggregate_report/view.py
```diff
@@ -1,71 +1,32 @@
...
+
+
+def get_serialized_aggregate_report(report_id: int) -> AggregateReportSerializer | None:
+    """
+    Get the aggregate report for the report id specified.
+
+    If the aggregate report does not exist for the report specified,
+    it will be created.
+    """
+    aggregate = get_aggregate_report_by_report_id(report_id)
```
bruno-fs, line 29: don't need to change anything if you don't want right now, but for v2 API I definitely would like to see the possibility of a automatic generation at request time to be completely gone UNLESS that has a completely async workflow. But honestly, I would just throw a error at it and have the user manually regenerating the whole report.

With the changes introduced in this PR, aggregate report generation should be part of the "fingerprinting" process and always be present in the report. If for some reason it wasn't generated, it should be considered an error IMO.

quipucords/api/aggregate_report/model.py
```diff
@@ -23,96 +23,103 @@
...
-    """
+def get_aggregate_report_by_report_id(report_id: int) -> AggregateReport | None:
+    """Build and get the aggregate report for the given report ID."""
     try:
         report = Report.objects.get(pk=report_id)
         aggregated = build_aggregate_report(report.id)
-        return reformat_aggregate_report_to_dict(aggregated)
+        return aggregated
     except Report.DoesNotExist:
         return None
```
bruno-fs, line 124: I'd remove this. I don't see a reason to handle `Report.DoesNotExist` - this is forcing us to handle something Django would automatically take care for us.

~~(Optionally I'd also completely replace this particular function with django rest framework `get_object_or_404` as the workflow of automatically generating aggregate report in a request cycle should be removed - but particular change could wait for v2 api)~~

Scratch that. Actually, after rereading the AC's on the JIRA issue for this story, there's one particular criteria that already states this change

> If it does not exist, too bad. Check the logs, redo the scan, etc.

So, please remove this whole function and use `get_object_or_404` (from `rest_framework.generics`) in the view instead.

quipucords/api/aggregate_report/view.py
bruno-fs: Per my other comment, after rereading the JIRA story for this issue I noticed it specifically calls out for simply throwing an error if the report does not exist.

bruno-fs: Per our discussion on Red Hat's slack, we will change this behavior only in v2 api.

bruno-fs (approved):

mirekdlugosz (approved):

PR #2895, opened 2025-04-09
bind InspectGroup to Report model

bruno-fs:
While adding a bit of redundancy, this streamlines the relationship
between a Report and its raw facts.

As a consequence, it also removes the need for a "fake" inspection type
ScanTask for fingerprint type ScanJobs.


quipucords/quipucords/settings.py
```diff
@@ -436,8 +435,8 @@ def warn_on_deprecation(message, category, filename, lineno, file=None, line=Non
     logger.warning(f"{message} (from {filename}:{lineno})")
 
 
-warnings.showwarning = warn_on_deprecation
-warnings.simplefilter("always", DeprecationWarning)
+# warnings.showwarning = warn_on_deprecation
+# warnings.simplefilter("always", DeprecationWarning)
```
infinitewarp, line 439: Don't forget to un-comment these.

bruno-fs: I rebased on top of the changes you made that fix this issue. Thanks!

bruno-fs:
It just ocurred to me that we might need to change the logic for when deleting scans/scanjobs. They might end up completely wiping the raw facts from "scanless" reports

infinitewarp (approved):
LGTM, including the `TODO` comments for future us. 🙂

PR #2931, opened 2025-05-05
build: stop pushing `main` container tags (but continue pushing `latest`)

infinitewarp:
See this conversation for context:
https://redhat-internal.slack.com/archives/C02QSNF1UKE/p1746208791358419

TLDR: I discovered that we are pushing the combined manifest for `latest` but not for `main`, `latest` basically aliases the `main` tag, and nothing we know currently uses `main`. So, drop `main` and keep `latest`.


ruda (approved):

nicolearagao (approved):
LGTM

PR #2957, opened 2025-05-29
Add a test signaling that backend change might confuse UI

mirekdlugosz:
In https://github.com/quipucords/quipucords-ui/pull/630 , UI expects that keys in aggregate report JSON "buckets" are unique. That is, if there ever exists "results.some_key" and "diagnostic.some_key", UI results might be unpredictable.

Add a test that will fail in this case, with explanation in docstring.


infinitewarp (approved):
This feels more like a potential bug in the UI's code, but I'm okay adding this test as an early warning for us.

PR #2967, opened 2025-06-04
build: unpin transitive build dependency rpds-py

infinitewarp:
See also https://github.com/quipucords/quipucords/pull/2955 which originally pinned rpds-py to `0.24.0` in d40676f467bc0b90fb27b010152ac5940dafcf4b because downstream Konflux builds were failing on newer versions of rpds-py.

## Summary by Sourcery

Build:
- Update project dependency configuration to remove the explicit rpds-py requirement from pyproject.toml and rely on transitive resolution instead.


infinitewarp:
https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1749054039351869?thread_ts=1748381294.852009&cid=C04PZ7H0VA8

> Currently this depends on Cargo versions availability in UBI9, there is an [issue](https://github.com/hermetoproject/hermeto/issues/929) in Hermeto which when resolved would allow us to use newer versions

ruda (approved):

PR #2998, opened 2025-07-14
feat(sources): add v2 bulk delete endpoint

nicolearagao:
Implements `bulk_delete` action for SourceViewSet under the v2 API namespace. Reuses shared logic across both v1 and v2.

Relates to JIRA: DISCOVERY-1006

## Summary by Sourcery

Implement a v2 bulk_delete endpoint for sources by refactoring shared deletion logic into the SourceViewSet, deprecating the old function-based view, and updating tests and URL patterns to cover both API versions

New Features:
- Add bulk_delete action to the v2 SourceViewSet for deleting multiple sources

Enhancements:
- Extract shared bulk delete logic into an internal method for reuse across v1 and v2
- Remove standalone function-based bulk delete view and route in favor of the viewset action

Tests:
- Parametrize source bulk delete tests to cover both v1 and v2 endpoints
- Update URL resolution tests with the new v2 source-bulk-delete route


mirekdlugosz (approved):
Looks good and works well :+1: .

The only thing I don't understand is why you decided to extract everything to `_perform_bulk_delete`, with `bulk_delete` only calling it and wrapping return value in `Response`. I mean - I can see why you would do that during development, but at this point, once all is done, it seems that `_perform_bulk_delete` could be merged back to `bulk_delete` and removed. That would keep it consistent with other methods, who don't have private helpers.

But I don't think it's blocking.

nicolearagao:
> Looks good and works well 👍 .
> 
> The only thing I don't understand is why you decided to extract everything to `_perform_bulk_delete`, with `bulk_delete` only calling it and wrapping return value in `Response`. I mean - I can see why you would do that during development, but at this point, once all is done, it seems that `_perform_bulk_delete` could be merged back to `bulk_delete` and removed. That would keep it consistent with other methods, who don't have private helpers.
> 
> But I don't think it's blocking.

Yeah, good point! I originally split it out because I was considering doing different things for V1 and V2, so I wanted to isolate the core logic early. That didn’t end up being necessary in the final version, so I agree it could be merged back for consistency with other actions. I will clean that up in a follow-up .

quipucords/api/source/view.py
```diff
@@ -246,3 +177,53 @@ def destroy(self, request, pk):
 
         except Source.DoesNotExist as exception:
             raise Http404 from exception
+
+    @action(detail=False, methods=["post"], url_path="bulk_delete")
+    def bulk_delete(self, request):
```
bruno-fs, line 182:
```suggestion
    @action(detail=False, methods=["post"])
    def bulk_delete(self, request):
```
Minor: no need to specify url_path if the method name matches the url you want to use.

nicolearagao (commented):
/retest

PR #3007, opened 2025-07-29
first

nicolearagao:


infinitewarp (approved):

bruno-fs (changes requested):
As I said in a slack message, the error can be reproduced by initializing `OpenShiftApi` class without mocks.

Since that reproduces the error, I recommend adding a test that does that to at least prevent a similar regression in the future.

something along this, would suffice
```python
def test_initialize_openshiftapi_with_proxy_config():
    client = OpenShiftApi.with_config_info(
        host="HOST",
        protocol="https",
        port="443",
        auth_token="TOKEN",
        proxy_url="http://proxy.example.com:8080/",
    )
```

Yes, it is weird to write a test without any assertions, but without the mocks the unpatched code (the fix on line 130 of scanner/openshift/api.py) would reproduce the error. I also recommend add a comment explaining the reason for a "test without assertions.

bruno-fs (approved):

PR #3021, opened 2025-08-21
fix: update `HostEntity.provider_type` and `HostEntity.provider_id` to satisfy new HBI requirements

infinitewarp:
`HostEntity.provider_type` should always return "discovery" to satisfy new HBI requirements.

`HostEntity.provider_id` should return `/etc/machine_id` or another unique value to satisfy new HBI requirements

If a system fingerprint includes a value from `/etc/machine_id`, we will use that as `HostEntity.provider_id`. Else, we hash a set of other fingerprint facts with the expectation that they will be sufficiently unique to identify that system.

However, this change is not that simple! HBI also recently revised the list of attributes that HBI uses to determine uniqueness. We reproduced logic locally to define a list of "canonical facts". Before generating payload for HBI, we would remove any hosts that did not have at least one canonical fact. Because this change now *always* results in a not-empty `provider_id` value and `provider_id` is an attribute HBI now uses to determine uniqueness, this change also removes the previous logic around canonical facts and the filtering out of hosts that did not have at least one canonical fact.

## Summary by Sourcery

Implement new HBI requirements by making HostEntity.provider_type always return “discovery”, and HostEntity.provider_id return /etc/machine_id if present or a deterministic hash of key fingerprint facts otherwise, while removing the concept of canonical facts and its host-filtering logic.

Bug Fixes:
- Ensure provider_type is always “discovery” to meet HBI expectations.

Enhancements:
- Generate provider_id from /etc/machine_id or by hashing a fixed set of fingerprint facts.
- Remove outdated canonical facts list and skip-hosts logic since provider_id now guarantees uniqueness.

Tests:
- Add unit tests for provider_type and provider_id behaviors.
- Remove tests for canonical facts-based filtering and update integration test to validate synthetic provider_id facts.

### DIFF
--- a/quipucords/api/common/entities.py
+++ b/quipucords/api/common/entities.py
@@ -23,6 +23,7 @@ from api.common.enumerators import (
 from api.models import DeploymentsReport, Product, SystemFingerprint
 from compat.db import StringAgg
 from constants import DataSources
+from utils.misc import deep_sort
 
 logger = getLogger(__name__)
 
@@ -64,6 +65,14 @@ FINGERPRINT_FACT_NAMES_TO_SYNTHESIZE_PROVIDER_ID = {
     "is_redhat",
     "etc_machine_id",
 }
+
+
+def hash_system_fingerprint(fingerprint: SystemFingerprint, *, fields: set[str]) -> str:
+    """Hash a SystemFingerprint to a single hopefully-unique identifier string."""
+    fingerprints_dict = model_to_dict(fingerprint, fields=fields)
+    fingerprints_dict = deep_sort(fingerprints_dict)
+    fingerprints_str = json.dumps(fingerprints_dict, sort_keys=True)
+    return hashlib.sha512(fingerprints_str.encode()).hexdigest()
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


infinitewarp:
@sourcery-ai review

quipucords/api/common/entities.py
```diff
@@ -21,18 +24,46 @@
...
+different hosts when running two scans over a short period of time.
+"""
+FINGERPRINT_FACT_NAMES_TO_SYNTHESIZE_PROVIDER_ID = {
+    "name",
+    "os_name",
+    "infrastructure_type",
+    "cloud_provider",
     "mac_addresses",
-    "satellite_id",
+    "ip_addresses",
```
abellotti, line 43: I would personally drop ip_addresses from what we use to create the provider_id, that can be updated by DNS, they unplug the machine one place and plug it somewhere else, different provider_id.  I think the key is to use hardware/static type data, i.e. mac_addresses, etc.

infinitewarp: I have no strong opinion on this one. On one hand, you are correct that IP addresses are mutable. On the other hand, if a system changes IP addresses but doesn't have an `/etc/machine_id`, is it even _the same system_ for counting purposes? We have a minor Ship of Theseus problem here. I'll gather feedback in Slack before making changes here.

nicolearagao (approved):

quipucords/api/common/entities.py
abellotti: Due to the time restrictions, this is ok for now. We can refine/fine-tune with a later PR.

abellotti (approved):

PR #3024, opened 2025-09-10
feat: Adding the source's scan task id and status in a source's view

abellotti:
Do not merge, this is a POC. Needs additional testing in UI for different scenarios.

- When querying a source, we are returning in the connection relationship details on the scan job status and related task statuses upon failure. What was missing in that response is the scan task specific to the source, its id and status. While we could report of the scan job's status (which may show failure for a multi-source job), we could not report on the scanning status of the one source at hand.

Relates-to: https://redhat.atlassian.net/browse/DISCOVERY-1105

## Summary by Sourcery

Expose source-specific scan task id and status in the source view response and update tests accordingly

New Features:
- Add source_task_id and source_task_status to the formatted source scan job response

Tests:
- Update source serializer tests to include and assert source_task_id and source_task_status

### DIFF
--- a/quipucords/api/source/view.py
+++ b/quipucords/api/source/view.py
@@ -46,6 +46,8 @@ def format_source(json_source):
         task_for_source = scan_job.tasks.filter(source=source_id).first()
 
         if task_for_source is not None:
+            json_scan_job["source_task_id"] = task_for_source.id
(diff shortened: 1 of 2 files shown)


abellotti:
Closing for now until we do work on DISCOVERY-1105

PR #3033, opened 2025-10-14
fix: no inline comments

abellotti:
- Having inline comments in embedded shell in the Makefile causes make to fail with a syntax error

```
    $ make update-lockfiles
    ...
    /bin/sh: -c: line 1: syntax error: unexpected end of file
    make: *** [lock-baseimages] Error 2
```

## Summary by Sourcery

Bug Fixes:
- Remove commented lines in the lock-baseimages target loop to avoid shell syntax errors


abellotti:
Closing this issue, later version of Make on the Mac (via brew), handles those fine.

PR #3142, opened 2026-03-28
feat: Adding support for HashiCorp Vault server settings API

abellotti:
- Adding support for HashiCorp Vault credentials leveraging the SecureToken model.
- Exposing the Singleton /api/v2/auth/hashicorp-vault/ endpoint
- Leverating the SecureToken model for secure storage
- Adding CRUD operations for the HashiCorp Vault details
- Supporting verbs and (http status codes)
  - POST for create - (201/400/401/409)
  - PATCH for partial update - (200/400/401)
  - PUT for full replacement - (200/400/401)
  - GET for fetching the server details - (200/401/404)
  - DELETE for deleting - (204/401)
- supported payload attributes:
    address, port, ssl_verify, client_key, client_cert, ca_cert
- client_key, client_cert and ca_cert would not be returned in responses.

Resolves JIRA: DISCOVERY-1206

Todo:
- [x] connecting to the HashiCorp Vault for connectivity/authentication validation (using hvac)
- [x] Add OpenAPI examples
- [x] Adding UNIT Tests

## Summary by Sourcery

Add a new singleton HashiCorp Vault server configuration API backed by SecureToken metadata, including validation and connectivity checks.

New Features:
- Introduce a HashiCorp Vault singleton auth endpoint under /api/v2/auth/hashicorp-vault/ supporting CRUD operations for server configuration.
- Add serializers and helpers to store HashiCorp Vault connection details in SecureToken metadata and authenticate against Vault using client certificates.
- Expose OpenAPI schema, examples, and error responses for HashiCorp Vault configuration operations.

Enhancements:
- Extend the SecureToken model to support a hashicorp-vault token type and allow unbounded encrypted metadata size for large CA bundles.
- Add configuration settings for HashiCorp Vault default port and request timeout.

Build:
- Add hvac as a dependency for communicating with HashiCorp Vault.

Documentation:
- Document HashiCorp Vault auth API behavior via OpenAPI annotations and examples in the codebase.

Tests:
- Add comprehensive unit tests for HashiCorp Vault helper functions, authentication logic, and all CRUD paths of the new endpoint, including error cases.


quipucords/tests/api/auth/test_auth_hashicorp_vault.py
```diff
@@ -0,0 +1,713 @@
...
+        vault_err_message = (
+            messages.HASHICORP_VAULT_FAILED_AUTHENTICATION % vault_address
+        )
+        assert vault_err_message in str(exc_info.value)
+        assert f"BaseHTTPError: {error_message}" in str(exc_info.value)
+
+
+@pytest.mark.django_db
+class TestHashiCorpVaultCreate:
+    """Test the CREATE operation for HashiCorp Vault endpoint."""
```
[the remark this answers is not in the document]
abellotti: added 5 more tests for the client_logged_out case.

quipucords/api/auth/view.py
```diff
@@ -169,3 +224,226 @@ def lightspeed_auth_status(request):
...
+                {"detail": _(messages.HASHICORP_VAULT_ALREADY_EXISTS)},
+                status=status.HTTP_409_CONFLICT,
+            )
+
+        serializer = self.serializer_class(data=request.data)
+
+        try:
+            serializer.is_valid(raise_exception=True)
+        except ValidationError as err:
+            return Response({"detail": err.message}, status=status.HTTP_400_BAD_REQUEST)
```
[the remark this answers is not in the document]
abellotti: actually returned as field specific error from the valid_<field> validators.

quipucords/api/auth/auth_hashicorp_vault.py
```diff
@@ -0,0 +1,222 @@
...
+                verify=ca_file.name,
+                timeout=settings.QUIPUCORDS_HASHICORP_VAULT_TIMEOUT,
+            )
+        else:
+            yield hvac.Client(
+                url=vault_url,
+                cert=(cert_file.name, key_file.name),
+                verify=False,
+                timeout=settings.QUIPUCORDS_HASHICORP_VAULT_TIMEOUT,
+            )
```
mirekdlugosz: Since `verify` is the only difference here, we may do something like:

```python
hvac_args = {
    "url": vault_url,
    "cert": (cert_file.name, key_file.name),
    "verify": False,
    "timeout": settings.QUIPUCORDS_HASHICORP_VAULT_TIMEOUT,
}
if ssl_verify:
    # that part where we write to file
    hvac_args["verify"] = ca_file.name

yield hvac.Client(**hvac_args)
```

quipucords/api/auth/view.py
```diff
@@ -169,3 +225,234 @@ def lightspeed_auth_status(request):
...
+        serializer = self.serializer_class(data=request.data)
+
+        try:
+            serializer.is_valid(raise_exception=True)
+        except HashiCorpVaultAuthError as err:
+            return Response(
+                {"detail": _(err.message)}, status=status.HTTP_400_BAD_REQUEST
+            )
+        except ValidationError as err:
+            return Response(err.detail, status=status.HTTP_400_BAD_REQUEST)
```
mirekdlugosz: That seems unnecessary? According to [DRF docs](https://www.django-rest-framework.org/api-guide/serializers/#raising-an-exception-on-invalid-data):

> These exceptions are automatically dealt with by the default exception handler that REST framework provides, and will return HTTP 400 Bad Request responses by default. 


I mostly paid attention because `HashiCorpVaultAuthError` path passes dict with "detail", while this path passes `err.detail`, which is going to be a list. I don't know if these two are equivalent.

mirekdlugosz (commented):
This is large PR and unfortunately I only skimmed over tests. I hope other reviewers can give a bit more attention to this part.

Also, I have not tried actually using this API. And, what's more important, checking if it correctly connects to vault.

@abellotti the way I read last bullet from jira issue, we want API to act as if vault configuration is a singleton, but we are free to do whatever in underlying implementation. Kind of what is happening with Lightspeed report publish. In your opinion, how much work would it be to support multiple vault configurations in the future, assuming we ever need that? Do you think it might be worth to do it right now to save ourselves headaches later?

quipucords/api/auth/view.py
abellotti: 👍 correct, unnecessary. attribute specific validation errors are still raised and expected errors seen by the tests.

quipucords/api/urls.py
```diff
@@ -73,6 +74,20 @@
...
+        "auth/hashicorp-vault/",
+        HashiCorpVaultViewSet.as_view(
+            {
+                "get": "retrieve",
+                "post": "create",
+                "put": "update",
+                "patch": "partial_update",
+                "delete": "destroy",
+            }
+        ),
```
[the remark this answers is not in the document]
abellotti: commit: [c878471d](https://github.com/quipucords/quipucords/pull/3142/changes/c878471d4315b78895e3fde173120d676331d4fd) for this update.

abellotti:
Indeed a large PR (sorry), in a follow-up refactoring PR, I want to split the auth/* file to be type specific, auth/view_lightspeed and auth/view_hashicorp_vault and such.

quipucords/api/auth/auth_hashicorp_vault.py
mirekdlugosz: Yes, one-liner will work too.

But since there's already `if ssl_verify` conditional, personally I would put it there instead of one-liner. Makes it easier to understand "here are the things that change when we verify ssl". With one-liner, that logic is spread in few  places (at least they are in the same function and just few lines apart).

quipucords/api/urls.py
```diff
@@ -74,6 +78,7 @@
 
 v2_urls = [
     *ROUTER_V2.urls,
+    # HashiCorp Vault singleton endpoint - maps all HTTP methods to base endpoint
```
mirekdlugosz: Not blocking: this comment is confusing, especially as it seems to explain the next line, which is about Lightspeed.

I think it can be removed. Or rephrased to explain there are other `/auth/*` endpoints, but they are defined above. Though that doesn't seem necessary, as the same applies to `reports/` endpoints that work on top of endpoints defined in line 51.

Eventually that line could be moved up to where `auth/hashicorp-vault` is registered, but we don't explain that for all the other ViewSets, so this also doesn't seem necessary.

quipucords/api/auth/view.py
```diff
@@ -169,3 +225,234 @@ def lightspeed_auth_status(request):
...
+        },
+    )
+    def retrieve(self, request, *args, **kwargs):
+        """Get the HashiCorp Vault server definition."""
+        hashicorp_vault_token = get_hashicorp_vault_token()
+        if hashicorp_vault_token is None:
+            return Response(
+                {"detail": _(messages.HASHICORP_VAULT_NOT_DEFINED)},
+                status=status.HTTP_404_NOT_FOUND,
+            )
```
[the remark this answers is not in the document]
mirekdlugosz: That specific comment is in context of `list` endpoint. I think I would prefer consistency in how various auth types handle "do I have this at all?" response.

I don't have strong feelings for `PUT` handler, where also return 404. I mean - just do the same thing we do for other `PUT` handlers. Since Lightspeed doesn't have a concept of `PUT`, we don't need to think about consistency here.

quipucords/api/auth/serializer.py
```diff
@@ -1,6 +1,137 @@
...
+        return instance
+
+    def to_representation(self, instance):
+        """Generate HashiCorp Vault server definition based on SecureToken object."""
+        if isinstance(instance, dict):
+            data = super().to_representation(instance)
+        else:
+            data = super().to_representation(instance.metadata)
+        for sensitive_attr in self.Meta.exclude:
+            data.pop(sensitive_attr, None)
```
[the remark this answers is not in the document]
mirekdlugosz: I was thinking in context of response. `has_ca_cert` doesn't make much sense for input data anyway, and I don't think we support it for credentials.

I agree there isn't much of a value. I imagine people will mostly interact through PATCH, so they will say "update these fields and leave everything else as it is". Right now there are no immediate plans to surface any of that in UI.

Keeping this open to let others of @quipucords/maintainers to chime in. I don't have very strong feelings, just looked like a place where we could be internally consistent.

mirekdlugosz (approved):
Accepting with few more comments. Hoping we have at least one more person reviewing before the merge.

quipucords/api/urls.py
abellotti: sorry, left over from before moving to doing register. I will update.

quipucords/api/auth/serializer.py
```diff
@@ -1,6 +1,124 @@
...
+    decode_cert_from_content,
+    get_or_create_hashicorp_vault_token,
+    hashicorp_vault_authenticate,
+)
+
+
+class HashiCorpVaultSerializer(serializers.Serializer):
+    """Define a serializer for the HashiCorp Vault API."""
+
+    address = serializers.CharField(required=True)
```
nicolearagao, line 29: AC explicitly says: "address may be an FQDN, IPv4, or IPv6 address". Should we implement basic validation here so it at least prevents obviously invalid values? 
If we decide as a team to defer it to the future, we should create a story.

quipucords/api/auth/serializer.py
```diff
@@ -1,6 +1,124 @@
...
+    port = serializers.IntegerField(
```
nicolearagao, line 30: I also think it would be interesting for us to have port validation, now or in the future.

quipucords/api/auth/serializer.py
```diff
@@ -1,6 +1,124 @@
...
+        required=False, default=settings.QUIPUCORDS_HASHICORP_VAULT_DEFAULT_PORT
+    )
+    ssl_verify = serializers.BooleanField(required=False, default=True)
+    client_key = serializers.CharField(required=True)
```
nicolearagao: I know we mentioned the files have unknown maximum size, but I am curious to know if we should consider adding a max_length to these serializer fields anyway? Since the client_cert, client_key, and ca_cert are being stored in the encrypted metadata dict, I’m wondering if we should set a "sane" ceiling.

quipucords/api/auth/view.py
```diff
@@ -159,3 +215,232 @@ def lightspeed_auth_status(request):
...
+        if get_hashicorp_vault_token():
+            return Response(
+                {"detail": _(messages.HASHICORP_VAULT_ALREADY_EXISTS)},
+                status=status.HTTP_409_CONFLICT,
+            )
+
+        serializer = self.serializer_class(data=request.data)
+
+        try:
+            serializer.is_valid(raise_exception=True)
```
nicolearagao, line 325: I was looking at how errors are handled in post and put and had a question. The HashiCorpVaultAuthError gets raised inside the serializer's validate(), but we're catching it in the view outside of is_valid(). I tried to follow how that works. I was wondering if it could be an alternative to catch HashiCorpVaultAuthError inside validate() and turn it into a ValidationError there instead. I don't know if I made myself clear.

quipucords/api/auth/view.py
```diff
@@ -159,3 +215,232 @@ def lightspeed_auth_status(request):
...
+    )
+    def list(self, request, *args, **kwargs) -> Response:
+        """Get the HashiCorp Vault server definition."""
+        hashicorp_vault_token = get_hashicorp_vault_token()
+        if hashicorp_vault_token is None:
+            return Response(
+                {"detail": _(messages.HASHICORP_VAULT_NOT_DEFINED)},
+                status=status.HTTP_404_NOT_FOUND,
+            )
+        serializer = self.serializer_class(data=hashicorp_vault_token.metadata)
```
nicolearagao: I might be getting this wrong, but it looks like the list method passes the stored metadata as data= to the serializer and then calls is_valid(). Doesn't that mean every GET request ends up running validate(), which calls hashicorp_vault_authenticate()? So every time someone just wants to read the vault config, it would try to connect to the Vault server?

abellotti: Good catch!!!  I see the HashiCorp auth calls made upon GETs. we'll need to fix this.

abellotti: Yep, had to use the HashiCorpVaultResponseSerializer for GET and other methods for consistency.  Thanks for catching this ❤️

quipucords/api/auth/view.py
abellotti: Yes, we can do that (raise ValidationError when we get a HashiCorpVaultAuthError.  The one thing I wanted  was to be consistent with rest of the server for non field errors.   Field errors (ValidationError) return an array, i.e. `{"client_cert":["invalid ..."]}` while our own errors, non field, we usually return `{"detail":"message ..."}` not a list of messages.

On the fence here, while I can switch (simpler code in our case not needing the try/except in caller), minor updates to the tests, but then inconsistent with other detail errors.

quipucords/api/auth/serializer.py
abellotti: Being an Integer is validated by DJANGO, and actual value is verified when we authenticate with HashiCorpVault. Should be sufficient validation, unless there's something else we should be checking.

quipucords/api/auth/serializer.py
abellotti: 👍 Added sane upper limits and related tests to verify they are checked/honored.

quipucords/api/auth/serializer.py
abellotti: Currently relying on the actual HashiCorp server auth check to verify this for us, but sure, we can add a validate_address to check for FQDN/IPv4/IPv6.

quipucords/api/auth/serializer.py
nicolearagao: agree!

quipucords/api/auth/serializer.py
nicolearagao: Thank you!

quipucords/api/auth/view.py
nicolearagao: Makes sense. I agree, thanks for explaining the reasoning. Keeping the catch in the view means we return {"detail": "message"} like we do everywhere else for non-field errors. If we raised ValidationError inside validate() instead, DRF would format it as {"non_field_errors": ["message"]}, which would be inconsistent. Not worth it just to save a few lines. Let's keep it as is.

quipucords/api/auth/serializer.py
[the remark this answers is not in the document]
nicolearagao: I agree, not worth adding right now. We can always add it later if needed.

nicolearagao (approved):
Thank you!

quipucords/api/auth/view.py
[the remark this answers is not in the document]
abellotti: 👍 while living under /auth, they are different things being managed, so yea, consistency may not always be possible.

PR #3160, opened 2026-04-22
feat(creds): add vault_secret_path and vault_mount_point to Credential

nicolearagao:
## Summary

- Add `vault_secret_path` (required) and `vault_mount_point` (optional) fields to the
  Credential model, allowing OpenShift and AAP credentials to reference secrets stored
  in HashiCorp Vault instead of storing passwords or tokens directly.
- Enforce mutual exclusivity: `vault_secret_path` cannot coexist with `password`,
  `auth_token`, or `username` on the same credential. `vault_mount_point` is only
  valid alongside `vault_secret_path`.
- Return `auth_type: "vault_secret_path"` in the V2 API response when the field is set.
  Both fields are visible in the API response (not secret, not write-only).
- Reject credential save with a 400 error if no global HashiCorp Vault configuration
  exists.

### Decisions

- Named `vault_secret_path` to align with HashiCorp Vault's "Secret Path" terminology.
- `vault_mount_point` defaults to `"secret"` at scan time (not at model level); stored
  as `null` when not provided.
- No synchronous Vault path verification on save — failures are deferred to scan time.
- V1 serializer is unchanged; V1 clients see `auth_type: "unknown"` for vault credentials.
- KV v2 hardcoded for MVP.

## Manual testing

Tested against a local HashiCorp Vault instance (container, TLS, `ssl_verify: false`).

### Vault setup

- Vault v2.0.0 running via Podman with self-signed server cert
- KV v2 engines at `secret/` (default) and `custom-mount/`
- cert auth enabled with a self-signed client cert
- Global Vault config registered in Discovery (`SecureToken` row)

### Test matrix

| # | Scenario | Expected | Result |
|---|----------|----------|--------|
| **D1** | Create openshift vault cred, no Vault config exists | 400 "Vault configuration is required" | PASS |
| **D2** | Create ansible vault cred, no Vault config exists | 400 same message | PASS |
| **A1** | Create openshift vault cred, no mount_point | 201, `auth_type: "vault_secret_path"`, `vault_mount_point: null` | PASS |
| **A2** | Create openshift vault cred + custom mount_point | 201, `vault_mount_point: "custom-mount"` | PASS |
| **A3** | Create ansible vault cred, no mount_point | 201, `auth_type: "vault_secret_path"` | PASS |
| **A4** | Create ansible vault cred + custom mount_point | 201, both vault fields populated | PASS |
| **B1** | Openshift: vault + auth_token | 400 exclusivity error | PASS |
| **B2** | Openshift: vault + username/password | 400 exclusivity error | PASS |
| **B3** | Openshift: vault + token + username | 400 exclusivity error | PASS |
| **B4** | Ansible: vault + username/password | 400 exclusivity error | PASS |
| **C1-C4** | Vault on network, vcenter, satellite, rhacs | 400 (vault fields not accepted) | PASS |
| **E1** | PATCH openshift from token to vault | 200, `auth_type` switches, token cleared | PASS |
| **E2** | PATCH openshift from vault to token | 200, `auth_type` switches, vault fields nulled | PASS |
| **E3** | PATCH ansible from password to vault | 200, `auth_type` switches, user/pass cleared | PASS |
| **E4** | PATCH ansible from vault to password | 200, `auth_type` switches, vault fields nulled | PASS |
| **F1** | Create with blank `vault_secret_path: ""` | 400 | PASS |
| **F2** | Create with `vault_mount_point` only (no secret_path) | 400 | PASS |
| **F3** | PATCH only `vault_mount_point` on vault cred | 200, mount updated, secret_path preserved | PASS |
| **F4** | GET credential list, vault cred | Vault fields visible, not masked | PASS |
| **F5** | V1 API GET of vault credential | `auth_type: "unknown"`, no vault fields | PASS |
| **G1** | Restart server, GET vault credential | All fields persisted | PASS |

Relates to JIRA: DISCOVERY-1208

## Summary by Sourcery

Add support for storing OpenShift and Ansible credential secrets in HashiCorp Vault via new vault fields and serializers, with validation, API behavior updates, and tests.

New Features:
- Introduce vault_secret_path and vault_mount_point fields on Credential to represent secrets stored in HashiCorp Vault.
- Expose vault-based authentication through new V2 serializers and viewset wiring for OpenShift and Ansible credentials, reporting auth_type as vault_secret_path.

Enhancements:
- Enforce mutual exclusivity between vault_secret_path and existing auth methods (auth_token, username/password) with dedicated validation and error messages.
- Require a global HashiCorp Vault configuration before accepting vault-based credentials, with clear API error messaging.
- Allow PATCH operations to switch credentials between token/password and vault-based authentication while cleaning up obsolete fields.
- Extend credential API responses and tests to cover vault fields and updated auth selection logic.

Tests:
- Add comprehensive serializer and view tests for creating, updating, and validating vault-based credentials across supported and unsupported credential types.


ruda (approved):

quipucords/api/credential/serializer.py
```diff
@@ -308,35 +340,98 @@ class Meta:
...
+            raise ValidationError(messages.USER_PASS_OR_VAULT_NOT_BOTH)
+
+        validator_class = (
+            VaultSecretPathSerializerV2 if has_vault else UsernamePasswordSerializerV2
+        )
+        validator = validator_class(self.instance, data=attrs, partial=self.partial)
+        validator.is_valid(raise_exception=True)
+        return validator.validated_data
+
+    def _detect_auth_mode(self, attrs: dict) -> tuple[bool, bool]:
```
mirekdlugosz: Can't say I'm a fan of calling this "auth mode". This seems to largely overlap with "auth type" we use in other places. As a human, it's hard for me when two very similar concepts are called very similar names.

Having said that, I don't have a good suggestion what other name should be used here...

mirekdlugosz (approved):

infinitewarp (approved):

quipucords/api/credential/serializer.py
nicolearagao: Maybe _check_provided_auth ? @mirekdlugosz @infinitewarp

mirekdlugosz: But we are returning a tuple of booleans, while `check` would suggest that you get back enum value (the core idea being that one bool is True and others are False)...

Maybe instead of having a helper that returns 2- or 3-tuple, we should have more granular helpers in `CredentialSerializerV2`? Then `validate` would do:

```
has_user_or_pass = self._has_user_or_pass(attrs)
has_vault = self._has_vault(attrs)
has_auth_token = self._has_auth_token(attrs)
```

Yes, all these `_has_*` would have very similar implementation, differing mostly by attribute name(s). You can probably have a generic helper to avoid duplicating logic, and `_has_*` would call it.

Otherwise, I think `_detect_provided_auth` would be an improvement over current `_detect_auth_mode`.

abellotti (approved):

quipucords/api/credential/serializer.py
nicolearagao: I went with _detect_provided_auth as a rename. I do think the granular _has approach is cleaner, but it's a bigger refactor than just a rename. I'd need to move the detection logic out of _detect_provided_auth into individual static methods on CredentialSerializerV2, and rework the PATCH fallback coordination in validate() for both AuthTokenOrUserPassSerializerV2 and UsernamePasswordOrVaultSerializerV2. Alberto has already tested the CLI against this code, so I'd rather not change the structure at this point. Happy to revisit in a follow-up if we feel strongly about it.

PR #3162, opened 2026-04-23
fix: improve EAP detection robustness

mirekdlugosz:
JBoss EAP 8 was released in early 2024.

- Improve scanner regular expression to match eap followed by one or more digits. This hopefully will work with EAP 9, EAP 10 and other future versions.
- In EAP version.txt processing, make "GA" suffix optional. It is not present in EAP 8. When suffix was mandatory, `eap_home_version_txt` fact contained the whole `version.txt` content. Now that suffix is optional, the fact will contain only version number.

I had three distinct VMs with EAP 8 installed through different installation methods:
- `jboss-eap-installation-manager`
- archive installation
- RPM

I've ran network scan against all three of them.

Before this change, `deployments.json` file showed "present" for archive and RPM, but "potential" for jboss-eap-installation-manager.

After this change, `deployments.json` file shows "present" for all three of them.

Additionally, change `eap_home_version_txt` fact in `details.json` file. Before, it contained full content of `version.txt` file. Now it contains only version number.

Relates to JIRA: DISCOVERY-1277

## Summary by Sourcery

Improve robustness of JBoss EAP version detection across installations, including newer EAP releases.

Bug Fixes:
- Correct JBoss EAP home version parsing so the stored fact captures only the version number even when the GA suffix is absent.
- Broaden RPM-based JBoss EAP package detection to recognize EAP versions beyond 7 using a generic numeric pattern.

### DIFF
--- a/quipucords/scanner/network/runner/roles/jboss_eap/tasks/main.yml
+++ b/quipucords/scanner/network/runner/roles/jboss_eap/tasks/main.yml
@@ -6,7 +6,7 @@
 
 # Tasks that can locate an EAP_HOME directory
 - name: Gather jboss.eap.running-paths
-  raw: export LANG=C LC_ALL=C; for proc_pid in $(find /proc -maxdepth 1 -xdev -name "[0-9]*" 2>/dev/null); do ls -l ${proc_pid}/fd 2>/dev/null | grep "java"; done | grep -e "/modules/system/layers/base\|/opt/rh/eap[0-9]/root/usr/share/wildfly" | sed -n "s/.*\-> //p" | sed -n 's/\/modules\/system\/layers\/base.*//p;s/\(.*wildfly\).*/\1/p' | sort -u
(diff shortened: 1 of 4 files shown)


quipucords/scanner/network/runner/roles/jboss_eap/tasks/main.yml
nicolearagao: hey, I was taking a look at the other commands, should we update this line:
"$(find /proc -maxdepth 1 -xdev -name "[0-9]*" " to eap[0-9]+ so 10 can also be detected ?

nicolearagao (approved):
Code makes sense, but relying on the tests you did on your end.

quipucords/scanner/network/runner/roles/jboss_eap/tasks/main.yml
mirekdlugosz: You are right. I've noticed that later, when working on another thing.

It is now changed. I also added `eap8` and systemd unit files to another task.

I've amended to existing commit instead of fixup, as it's still very small commit.

PR #3171, opened 2026-05-04
feat(vault): fetch credentials from HashiCorp Vault during OCP and AAP scans

nicolearagao:
## Summary

- Add `read_vault_secret` helper that authenticates to HashiCorp Vault via TLS cert
  and reads the `auth_token` from a KV v2 secret using the credential's
  `vault_secret_path` and `vault_mount_point`.
- Wire vault credential support into OpenShift scan tasks - when a credential has a
  `vault_secret_path`, the scanner fetches the bearer token from Vault at scan time
  instead of using stored credentials.
- Wire vault credential support into AAP scan tasks - same approach, plus extend
  `AnsibleControllerApi` to support bearer-token (OAuth2) authentication.

Credentials retrieved from Vault are never cached, logged, or persisted; they are
fetched fresh on every scan execution.

If Vault is unreachable or denies access, the scan task fails with an appropriate
error message via `ScanFailureError` → `ScanTask.FAILED`.

## Test plan

- [x] Unit tests for `read_vault_secret`: happy path, missing key, forbidden,
  invalid path, connection error, vault-not-configured.
- [x] Integration tests for OCP `_get_connection_info` with vault credential and
  failure propagation.
- [x] Integration tests for AAP `_get_connection_info` with vault credential and
  failure propagation.
- [x] Unit tests for `AnsibleControllerApi` bearer-token instantiation.
- [x] Manual test against a live Vault instance with TLS cert auth and KV v2 engine.

Relates to JIRA: DISCOVERY-1228

## Summary by Sourcery

Integrate HashiCorp Vault-backed bearer-token credentials into OpenShift and Ansible scan flows, including a shared Vault secret reader and updated Ansible controller API authentication behavior.

New Features:
- Add helper to fetch OAuth bearer tokens from HashiCorp Vault for credentials using a Vault secret path.
- Enable OpenShift scan tasks to retrieve authentication tokens from HashiCorp Vault at scan time instead of stored credentials.
- Enable Ansible Automation Platform scan tasks to retrieve authentication tokens from HashiCorp Vault and use bearer-token authentication in the controller API client.

Enhancements:
- Improve error messaging and failure handling when Vault is not configured, secrets are malformed, or Vault access fails.

Tests:
- Add unit tests for Vault secret retrieval helper covering success and multiple failure modes.
- Add integration-style tests ensuring OpenShift and Ansible scanners use Vault-backed credentials and propagate Vault errors.
- Add tests for AnsibleControllerApi instantiation with bearer tokens versus basic auth.

### DIFF
--- a/quipucords/api/auth/hashicorp_vault/auth.py
+++ b/quipucords/api/auth/hashicorp_vault/auth.py
@@ -14,6 +14,7 @@ from urllib3.exceptions import HTTPError as BaseHTTPError
 
 import api.messages
 from api.secure_token.model import SecureToken
+from scanner.exceptions import ScanFailureError
 
 logger = getLogger(__name__)
 
@@ -231,3 +232,42 @@ def hashicorp_vault_authenticate(vault_token=None, metadata=None):
             raise HashiCorpVaultAuthError(
                 _(api.messages.HASHICORP_VAULT_HTTP_ERROR % (vault_address, str(err)))
             )
+
+
+def read_vault_secret(credential) -> str:
+    """Fetch a secret value from HashiCorp Vault using credential configuration."""
+    vault_token = get_hashicorp_vault_token()
+    if vault_token is None:
+        raise ScanFailureError(api.messages.VAULT_SECRET_NOT_CONFIGURED)
+
+    path = credential.vault_secret_path
+    mount_point = credential.vault_mount_point
+    secret_key = credential.vault_secret_key
+
+    try:
+        with hashicorp_vault_client(vault_token=vault_token) as client:
+            if not client.is_authenticated():
+                raise ScanFailureError(api.messages.VAULT_SECRET_AUTH_FAILED)
+            read_kwargs = {"path": path}
+            if mount_point:
+                read_kwargs["mount_point"] = mount_point
+            response = client.secrets.kv.v2.read_secret_version(**read_kwargs)
+    except (
+        hvac.exceptions.VaultError,
+        ConnectionError,
+        BaseHTTPError,
+        HashiCorpVaultAuthError,
+    ) as err:
+        raise ScanFailureError(
+            api.messages.VAULT_SECRET_FETCH_FAILED % (path, mount_point, err)
+        ) from err
+
+    secret_data = response.get("data", {}).get("data", {})
+    if not secret_data:
+        raise ScanFailureError(api.messages.VAULT_SECRET_NO_DATA % (path, mount_point))
+    value = secret_data.get(secret_key)
+    if not value:
+        raise ScanFailureError(
+            api.messages.VAULT_SECRET_MISSING_KEY % (secret_key, path)
+        )
+    return value
--- a/quipucords/scanner/openshift/runner.py
+++ b/quipucords/scanner/openshift/runner.py
@@ -2,6 +2,7 @@
 
 from abc import ABCMeta
 
+from api.auth.hashicorp_vault.auth import read_vault_secret
 from api.models import ScanTask
 from api.vault import decrypt_data_as_unicode
 from scanner.openshift.api import OpenShiftApi
@@ -25,8 +26,10 @@ class OpenShiftTaskRunner(ScanTaskRunner, metaclass=ABCMeta):
             "ssl_verify": ssl_verify,
             "proxy_url": proxy_url,
         }
-        if credential.auth_token:
-            conn_info.update(auth_token=decrypt_data_as_unicode(credential.auth_token))
+        if credential.vault_secret_path:
+            conn_info["auth_token"] = read_vault_secret(credential)
+        elif credential.auth_token:
+            conn_info["auth_token"] = decrypt_data_as_unicode(credential.auth_token)
         else:
             conn_info.update(
                 username=credential.username,
--- a/quipucords/tests/scanner/ansible/test_api.py
+++ b/quipucords/tests/scanner/ansible/test_api.py
@@ -196,3 +196,32 @@ def test_ansible_api_instantiation_with_ipv6_host():
     )
 
     assert api.base_url == "https://[fd00:dead:beef::126]:6443"
+
+
+def test_ansible_api_instantiation_with_auth_token():
+    """Assert bearer token auth sets Authorization header without HTTPBasicAuth."""
+    api = AnsibleControllerApi.from_connection_info(
+        host="localhost",
+        protocol="https",
+        port=8443,
+        auth_token="my-bearer-token",
+        ssl_verify=True,
+    )
+
+    assert api.base_url == "https://localhost:8443"
+    assert api.headers["Authorization"] == "Bearer my-bearer-token"
+    assert api.auth is None
+
+
+def test_ansible_api_instantiation_with_user_pass_no_bearer():
+    """Assert username/password uses HTTPBasicAuth and no bearer header."""
+    api = AnsibleControllerApi.from_connection_info(
+        host="localhost",
+        protocol="https",
+        port=8443,
+        username="admin",
+        password="secret",
+    )
+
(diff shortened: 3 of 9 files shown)


quipucords/scanner/openshift/runner.py
```diff
@@ -25,8 +26,10 @@ def _get_connection_info(cls, scan_task: ScanTask):
             "ssl_verify": ssl_verify,
             "proxy_url": proxy_url,
         }
-        if credential.auth_token:
-            conn_info.update(auth_token=decrypt_data_as_unicode(credential.auth_token))
+        if credential.vault_secret_path:
+            conn_info["auth_token"] = read_vault_secret(credential)
+        elif credential.auth_token:
+            conn_info["auth_token"] = decrypt_data_as_unicode(credential.auth_token)
```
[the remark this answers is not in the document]
nicolearagao: The mutual exclusivity is already enforced at the credential serializer level (USER_PASS_OR_VAULT_NOT_BOTH validation). So in practice both fields shouldn't coexist on the same credential.

quipucords/tests/scanner/ansible/test_api.py
```diff
@@ -196,3 +196,32 @@ def test_ansible_api_instantiation_with_ipv6_host():
...
+def test_ansible_api_instantiation_with_auth_token():
+    """Assert bearer token auth sets Authorization header without HTTPBasicAuth."""
+    api = AnsibleControllerApi.from_connection_info(
+        host="localhost",
+        protocol="https",
+        port=8443,
+        auth_token="my-bearer-token",
+        ssl_verify=True,
+    )
+
```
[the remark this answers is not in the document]
nicolearagao: The caller (_get_connection_info) already enforces mutual exclusivity - it passes either auth_token or username/password to from_connection_info, never both. This is guaranteed by the credential model validation (USER_PASS_OR_VAULT_NOT_BOTH).

quipucords/api/auth/hashicorp_vault/auth.py
```diff
@@ -231,3 +232,36 @@ def hashicorp_vault_authenticate(vault_token=None, metadata=None):
...
+    vault_token = get_hashicorp_vault_token()
+    if vault_token is None:
+        raise ScanFailureError(api.messages.VAULT_SECRET_NOT_CONFIGURED)
+
+    path = credential.vault_secret_path
+    mount_point = credential.vault_mount_point
+
+    try:
+        with hashicorp_vault_client(vault_token=vault_token) as client:
+            client.auth.cert.login()
```
abellotti: ☝️ fails where there's no CA Cert and ssl_verify is False. Could be replaced with a client.is_authenticated() condition block.

quipucords/api/auth/hashicorp_vault/auth.py
```diff
@@ -231,3 +232,36 @@ def hashicorp_vault_authenticate(vault_token=None, metadata=None):
...
+        hvac.exceptions.VaultError,
+        ConnectionError,
+        BaseHTTPError,
+        HashiCorpVaultAuthError,
+    ) as err:
+        raise ScanFailureError(
+            api.messages.VAULT_SECRET_FETCH_FAILED % (path, mount_point, err)
+        ) from err
+
+    secret_data = response.get("data", {}).get("data", {})
```
infinitewarp, line 265: Since we have `if not auth_token:` raise an exception, should we similarly have `if not secret_data:` raise an exception after this line? The only reason I suggest this is there *might* be some value in saying "we didn't get any data" versus "we got data but it didn't have the auth token". Maybe I'm overthinking this, though!

quipucords/api/auth/hashicorp_vault/auth.py
```diff
@@ -231,3 +232,36 @@ def hashicorp_vault_authenticate(vault_token=None, metadata=None):
...
+    auth_token = secret_data.get("auth_token")
```
infinitewarp: I'm probably missing some details about the Vault behaviors here; so, please help me connect the gaps...

I read this `secret_data.get("auth_token")` command, and I found a corresponding unit test:

```python
def test_read_vault_secret_happy_path():
    """Test read_vault_secret returns auth_token on success."""
...
    mock_client.secrets.kv.v2.read_secret_version.return_value = {
        "data": {"data": {"auth_token": "tok"}}
    }
```

Where is this `{"auth_token": ...}` dict originally coming from? Is this nested dictionary just the standard response from Vault's KV API? Or does this imply that the user is storing the original value as a `{"auth_token": ...}` dict at the specified path?

I believe our end user will be storing their raw auth token values in Vault as _just a raw string_, like `aGVsbG8gd29ybGQKCg==`. When I look a this `secret_data.get("auth_token")` get, is seems to expect that the user is *not* storing the raw string but is storing it encapsulated in a custom dict. Am I misunderstanding this?

infinitewarp: I'm also curious about the nested `{"data": {"data":` dicts. Are those just a standard wrapper that *all* KV secrets include as they are returned?

abellotti: was curious about that too, that's the KV v2 (i.e. now versioned) engine responses.

we have: `client.secrets.kv.v2.read_secret`(path=, mount_point=) this will give us the latest version
we also have `client.secrets.kv.v2.read_secret_version`(...., version=) to give us a particular version.

here's an example response:

```
{
  "request_id": "f23d06fd-4c2b-0ca2-3d3b-824003d57a26",
  "lease_id": "", 
  "renewable": False,
  "lease_duration": 0,
  "data": {
    "data": {
      "auth_token": "SAMPLE_AAP_USER_TOKEN",
      "something_else": "there you go" 
    },  
    "metadata": {
      "created_time": "2026-05-07T22:12:40.241516713Z",
      "custom_metadata": None,
      "deletion_time": "", 
      "destroyed": False,
      "version": 3
    }   
  },  
  "wrap_info": None,
  "warnings": None,
  "auth": None,
  "mount_type": "kv"
}
```

Here I created another version of the secret and added a `something_else` key.

infinitewarp: New changes are coming to main that affect this code! Closing this conversation as we have discussed more offline.

infinitewarp: After rebasing main following Alberto's latest additions, I think this would become:
```suggestion
    auth_token = response.get("data", {}).get("data", {}).get(credential.vault_secret_key)
```

infinitewarp: Or just 
```suggestion
    auth_token = response.get("data", {}).get("data", {}).get(vault_secret_key)
```
if you add `vault_secret_key = credential.vault_secret_key` next to the other local variables near the top of this function.

quipucords/api/auth/hashicorp_vault/auth.py
nicolearagao: I think it's a good call, Brad. IMO read_vault_secret distinguishing "the secret returned no data at all" (wrong path, deleted secret, etc.) from "the secret has data but is missing the expected key" (credential misconfiguration) brings value. Each case gets its own error message so it's clearer where to look when debugging.

quipucords/api/auth/hashicorp_vault/auth.py
nicolearagao: I am doing something similar but keeping the intermediate secret_data variable so we can distinguish "vault returned no data" from "data exists but the expected key is missing" (per your other comment).

quipucords/api/auth/hashicorp_vault/auth.py
abellotti: Thanks for fixing this.

infinitewarp (approved):

abellotti (approved):

PR #3172, opened 2026-05-05
fix: Improve publish error messages

abellotti:
- Improve a couple of the Lightspeed report publish error messages to include user recommendations (i.e. retry).

For: https://redhat.atlassian.net/browse/DISCOVERY-1215

## Summary by Sourcery

Clarify Lightspeed report publish error messages with specific user guidance for connection and authentication failures.

Enhancements:
- Improve connection error message to suggest retrying later.
- Improve authentication rejection message to instruct users to log in again.


infinitewarp (approved):
Gut feel tells me that the UI should explicitly own the more flowery "please try again" text since it could be styled and made consistent across errors, but since we just float the raw message up from the API response, this is probably okay.

nicolearagao (approved):

PR #3174, opened 2026-05-05
build: Upping ansible package version to 13.6.0

abellotti:
- Updating the ansible package used to version 13.6.0

Todo:
- [x] fix playbooks
- [x] compare reports with quipucords running version 10.7.0 of ansible.
- [x] check problematic patterns in the other playbooks.


infinitewarp:
I hope this goes better than the last attempt https://github.com/quipucords/quipucords/pull/3060 which we abandoned because the upgrade to Ansible v12 broke our scans. 🍀

mirekdlugosz:
FYI, there are tasks that are never executed as part of our pipeline, because we don't have required machines. See https://github.com/quipucords/quipucords/pull/2722 for list of tasks, and instructions how to create it.

abellotti:
Compared result json's from this PR to 10.7.0 upstream PR, only differences were metadata related and not content differences, i.e. things like `start_time`, `duration`, `jenkins_job_number`,.... were the differences.

abellotti:
Looked at the other ansible playbooks for potential issues similar to what was fixed here:
- no other playbooks sets a value to "True"
- one other playbook does a result coercion to string(), and that's the subman_overall_status, but that is a string to begin with so no issues there.
- no other tests accesses the `__class__.__name__` for messages in a playbook.

mirekdlugosz:
While testing this, I noticed one more change:
`system_memory_bytes` used to be a string, now it is an int (`'1857429504'` -> `1857429504`).
Technically that's a regression, but maybe we are fine with such change. Not sure if other teams using Discovery reports would mind.

I have not checked all variables mentioned in https://github.com/quipucords/quipucords/pull/2722

ruda (approved):

mirekdlugosz (approved):

infinitewarp (approved):

PR #3199, opened 2026-05-21
Optimize AAP scans using host_metrics endpoint

infinitewarp:
AAP (Ansible Automation Platform) scans were experiencing severe performance issues, with customer scans taking 37+ hours for controllers with 100k+ job history. The root cause was O(n) sequential API calls to /api/v2/jobs/{id}/job_events/ where n = total jobs.

This change switches to using the /api/v2/host_metrics/ endpoint, which provides the same unique hostname data in a single paginated endpoint instead of thousands of nested calls. This reduces scan time from hours to minutes (99.9% improvement, ~300-13,000x speedup).

Implementation:
- Add QUIPUCORDS_AAP_USE_HOST_METRICS environment variable (default: True) as feature flag to enable/disable new endpoint
- Detect host_metrics endpoint availability at runtime via endpoint discovery
- Fall back gracefully to old /api/v2/jobs/ method if host_metrics unavailable (older AAP versions) or feature flag disabled
- Maintain backward compatibility in aggregate reports by supporting both data formats (jobs.unique_hosts and unique_hosts)
- Preserve comparison data generation for details reports consumed by external users

Memory impact is acceptable: new implementation uses slightly more memory at very large scales (>100k hosts) due to storing hostnames in both set and list, but peak usage remains well under 1 GB even for 1M hosts. The massive time savings far outweigh the modest memory trade-off.

Relates to JIRA: DISCOVERY-1256

Assisted-by: Claude Sonnet 4.5 <noreply@anthropic.com>

## Summary by Sourcery

Optimize Ansible inspect scans to derive unique automated hosts via the host_metrics endpoint while preserving backward compatibility and reporting behavior.

New Features:
- Add support for discovering and using the Ansible Controller /api/v2/host_metrics/ endpoint to obtain unique automated hosts.
- Introduce the QUIPUCORDS_AAP_USE_HOST_METRICS feature flag to toggle use of the host_metrics-based collection path.

Enhancements:
- Refine the inspect workflow to separately collect instance details and inventory hosts, then derive unique_hosts via either host_metrics or jobs, and generate comparison data only on successful inspections.
- Extend endpoint detection and logging to account for optional host_metrics availability and document fallback behavior.
- Adjust Ansible comparison logic to operate on a unified unique_hosts field regardless of whether data came from jobs or host_metrics.
- Update aggregate report generation to handle both legacy jobs.unique_hosts data and the new top-level unique_hosts layout, keeping existing metrics consistent.
- Refresh test data, integration tests, and raw facts generators to align with the new unique_hosts representation and host_metrics behavior.

Tests:
- Add unit tests covering get_unique_hosts_from_metrics, host_metrics vs jobs fallback, feature-flag behavior, comparison generation, and failure scenarios in the inspect flow.
- Add aggregate report tests for scans using only the new unique_hosts format and mixed legacy/new Ansible formats to ensure backward compatibility.

## Summary by Sourcery

Optimize Ansible inspect scans to derive unique automated hosts via the host_metrics endpoint with feature-flagged fallback while preserving reporting and backward compatibility.

New Features:
- Introduce the QUIPUCORDS_AAP_USE_HOST_METRICS feature flag to control use of the Ansible Controller /api/v2/host_metrics/ endpoint for discovering automated hosts.

Enhancements:
- Extend Ansible endpoint detection and logging to account for optional host_metrics support and to fall back to the jobs endpoint when disabled or unavailable.
- Refine the inspect workflow to collect instance details and inventory hosts separately, then obtain unique automated hosts via host_metrics or jobs while generating comparison data only on successful inspections.
- Ensure comparison and aggregate reporting logic handle both legacy jobs-based unique_hosts and host_metrics-derived data, including cases with and without job_ids.
- Minorly adjust the network inspect callback test logging to avoid leaking warnings between tests.

Tests:
- Add unit and integration tests covering host_metrics-based unique host collection, feature-flag behavior, fallback to jobs, comparison generation, and failure scenarios in the inspect flow.
- Add aggregate report tests to validate behavior with host_metrics-only data, mixed jobs/host_metrics formats, and preservation of jobs.job_ids for backward compatibility.


quipucords/api/aggregate_report/model.py
```diff
@@ -306,9 +306,13 @@ def _aggregate_from_raw_facts(
...
                     ]
                 )
-                ansible_hosts_in_jobs.extend(
-                    raw_fact.get("jobs", {}).get("unique_hosts", [])
-                )
+                # unique_hosts can be stored in two locations:
+                # - Top-level field: collected from /api/v2/host_metrics/
+                # - Nested under jobs: collected from /api/v2/jobs/ + job_events
+                unique_hosts = raw_fact.get("unique_hosts")
+                if unique_hosts is None:
```
mirekdlugosz: I wonder if we should use sentinel value here? Just in unlikely case someone goes crazy and starts returning `null` instead of empty list?

If we can ever update to Python 3.15, they have [`sentinel` function built-in](https://docs.python.org/3.15/library/functions.html#sentinel).

quipucords/scanner/ansible/inspect.py
```diff
@@ -262,11 +313,72 @@ def get_hosts_from_job_events(self, job_id) -> set:
...
+            self.endpoints.host_metrics,
+            max_concurrency=self.max_concurrency,
+            **self.REQUEST_KWARGS,
+        )
+
+        unique_hosts = set()
+        for metric in metrics_generator:
+            hostname = metric.get("hostname")
+            if hostname:
+                unique_hosts.add(hostname)
```
mirekdlugosz: 🦭 

```suggestion
            if hostname := metric.get("hostname"):
                unique_hosts.add(hostname)
```

quipucords/scanner/ansible/inspect.py
```diff
@@ -262,11 +313,72 @@ def get_hosts_from_job_events(self, job_id) -> set:
...
         hosts_in_inventory = {host["name"] for host in data["hosts"]}
-        hosts_in_jobs = data["jobs"]["unique_hosts"]
-        hosts_not_in_inventory = hosts_in_jobs - hosts_in_inventory
+        # unique_hosts comes from either /api/v2/host_metrics/ or /api/v2/jobs/
+        hosts_automated = set(data["unique_hosts"])
+        hosts_not_in_inventory = hosts_automated - hosts_in_inventory
+
         return {
             "hosts_in_inventory": hosts_in_inventory,
             "hosts_only_in_jobs": hosts_not_in_inventory,
```
mirekdlugosz, line 380: I guess we can't rename this, because this will become a raw fact and part of public API, but... maybe we can? You took the effort to avoid "jobs" in variable names, so it feels a bit strange to leave it here.

quipucords/tests/integration/test_ansible_scan.py
```diff
@@ -129,10 +129,7 @@ def expected_facts():
                 "system_name": "<NODE_IP>",
                 "version": "<ANSIBLE_CONTROLLER_VERSION>",
             },
-            "jobs": {
-                "job_ids": [1, 2],
-                "unique_hosts": ["<DELETED_HOST>", "<LISTED_HOST>"],
-            },
+            "unique_hosts": ["<DELETED_HOST>", "<LISTED_HOST>"],
```
mirekdlugosz: I'm confused. This is part of public API as well, right? So it's OK to remove an object and move one of fields up, but we don't want to rename key in another place? Is there a reason I am missing why one is fine and other is not?

mirekdlugosz (approved):
Tested against AAP in testlab without `QUIPUCORDS_AAP_USE_HOST_METRICS` (to use new approach) and with it set to `0` (to force old approach).

In both cases, scan succeeds.

FWIW, `host_metrics` returns more hosts than old approach. Old approach captures only `localhost`, while new approach captures also one of machines in OpenStack. I think this is a machine I did / attempted to run a job against when preparing AAP (so it isn't completely empty). I don't know why old approach wouldn't capture it - maybe because job was run ages ago?

Left few comments with suggestions / questions. Not really blocking.

abellotti (approved):

abellotti:
I ran this PR against one of my test AAP 2.5

Took 45 seconds to scan the AAP with over 10K hosts.

From the downloaded report:

```
$ cat details-305.json | jq '.sources[].facts[].hosts[].name'   |  wc -l
10243
```

infinitewarp:
> I ran this PR against one of my test AAP 2.5
> 
> Took 45 seconds to scan the AAP with over 10K hosts.
> 
> From the downloaded report:
> 
> ```
> $ cat details-305.json | jq '.sources[].facts[].hosts[].name'   |  wc -l
> 10243
> ```
Thank you for this test run! ❤️ 🎉 

Just following up since Alberto replied in Slack but not here. He also ran against a scan using the stable/main version of quipucords, and it returned the same 10243 hosts but that older version took _6 minutes_ to complete.

Reducing 6 minutes to 45 seconds may not be "multiple orders of magnitude" faster, but it _is_ already _8x_ faster, and since his AAP is populated with synthetic data and may not also be full of job events like a real server might be, I suspect scans of real, in-the-field servers would have an even greater performance improvement.

ruda (approved):

quipucords/tests/integration/test_ansible_scan.py
infinitewarp: Yes, unfortunately this was technically an API-breaking change on my part. When pulling from AAP's host_metrics API, since we get no jobs, removing it here was a premature optimization, but since we don't know how long we need to keep supporting old AAP versions, and we presumably don't want those users to lose the old job_ids values from the details report (the only place it shows up from raw facts), I'll restore this data structure.

quipucords/scanner/ansible/inspect.py
infinitewarp: Agreed, but also I undid the variable renaming (hosts_in_jobs -> hosts_automated -> hosts_in_jobs) in the lines above since I'm keeping the job IDs around for legacy/backwards-compatibility sake. So, at least this code looks somewhat more consistent again. 🤷 

(But I agree, it would be nice to clean up these names some day...)

quipucords/tests/integration/test_ansible_scan.py
infinitewarp: The latest push(es) include a commit that undoes this change, effectively preserving the old data structure and backwards compatibility with old reports. This means reports from modern AAP servers that don't need the jobs API calls will have an empty job_ids list. I left a TODO reminder in the code to improve that data structure if/when we finally rip our support for the old AAP versions.

quipucords/api/aggregate_report/model.py
infinitewarp: Neat idea, but this code is gone as of the latest push.

infinitewarp:
@sourcery-ai summary

infinitewarp:
@sourcery-ai guide

infinitewarp:
@sourcery-ai review

PR #3201, opened 2026-05-26
feat: Set Job as "completed" if any task "completed"

mirekdlugosz:
There are two commits.

The first changes logic in `finalize_scan` to consider entire ScanJob completed (succeeded) if any task inside succeeded. That fixes a problem of multi-source scans where one source has only invalid hosts.

The second ensures that `finalize_scan` will set a ScanJob status, at least in usual cases. Historically, ScanJob status was set all over the place, often in response to a failed task. In multi-source scan scenario, a source with only invalid hosts will fail faster, and would set entire ScanJob as failed even when the other source is being scanned. It would also prevent the logic from first commit, as ScanJob can't move from failed to completed.

Relates to JIRA: DISCOVERY-1363

## Summary by Sourcery

Adjust scan job finalization so job status is derived centrally from task outcomes, and add coverage to ensure multi-source scans complete successfully when any task completes.

Bug Fixes:
- Ensure multi-source scan jobs are marked completed when at least one source task completes successfully instead of remaining failed due to earlier task failures.

Enhancements:
- Centralize ScanJob status setting in finalize_scan based on aggregate task statuses, including handling for all-canceled and no-task edge cases and improved debug logging.
- Update end-to-end and unit tests around scan job/task execution to reflect the new status derivation rules and to avoid invalid source setups in tests.

Tests:
- Expand finalize_scan unit tests to cover combinations of task statuses and expected job status and logging.
- Add an end-to-end test for celery-based multi-source inspect jobs to verify that finalize_scan ultimately sets the job to completed when appropriate.
- Adjust existing task runner tests so job status is no longer set directly by individual task runs.

### DIFF
--- a/quipucords/scanner/job.py
+++ b/quipucords/scanner/job.py
@@ -106,15 +106,12 @@ def run_task_runner(runner: ScanTaskRunner, *run_args):
     # Save Task status
     if task_status == ScanTask.CANCELED:
         runner.scan_task.status_cancel()
-        runner.scan_job.status_cancel()
     elif task_status == ScanTask.PAUSED:
         runner.scan_task.status_pause()
-        runner.scan_job.status_pause()
     elif task_status == ScanTask.COMPLETED:
         runner.scan_task.status_complete(status_message)
     elif task_status == ScanTask.FAILED:
         runner.scan_task.status_fail(status_message)
-        runner.scan_job.status_fail(status_message)
     else:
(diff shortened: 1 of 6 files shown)


infinitewarp (approved):
Short and clean. Love it. 👍 

Also, thanks for hunting down those old leftover bits from when we excised the connect task.

PR #3205, opened 2026-06-02
fix: Actually login to Vault

mirekdlugosz:
Explicitly call `vault_client.auth.cert.login()` to log in using provided certificates.

hvac.Client() will read token from `~/.vault-token`. If developer ever did `vault login` and then ran quipucords through dev server, it appeared that client is correctly authenticating to Vault - but it is using token created `vault` binary.

When quipucords was running in container, it never had `~/.vault-token` and `qpc vault add` always failed.

The returned error message is a bit awkward, with duplicated host:
    Error: Failed to authenticate to HashiCorp Vault 127.0.0.1:8200, VaultError: failed to match all constraints for this login certificate, on post https://127.0.01:8200/v1/auth/cert/login

The first host is kept for consistency with other Vault authentication error conditions. The second host (full API URL) comes from hvac library.

## Summary by Sourcery

Ensure certificate-based authentication to HashiCorp Vault is explicitly performed and surfaced with clearer upstream error handling.

Bug Fixes:
- Trigger explicit cert-based login to Vault instead of relying on any pre-existing token state.

Enhancements:
- Add dedicated handling and messaging for generic VaultError conditions from the hvac client during authentication.

Tests:
- Extend HashiCorp Vault authentication tests to cover upstream VaultError scenarios and verify improved error messages.

### DIFF
--- a/quipucords/api/auth/hashicorp_vault/auth.py
+++ b/quipucords/api/auth/hashicorp_vault/auth.py
@@ -206,6 +206,7 @@ def hashicorp_vault_authenticate(vault_token=None, metadata=None):
     ) as vault_client:
         vault_address = hashicorp_vault_address(metadata)
         try:
+            vault_client.auth.cert.login()
             if vault_client.is_authenticated():
                 logger.debug(api.messages.HASHICORP_VAULT_AUTHENTICATED, vault_address)
                 return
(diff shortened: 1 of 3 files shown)


infinitewarp (approved):
LGTM Thanks!

PR #3207, opened 2026-06-03
Fix permanent scan hang caused by Celery chord expiration and task deletion race

nicolearagao:
## Summary by Sourcery

Extend scan cleanup and Celery configuration to prevent long-running scans from hanging or leaving inconsistent task state.

Bug Fixes:
- Ensure set_scan_task_failure_on_exception always returns a failure tuple, even when the underlying ScanTask is missing or status update fails.
- Fix scan job failure wrapper to mark the ScanJob as failed on unexpected exceptions and to return a consistent value.
- Delete scan tasks and inspect results before removing related TaskConnectionResult records and scan jobs to avoid dangling references and race conditions.
- Configure a sufficiently long Celery result expiration to prevent chord expiration during long-running scans.

Enhancements:
- Increase the default scan job cache TTL to align with extended Celery result retention.

Tests:
- Add tests verifying wrapper behavior when ScanTask is missing or raises exceptions.
- Add a regression test ensuring scans are destroyed in the correct order so tasks are deleted before their connection_result targets.
- Add a test asserting CELERY_RESULT_EXPIRES is configured and long enough for lengthy scans.


quipucords/quipucords/settings.py
```diff
@@ -538,6 +538,7 @@ def warn_on_deprecation(message, category, filename, lineno, file=None, line=Non
 CELERY_BROKER_URL = REDIS_URL
 CELERY_RESULT_BACKEND = REDIS_URL
 CELERY_TASK_ALWAYS_EAGER = env.bool("CELERY_TASK_ALWAYS_EAGER", False)
+CELERY_RESULT_EXPIRES = 90 * 24 * 3600
```
nicolearagao, line 541: env configurable?

infinitewarp: 🤷  Yes, ideally, but pragmatically I don't think this is something we need to worry about in the foreseeable future. Anyone running celery tasks for more than a few days is already a extreme edge case.

infinitewarp (approved):

PR #3240, opened 2026-07-10
feat(api): add scan_name and scan_id filter to reports API

infinitewarp:
Two small API additions to support improved data in the reports table view in quipucords-ui:

- `scan_name` field: The report list and detail endpoints now include the originating scan's name (via `scanjob.scan.name`), or `null` for reports that don't have a related scan (uploaded or merged reports don't have scans).
- `scan_id` filter: The reports list endpoint (`GET /api/v2/reports/`) now accepts a `scan_id` query parameter to return only reports from a specific scan.

Relates to JIRA: DISCOVERY-1343

## Summary by Sourcery

Expose scan metadata on reports and allow filtering reports by originating scan in the v2 reports API.

New Features:
- Add scan_name to report API responses to surface the originating scan's name.
- Allow filtering the reports list by scan_id to return reports from a specific scan.

Tests:
- Add API tests validating scan_name presence or null value on report detail responses.
- Add API test verifying reports list filtering by scan_id returns only matching reports.

### DIFF
--- a/quipucords/tests/api/report/test_reports.py
+++ b/quipucords/tests/api/report/test_reports.py
@@ -54,7 +54,12 @@ from tests.constants import (
     FILENAME_LIGHTSPEED_TGZ,
     FILENAME_SHA256SUM,
 )
-from tests.factories import DeploymentReportFactory, ReportFactory
+from tests.factories import (
+    DeploymentReportFactory,
+    ReportFactory,
+    ScanFactory,
+    ScanJobFactory,
+)
 from tests.report_utils import extract_files_from_tarball
 from tests.utils import raw_facts_generator
 from tests.utils.auth import create_lightspeed_secure_token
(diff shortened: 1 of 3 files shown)


quipucords/tests/api/report/test_reports.py
```diff
@@ -620,3 +625,48 @@ def test_cannot_download_deployments_failed():
...
+    assert results[0]["scan_name"] is None
+
+
+@pytest.mark.django_db(transaction=True)
+def test_reports_filter_by_scan_id(client_logged_in):
+    """Test that reports list can be filtered by scan_id."""
+    scan1 = ScanFactory()
+    scan2 = ScanFactory()
+    report1 = ReportFactory(scanjob=ScanJobFactory(scan=scan1))
+    ReportFactory(scanjob=ScanJobFactory(scan=scan2))
```
cmchase, line 665: Question: is this testing for a side effect? How is scan2 used here?

infinitewarp: We want two reports to exist for two different scans so this test can assert the reports are being filtered correctly by scan ID.

We don't capture the return value from the second ReportFactory (line 665) because we just need the object to be created; nothing in the test needs to assert anything about that object.

abellotti:
lgtm!    Don't we need to update the OpenAPI schema to mention the new attrs?

mirekdlugosz (approved):
Tests feel a bit like testing Django / django_filters. They aren't wrong, and we aren't consistent whether this kind of stuff should be unit-tested or not.

PR #3252, opened 2026-07-27
Improve and future-proof JWS version detection

mirekdlugosz:
JWS installed from zip will count as "potential", unless we also identify a version, when it will be upgraded to "present".

However, JWS version identification depended on static map. That made perfect sense back in 2018, when some JWS versions only reported Apache and Tomcat versions. But since JWS 3.x, JWS is self-reporting correctly and we can just extract version from one of handful of well-known strings.

As long as strings remain unchanged, this also allows us to detect versions released in the future, including versions with multiple digits.

Relates to JIRA: DISCOVERY-1443

## Summary by CodeRabbit

- **New Features**
  - Expanded JBoss Web Server (JWS) version detection to handle more version string formats, including GA, patch, underscore/legacy forms, and unit-name patterns.
  - Improved normalization so more raw inputs map to consistent canonical version outputs.

- **Bug Fixes**
  - Version detection now safely returns no version results when version information is missing.
  - Fixed a minor template whitespace issue in JWS version list combination.

- **Tests**
  - Added unit tests covering successful JWS version normalization and negative/ambiguous inputs, including `None`.

### DIFF
--- a/quipucords/fingerprinter/jboss_web_server.py
+++ b/quipucords/fingerprinter/jboss_web_server.py
@@ -1,5 +1,7 @@
 """Ingests raw facts to determine the status of JBoss Web Server on system."""
 
+import re
+
 from api.models import Product
 from fingerprinter.constants import META_DATA_KEY, PRESENCE_KEY
 from fingerprinter.utils import generate_raw_fact_members, product_entitlement_found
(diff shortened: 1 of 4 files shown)


quipucords/fingerprinter/jboss_web_server.py
```diff
@@ -24,44 +26,58 @@
...
-    "Red Hat JBoss Web Server - Version 5.2.0 GA": "JWS 5.2.0",
-    "Red Hat JBoss Web Server - Version 5.3 GA": "JWS 5.3",
-    "Red Hat JBoss Web Server - Version 5.3.0 GA": "JWS 5.3",
-    "Red Hat JBoss Web Server - Version 5.3.1 GA": "JWS 5.3.1",
 }
 
+JWS_GA_PATTERN = re.compile(
+    r"Red Hat JBoss Web Server - Version (?P<found_version>[0-9\.]+) GA"
+)
+JWS_PATCH_PATTERN = re.compile(
```
ruda, line 34: Using re.VERBOSE ignores all unescaped whitespace and newlines inside the regex pattern string.

In ORIGINAL_JWS_PATCH_PATTERN, the unescaped space before SP (as well as the multiline line breaks) gets stripped by re.VERBOSE. As a result, the compiled regex expects inputs formatted like Version 5.7.0SP1 instead of Version 5.7.0 SP1, causing matches against standard release strings to return None.

Fix: Explicitly use \s+ (or escape spaces with \ ) for any whitespace required between tokens when using re.VERBOSE.

---

Assisted-By: Gemini (Flash).

[pr-3252-JWS_PATCH_PATTERN.py](https://github.com/user-attachments/files/30511960/pr-3252-JWS_PATCH_PATTERN.py)

ruda (changes requested):

quipucords/fingerprinter/jboss_web_server.py
mirekdlugosz: But all spaces are escaped?

> As a result, the compiled regex expects inputs formatted like Version 5.7.0SP1 instead of Version 5.7.0 SP1, causing matches against standard release strings to return None.

This is not true. Standard release strings are matched, and that is covered by unit tests added in this PR. Also that PR fixes JWS detection on real machine.

But lack of space between version and SP are good cases for `test_get_version_negative`. I have now added them (I took the liberty to amend - this is only two lines in params of the test).

FabriciaDinizRH (approved):
LGTM! Nice work making the version check dynamic.

ruda (approved):
LGTM.

ruda (approved):

ruda (approved):

infinitewarp (approved):

quipucords/fingerprinter/jboss_web_server.py
```diff
@@ -24,44 +26,58 @@
...
+    r"""Red\ Hat\ JBoss\ Web\ Server\ -\ Version
+    \ (?P<major_version>[0-9\.]+)\ SP(?P<patch_version>[0-9]+)
+    """,
+    re.VERBOSE,
+)
+JWS_UNIT_NAME_PATTERN = re.compile(r"jws(?P<major_version>[0-9]+)")
```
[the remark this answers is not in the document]
mirekdlugosz: I don't think we are in position to tell JWS that their version string is "malformed".

PR #3254, opened 2026-07-28
fix(ansible): fall back to jobs when host_metrics returns 401/403/404

ruda:
Avoid marking AAP inspect as failed when host_metrics is advertised but denied by RBAC; collect unique hosts via /jobs/ instead.

Relates to JIRA: DISCOVERY-1403
Generated by: Cursor (Grok 4.5)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved host discovery when host-metrics is enabled but unavailable.
  * When host-metrics fails with authorization/not-found errors, the scanner now automatically falls back to jobs-based host discovery and preserves legacy-shaped job results.
  * Keeps existing failure behavior for unexpected server errors so genuine service issues are surfaced.
* **Tests**
  * Added and extended coverage for host-metrics fallback behavior and non-fallback scenarios, aligned to the supported error statuses.

### DIFF
--- a/quipucords/scanner/ansible/inspect.py
+++ b/quipucords/scanner/ansible/inspect.py
@@ -20,6 +20,10 @@ from scanner.ansible.runner import AnsibleTaskRunner
 
 logger = getLogger(__name__)
 
+# HTTP statuses that mean host_metrics is listed but not usable for this
+# credential/server. Fall back to the slower /jobs/ path instead of failing.
+HOST_METRICS_FALLBACK_HTTP_STATUSES = (401, 403, 404)
+
 
 class InspectTaskRunner(AnsibleTaskRunner):
     """Inspection phase task runner for ansible scanner."""
@@ -199,8 +203,6 @@ class InspectTaskRunner(AnsibleTaskRunner):
                 )
                 inspection_status = InspectResult.FAILED
 
-        # Collect unique hosts that have run jobs
-        # Try host_metrics endpoint first (much faster), fall back to jobs if needed
         try:
             results["jobs"] = self.get_jobs_fact()
         except RequestException:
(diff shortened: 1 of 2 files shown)


ruda:
@coderabbitai review

quipucords/scanner/ansible/inspect.py
```diff
@@ -202,40 +202,7 @@ def inspect(self):
         # Collect unique hosts that have run jobs
         # Try host_metrics endpoint first (much faster), fall back to jobs if needed
```
infinitewarp: Since all collection and fallback logic was lifted and moved into `get_jobs_fact`, this comment should move there too because it's not useful out of context.

quipucords/scanner/ansible/inspect.py
```diff
@@ -253,6 +220,67 @@ def inspect(self):
             return self.success_message, ScanTask.COMPLETED
         return self.failure_message, ScanTask.FAILED
 
+    # HTTP statuses that mean host_metrics is listed but not usable for this
+    # credential/server. Fall back to the slower /jobs/ path instead of failing.
+    HOST_METRICS_FALLBACK_HTTP_STATUSES = (401, 403, 404)
```
infinitewarp: Pseudo-constants shouldn't live hidden in the middle of a class between methods. I would move this to the top of the module.

quipucords/tests/scanner/ansible/test_inspect.py
```diff
@@ -268,6 +268,95 @@ def test_inspect_falls_back_to_jobs_when_host_metrics_unavailable(
     )
 
 
+@pytest.mark.django_db
+@pytest.mark.parametrize("status_code", (401, 403, 404))
```
infinitewarp: If we move HOST_METRICS_FALLBACK_HTTP_STATUSES to be a module-level "constant", then we could just use it here instead of recreating its values.

infinitewarp (approved):

PR #3264, opened 2026-08-11
build: remove unused pip from container image

infinitewarp:
We don't actually use pip at runtime. Removing it at build time reduces the risk of later misuse.

## Summary by CodeRabbit

* **Chores**
  * Updated the container build process to install Quipucords in editable mode.
  * Removed packaging tools from the final container image to reduce unnecessary contents.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -74,8 +74,8 @@ COPY --from=builder /opt/venv /opt/venv
 # Copy server code
 COPY . .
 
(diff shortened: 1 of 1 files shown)


mirekdlugosz (approved):
Don't we pull pip, setuptools and wheel with dnf? Or did we end up with both RPM and PyPI copies before?

PR #3277, opened 2026-09-02
refactor: Drop paramiko

FabriciaDinizRH:
Drop Paramiko option from Network scan for [DISCOVERY-1482](https://redhat.atlassian.net/browse/DISCOVERY-1482)

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed the `use_paramiko` option from source configuration and API responses.
  * Existing configurations using this option must be updated.
* **Behavior Changes**
  * Network connections now consistently use the default SSH connection mechanism.
  * Paramiko-specific connection selection is no longer available.

### DIFF
--- a/quipucords/api/source/serializer.py
+++ b/quipucords/api/source/serializer.py
@@ -84,7 +84,7 @@ class SourceSerializerBase(ModelSerializer):
     )
 
     SSH_SOURCE_TYPES = (DataSources.NETWORK,)
-    SSL_OPTIONS = ["ssl_cert_verify", "ssl_protocol", "disable_ssl", "use_paramiko"]
+    SSL_OPTIONS = ["ssl_cert_verify", "ssl_protocol", "disable_ssl"]
 
     class Meta:
         """Metadata for the serializer."""
@@ -106,7 +106,7 @@ class SourceSerializerBase(ModelSerializer):
         # TODO: Remove `apply_defaults` logic once V1 is fully deprecated.
         # V1 sets defaults here (e.g., ssl_cert_verify=True).
         # V2 handles this in validate_http_source() to follow DRF style.
-        valid_ssh_options = ["use_paramiko"]
+        valid_ssh_options = []
         valid_http_options = ["ssl_cert_verify", "ssl_protocol", "disable_ssl"]
 
         if source_type in cls.HTTP_SOURCE_TYPES:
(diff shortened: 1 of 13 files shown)


quipucords/tests/api/source/test_source.py
```diff
@@ -3163,17 +3121,6 @@ def test_openshift_missing_host(self, client_logged_in, openshift_cred):
...
         self.create_expect_400(data, client_logged_in)
 
-    def test_openshift_extra_unallowed_fields(self, client_logged_in, openshift_cred):
-        """Ensure unallowed fields are not accepted when creating openshift source."""
-        data = {
-            "name": "openshift_source_1",
-            "source_type": DataSources.OPENSHIFT,
-            "hosts": ["1.2.3.4"],
-            "credentials": [openshift_cred.id],
-            "use_paramiko": True,
```
abellotti, line 3173: is that the only unallowed field  (use_paramiko) or can we replace that with another unallowed field to make sure we still return 400's in such cases. same with rhacs test removed above.

quipucords/api/source/serializer.py
```diff
@@ -121,8 +120,8 @@ def validate_opts(cls, options, source_type, *, apply_defaults=True):
         elif source_type in cls.SSH_SOURCE_TYPES:
             cls._check_for_disallowed_fields(
                 options,
-                messages.NET_SSL_OPTIONS_NOT_ALLOWED,
-                valid_ssh_options,
+                messages.INVALID_OPTIONS,
```
mirekdlugosz: If you feel strongly about error message consistency, then please remove `NET_SSL_OPTIONS_NOT_ALLOWED` constant from `messages`. This was the only place it was used.

quipucords/api/source/serializer.py
```diff
@@ -106,7 +106,6 @@ def validate_opts(cls, options, source_type, *, apply_defaults=True):
         # TODO: Remove `apply_defaults` logic once V1 is fully deprecated.
         # V1 sets defaults here (e.g., ssl_cert_verify=True).
         # V2 handles this in validate_http_source() to follow DRF style.
-        valid_ssh_options = ["use_paramiko"]
```
mirekdlugosz, line 109: Personally, I would keep `valid_ssh_options` as an empty list here. This gives us clear place where we can define options valid for each connection type. This will come in handy if we ever re-introduce options valid only for network scans.

mirekdlugosz (approved):

quipucords/tests/api/source/test_source.py
abellotti: use_paramiko was the only one. I'm ok with this.

abellotti (approved):

PR #3279, opened 2026-09-04
fix(scanner): use uname -m instead of uname -p for architecture

ruda:
## Motivation

While testing Quipucords with newer distributions like RHEL 10, we discovered an issue with how architecture is reported:
* In the **deployments report**, the `architecture` attribute is incorrectly showing up as `unknown`.
* In the **details report**, the raw fact `uname_processor` (which is collected via `uname -p`) is also returning `unknown`.

Historically in older RHEL versions (7, 8, 9), the OS was configured to make `uname -p` fall back to mimicking `uname -m` (returning `x86_64` or `aarch64`). In newer distributions like RHEL 10, this behavior has been strictly aligned with upstream GNU Coreutils, meaning `uname -p` now correctly reports `unknown` rather than guessing.

Because the `-p` flag is non-standard and non-portable across POSIX systems, this PR switches the network scanner ansible fact collection to use `uname -m`. This reliably queries the machine hardware name from the kernel and serves as the true source for the system architecture for both reports.

To reflect this change accurately, the raw fact is now **fully renamed** from `uname_processor` to `uname_machine` throughout the entire stack:
* The `uname` ansible role now publishes the fact as `uname_machine`.
* The network normalizer and the fingerprinter (network and vCenter paths) consume the new key.
* A **data migration** renames existing `uname_processor` raw facts stored in previously collected details reports, so architecture reporting is preserved for historical data.
* Test fixtures now use the real-world `aarch64` value instead of the placeholder `ARM`.

### Cross-repository change

Because `uname_processor` was part of the details report contract, the corresponding [camayoc](https://github.com/quipucords/camayoc) test suite is updated in quipucords/camayoc#643. That PR must land together with (or after) this one.

**Relates to JIRA**: DISCOVERY-1469

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected machine architecture detection to use the system’s machine identifier, improving accuracy across supported environments.
  * Updated network and virtualization scans to consistently report architecture values such as x86_64 and aarch64.


mirekdlugosz (approved):
The change is correct.

[Fedora bug 2189656](https://bugzilla.redhat.com/show_bug.cgi?id=2189656) has comments explaining why this change was made. It also links to other bugs that provide a decision trail.

One thing I'm slightly worried about is that we keep calling it `(internal_)uname_processor`, which clearly comes from `uname` long flag name, and now we are breaking that link between variable and actual source. Perhaps we should rename it to `(internal_)uname_machine`. 

But then, `uname_processor` is a key in details report. In deployments report we map this to `architecture` field. I never remember which of these is a public contract.

infinitewarp (approved):

quipucords/api/migrations/0011_rename_uname_processor_raw_fact.py
```diff
@@ -0,0 +1,26 @@
...
+    """Rename stored ``uname_processor`` raw facts to ``uname_machine``."""
+    RawFact = apps.get_model("api", "RawFact")
+    RawFact.objects.filter(name="uname_processor").update(name="uname_machine")
+
+
+def revert_uname_machine_to_uname_processor(apps, schema_editor):
+    """Restore the legacy ``uname_processor`` raw fact name."""
+    RawFact = apps.get_model("api", "RawFact")
+    RawFact.objects.filter(name="uname_machine").update(name="uname_processor")
+
```
infinitewarp: JFYI, many of our past migrations do not include reverse/revert operations. We can keep this, but I'm just letting you know that generally speaking our migrations are one way forward exclusively. You probably shouldn't spend any effort writing the reverse functions like this in the future unless there's a special reason.

infinitewarp (changes requested):
We should drop the several empty `ci` commits in this branch before merging.

quipucords/api/migrations/0011_rename_uname_processor_raw_fact.py
ruda: I dropped the reverse function, migration is forward-only now!

ruda:
@infinitewarp done — I squashed the branch down to a single commit, so the empty `ci` commits are gone. Ready for another look.

infinitewarp (approved):
Thanks for the quick commit cleanups! LGTM 👍

FabriciaDinizRH (approved):
LGTM!

mirekdlugosz (approved):

PR #3284, opened 2026-09-14
fix: use `pwd -P` in Makefile to resolve symlinks

infinitewarp:
This change helps if the project is cloned to /tmp and /tmp is a symlink. On macOS, /tmp is usually a symlink to /private/tmp.

Relates to JIRA: DISCOVERY-1430

## Summary by CodeRabbit

* **Chores**
  * Updated project path resolution to use the physical directory path, including resolved symlinks.


ruda (approved):

mirekdlugosz (approved):
I don't know how portable that flag is, but for what it's worth, `pwd` on Fedora 43 does recognize it. I assume the same is true for macOS.

PR #3288, opened 2026-09-21
fix(scanner): collect IP addresses without requiring net-tools

ruda:
## Motivation

Follow-up to the `uname -m`/`uname_machine` work in #3279. While validating that fix across x86_64 and aarch64 scans (RHEL 7-10), we found network scans report **empty `ip_addresses`** on any host without `net-tools`/`ifconfig` installed — i.e. all modern RHEL (8/9/10) and ARM images. Only RHEL 7 (which ships `ifconfig`) got IPs.

## Root cause

The `ip` role already collects addresses via iproute2 (`ip address show`), but the fingerprinter never used them. The `ifconfig` role's IP task ran **unconditionally** with an `else hostname -I` fallback that produced output the ifconfig processor can't parse, so `ifconfig_ip_addresses` became an empty (but non-`None`) list — which the fingerprinter preferred over the populated `ip_address_show_*` facts. (MAC worked only by luck: its ifconfig task was already gated on `internal_have_ifconfig`, so it fell back cleanly.)

## Changes

- **Fingerprinter** (`fingerprinter/runner.py`): use `ifconfig_ip_addresses` only when it actually found addresses; otherwise fall back to the `ip` role and **combine its separate IPv4 + IPv6 facts** (IPv6 was also being dropped on the old fallback). Same truthiness check applied to MAC selection. Flat-key lookups use `fact.get()`.
- **`ifconfig` role** (`roles/ifconfig/tasks/main.yml`): gate the IP task on `internal_have_ifconfig` (mirroring the MAC task) and drop the misleading `hostname -I` fallback, so it no longer emits unparseable data on net-tools-less hosts.
- **Tests**: added coverage for the net-tools-less path (IPv4+IPv6+MAC come from `ip_address_show_*`) and for preferring `ifconfig` when it is present.

Hosts that do have `ifconfig` (e.g. RHEL 7) are unaffected and produce identical output.

## Known follow-up

The new (currently inactive) `scanner/network/normalizer.py` path still reads `ifconfig_*` only and will reintroduce this gap when it is activated — it needs the same fallback.

**Relates to JIRA**: DISCOVERY-1469

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved network fingerprinting when `ifconfig` returns no IP or MAC addresses by using available IPv4, IPv6, and MAC data from the `ip` role.
  * Preserved `ifconfig` data when available, keeping it as the preferred source for network details.
  * Network discovery no longer falls back to `hostname -I` on hosts without `ifconfig`.


quipucords/scanner/network/runner/roles/ifconfig/tasks/main.yml
```diff
@@ -30,17 +30,17 @@
   when: 'user_has_sudo and "stdout_lines" in internal_sudo_ifconfig_mac_addresses_cmd'
 
 - name: gather ip-address fact (no sudo)
-  raw: export LANG=C LC_ALL=C; if command -v ifconfig >/dev/null ; then ifconfig -a; else hostname -I 2>/dev/null; fi
+  raw: export LANG=C LC_ALL=C; ifconfig -a
```
infinitewarp, line 33: So, by removing the `command -v` check, we're now expecting this command to fail if `ifconfig` is not installed or not executable by the current user. Is that the intended behavior? What are the side effects when `ifconfig -a` fails?

ruda:
### Verified end-to-end on the `pr-3288` image

Ran a real network scan (report 15) against a 7-host source (`rmoura-arm64+intel64-rhels`) with the backend on `quay.io/quipucords/quipucords:pr-3288`:

| host | arch | ip_addresses |
|---|---|---|
| discovery-rhel7 (has ifconfig) | x86_64 | `fe80::…`, `10.28.72.5` |
| discovery-rhel8 | x86_64 | `10.28.72.126`, `fe80::…` |
| discovery-rhel9 | x86_64 | `10.28.72.119`, `fe80::…` |
| discovery-rhel10 | x86_64 | `10.28.72.89`, `fe80::…` |
| ip-…-21 (arm) | aarch64 | `10.28.72.21`, `fe80::…` |
| ip-…-42 (arm) | aarch64 | `10.28.72.42`, `fe80::…` |
| ip-…-79 (arm) | aarch64 | `10.28.72.79`, `fe80::…` |

**All 7 hosts now report non-empty `ip_addresses` (IPv4 + IPv6).** Before the fix (on `:latest`), the 6 net-tools-less hosts (RHEL 8/9/10 + the 3 ARM boxes) returned `[]`; only RHEL 7 (which ships `ifconfig`) had IPs. RHEL 7 is unchanged. No regression to the ifconfig path.

quipucords/scanner/network/runner/roles/ifconfig/tasks/main.yml
ruda: The `command -v` check isn't gone, it moved to the `when:` gate. `check_dependencies` already runs `command -v ifconfig` (`internal_have_ifconfig` for sudo, `internal_have_ifconfig_user` for non-sudo), and these tasks now gate on those, just like the MAC tasks in this role already do. So if ifconfig isn't installed/on PATH, the task is **skipped**, `ifconfig -a` never runs-and-fails.

If it's present but fails at runtime: `ignore_errors: yes` keeps the play going, the fact stays empty/NO_DATA, and the fingerprinter falls back to the `ip` role's `ip_address_show_*`. Worst case is a graceful fallback, no failed scan.

quipucords/fingerprinter/runner.py
```diff
@@ -1106,10 +1097,39 @@ def _process_network_fact(self, source: dict, fact: dict) -> dict:
...
+            # did not collect them, so coerce to [] before concatenating (a plain
+            # .get(key, []) would leave None/"" in place and break the +).
+            ip_show_addresses = (fact.get("ip_address_show_ipv4") or []) + (
+                fact.get("ip_address_show_ipv6") or []
+            )
+            self._add_fact_to_fingerprint(
+                source,
+                "ip_address_show_ipv4",
+                fact,
+                "ip_addresses",
```
mirekdlugosz, line 1126: This will produce a fingerprint dict with `raw_fact_key` set to `ip_address_show_ipv4`, but value will be a combination of `ip_address_show_ipv4` and `ip_address_show_ipv6`.

As far as I know, we can't express fingerprint "this value is a combination of bunch of facts".

I'm leaving that to others to decide if that's a problem or not.

mirekdlugosz (approved):
Tentatively approving. I left one comment about something that could or could not be a problem - I would like others to chime in.

quipucords/fingerprinter/runner.py
ruda: Good catch — addressed by recording both keys in the metadata (`ip_address_show_ipv4/ip_address_show_ipv6`), same `/`-joined style `system_creation_date` uses for its combined sources. The value still comes from `fact_value`, so it's just the metadata label that now reflects both facts.

quipucords/tests/fingerprinter/test_fingerprint_task.py
abellotti, line 1439: doesn't sound like this comment applies.

quipucords/fingerprinter/runner.py
```diff
@@ -1106,10 +1097,41 @@ def _process_network_fact(self, source: dict, fact: dict) -> dict:
...
+                fingerprint,
+                fact_formatter=formatters.list_or_none,
+            )
+        else:
+            # These facts may be absent, None, or "" (NO_DATA) when the ip role
+            # did not collect them, so coerce to [] before concatenating (a plain
+            # .get(key, []) would leave None/"" in place and break the +).
+            ip_show_addresses = (fact.get("ip_address_show_ipv4") or []) + (
+                fact.get("ip_address_show_ipv6") or []
+            )
```
abellotti, line 1119: facts may be missing, None, or "",  so _might_ get a TypeError in the concatenation in certain conditions.

Just to be safe, preferring:

```
 ip_show_addresses = (
     formatters.list_or_none(fact.get("ip_address_show_ipv4")) or []
 ) + (
     formatters.list_or_none(fact.get("ip_address_show_ipv6")) or []
)
```

ruda: Good point — switched to `formatters.list_or_none(...) or []` so any non-list (None/""/unexpected value) coerces to [] before the concatenation. Bonus: it now matches the ifconfig branch above, which already uses `list_or_none`. Thanks!

quipucords/tests/fingerprinter/test_fingerprint_task.py
ruda: You're right — updated the comment to match the code: ip_addresses now combines ip_address_show_ipv4 + ip_address_show_ipv6 (and mac_addresses falls back to ip_address_show_mac).

